PR #12, opened 2023-02-17
🐛 Use urlbase for cross-reference links

rowanc1:
See executablebooks/mystjs#225


rowanc1:
Thank you @kolibril13 for finding this. I am always testing in dev mode, and it works there. :)

PR #26, opened 2023-02-27
🚨 Display tex warnings in myst-demo

fwkoch:
This should resolve https://github.com/executablebooks/mystjs/issues/261

### DIFF
--- a/.changeset/weak-lamps-give.md
+++ b/.changeset/weak-lamps-give.md
@@ -1,0 +1,5 @@
+---
+'myst-demo': patch
+---
+
+Display tex warnings in myst-demo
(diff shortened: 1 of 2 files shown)


.changeset/weak-lamps-give.md
```diff
@@ -0,0 +1,5 @@
+---
+'myst-demo': patch
+---
+
+Writing objects: 100% (6/6), 609 bytes | 609.00 KiB/s, done.
```
rowanc1, line 5: I think this might be a typo?

rowanc1 (approved):
Changes look good, I think the changeset picked up some incorrect text?

.changeset/weak-lamps-give.md
fwkoch: Hah. Yeah prolly so...

Issue #34, opened 2023-03-14
mixed output types can be rendered as safe


stevejpurves:
### Describe the bug

**context**
When I do include a jupyter notebook in a myst site that contains  the following code:

```
display(HTML('<div>hello MyST!</div>'))
print('some caption')
```

This results in a code cell with two outputs attached. When the notebook is then rendered as a myst site, the output is erroneously considered as *safe* and the `text/plain` part of the mimebundle of the first output is rendered in place of the `text/html`

**expectation**
I expected the `text/html` to be rendered

**bug**
But instead the whole array of outputs are classes as safe and the `text/html` content is ignored.

**problem**
This is a problem as html content produced in jupyter does not appear on the page as expected.

### Reproduce the bug

1. Create a minimal notebook with the above code in a cell
2. myst init
3. myst start
4. go to the notebook page in the browser

PR #35, opened 2023-03-14
🐛 Fixes incorrect all-safe behaviour for mixed outputs

stevejpurves:
addresses #34


packages/jupyter/src/output.tsx
```diff
@@ -52,6 +52,8 @@ export function Output(node: GenericNode) {
     );
   }
 
+  console.log('Output', allSafe, outputs);
+
```
rowanc1, line 56:
```suggestion
```

rowanc1 (commented):
Can we enable github actions as well for the new tests?!

rowanc1:
Can you also add a changeset please!

PR #46, opened 2023-04-02
🎨  Fix line break and space rendering and align directive margin for code blocks

GYHHAHA:
Hi, I find when we render the text with `\n` and space, it will only be considered as single space since the default html behavior. I have a quick fix. The reason we use the font-mono is we have to keep the original character width the same to copy with some formatted output like NumPy array and pandas Series. Besides, the current margin for the code blocks is too big, and I checked the directive block margin is m-4, thus it much better to change this for unified visual experience. @rowanc1 

Here is the contrast:

---------------------------------

Thanks!

### DIFF
--- a/.changeset/brown-carpets-drop.md
+++ b/.changeset/brown-carpets-drop.md
@@ -1,0 +1,6 @@
+---
+'myst-to-react': patch
+'@myst-theme/jupyter': patch
+---
+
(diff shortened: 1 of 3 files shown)


rowanc1:
Thank you again @GYHHAHA for the pull request, sorry it took so long to get in!! This is an awesome improvement. 🚀

Issue #31, opened 2023-03-05
Page rerenders on scroll


stevejpurves:
### Describe the bug

**context**
When I do scroll up of down the page.

**expectation**
I expected React not to re-render the page

**bug**
But instead rerendering is happening, as confirmed by a console log message in development

**problem**
This is a problem as this re-rendering is unnecessary and can affect page performance.

### Reproduce the bug

1. run the theme in development
2. add a console.log message to a page
3. scroll and watch the console

rowanc1:
Before:
![image](https://user-images.githubusercontent.com/913249/232981248-b63fb199-dcaf-40a3-8516-2fcdd64e9843.png)

After:
![image](https://user-images.githubusercontent.com/913249/232981111-589327db-f6a1-40e3-9685-0eab412df6aa.png)

PR #55, opened 2023-04-19
🐛 Stop the page from rendering on scroll

rowanc1:
See #31

Before:
image

After:
image


rowanc1:
@stevejpurves -- to merge this in we will need to rebase your thebe changes on the book theme (on the dev branch). I think it is good, tested locally -- might be off by a few pixels but the actual page doesn't render which is the bug!

PR #57, opened 2023-04-26
🔧 Consume new thebe myst frontmatter

fwkoch:
This PR consumes the changes here https://github.com/executablebooks/mystjs/pull/367 where myst frontmatter thebe configuration diverges from the thebe-core types.


packages/jupyter/src/providers.tsx
```diff
@@ -14,10 +14,58 @@ import { useNotebookBase, useThebeConfig, useThebeCore, ThebeServerProvider } fr
...
+      binderUrl: useBinder.url ?? binderUrl,
+      ref: useBinder.ref,
+      repo: useBinder.repo,
+      repoProvider: useBinder.provider,
+    };
+  }
+  const useServer = server ?? local;
+  if (server) {
+    const splitUrl = useServer.url?.split('://');
+    const wsUrl = splitUrl?.length === 2 ? `ws://${splitUrl[1]}` : undefined;
```
fwkoch, line 42: I kinda doubt this is what we want to do for the default ws url.

packages/jupyter/src/providers.tsx
```diff
@@ -14,10 +14,58 @@ import { useNotebookBase, useThebeConfig, useThebeCore, ThebeServerProvider } fr
...
+  if (binder) {
+    const useBinder = binder === true ? {} : binder;
+    output.binderOptions = {
+      binderUrl: useBinder.url ?? binderUrl,
+      ref: useBinder.ref,
+      repo: useBinder.repo,
+      repoProvider: useBinder.provider,
+    };
+  }
+  const useServer = local ?? server;
```
fwkoch, line 39: I think we only choose one of these to use - preference goes to local config, when deploying we should strip `local` from the config.

packages/jupyter/src/providers.tsx
```diff
@@ -14,10 +14,58 @@ import { useNotebookBase, useThebeConfig, useThebeCore, ThebeServerProvider } fr
...
+  const {
+    mathjaxUrl,
+    mathjaxConfig,
+    binder,
+    server,
+    kernelName,
+    sessionName,
+    disableSessionSaving,
+    local,
+  } = thebe;
```
fwkoch, line 28: This isn't using the `lite` option, i.e. `useJupyterLite` from before. Not sure where that comes in?

packages/jupyter/src/providers.tsx
```diff
@@ -14,10 +14,58 @@ import { useNotebookBase, useThebeConfig, useThebeCore, ThebeServerProvider } fr
...
+  }
+  output.kernelOptions = {
+    kernelName: useServer?.kernelName ?? kernelName,
+    name: useServer?.kernelName ?? kernelName,
+    path: useServer?.sessionName ?? sessionName,
+  };
+  if (!disableSessionSaving) {
+    output.savedSessionOptions = {
+      enabled: true,
+      maxAge: 38300,
```
fwkoch, line 58: Do we want to be filling in these defaults here?

Issue #61, opened 2023-04-26
When installing, it now fails


rowanc1:
The patches need to also be in the theme repos at the moment. The install now fails -- @stevejpurves any idea how to get around that?

rowanc1:
Ok, I think that to change this we:

(1) need to move the source of the themes over into this repo
(2) when building, bundle everything by changing the remix config:

remix.config:
```
  appDirectory: 'app',
  assetsBuildDirectory: 'public/build',
  serverBuildPath: 'build/index.js',
  serverMinify: true,
  publicPath: '/build/',

// Get rid of this line:
// server: process.env.NODE_ENV === 'development' ? undefined : './server.js',

// Bundle everything
serverDependenciesToBundle: [/.*/],
```

3. When build this should create a full api bundle and copy of the public repo.
4. Copy those assets and the standard remix electron shell in build and public folders
5. have the packagejson/lock look like:

```
{
  "name": "@myst-theme/book",
  "private": true,
  "sideEffects": false,
  "scripts": {
    "start": "cross-env NODE_ENV=production node ./server.js"
  },
  "dependencies": {
    "@remix-run/express": "^1.15.0",
    "@remix-run/node": "^1.15.0",
    "compression": "^1.7.4",
    "cross-env": "^7.0.3",
    "express": "^4.18.2",
    "isbot": "^3.6.8",
    "morgan": "^1.10.0"
  },
  "engines": {
    "node": ">=16"
  }
}
```

6. These should be copied into a build folder that are the git repos, I think it is fine to have these be in git for now rather than some other deployment
7. npm install in that build folder to make the lock files (should only take a second as there are barely any deps)
8. push that to the template repository, which has a template.yml, server.js, .gitignore, public and build folder.

The theme needs to be fixed up to allow for changes in development mode ([here](https://github.com/executablebooks/myst-theme/blob/main/packages/site/src/components/ContentReload.tsx#L59)).

PR #62, opened 2023-04-28
📦 Packaging and deployment of the themes

rowanc1:
This fixes and follows the process of #61.


rowanc1:
@fperez this fixes an issue that you brought up on the massive bundle that is downloaded for each theme. This PR now is pre-building the templates, and deploying the built artifacts. This means a faster startup time for cloning a template, a faster install of npm packages, and MUCH smaller folder of assets. Now around 30x-35x better than before.

image

We are shipping a good chunk of Jupyter now with Thebe merged in, so the files aren't tiny for the server, but we are down to <25MB download the first time you get the theme. Better than 800 MB for sure. :)

Issue #60, opened 2023-04-26
Navigating between pages looses outputs


rowanc1:
This is with the thebe updates:

![image](https://user-images.githubusercontent.com/913249/234723267-0610afca-ac56-42de-b755-6baf834f89c0.png)

rowanc1:
This gets fixed on refresh.

Issue #59, opened 2023-04-26
New thebe theme recursive render freezes page


rowanc1:
In the book theme, clicking on this link in the mystjs docs will break the page:

http://localhost:3000/mystjs/cross-references#targeting-cells

![image](https://user-images.githubusercontent.com/913249/234723041-c96f0436-84bc-41f9-8714-1b17e198cf43.png)

PR #64, opened 2023-04-30
fix/60

stevejpurves:
* [x] #60 
* [x] #59
* [x] try for jlite support - see #70 
* [ ] 🐞cannot set a local server connection to a different port - websocket error
* [ ] fix ipywidgets support

### DIFF
--- a/packages/jupyter/src/utils.ts
+++ b/packages/jupyter/src/utils.ts
@@ -1,0 +1,213 @@
+import type {
+  Thebe,
+  ThebeServerOptions,
+  ThebeLocalOptions,
+  ThebeBinderOptions,
+} from 'myst-frontmatter';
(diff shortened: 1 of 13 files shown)


packages/jupyter/src/utils.ts
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  // handle thebe.local.*
+  // as local OVERIDES other binder and server settings, handle these first
+  // TODO need to expose NODE_ENV somehow via a loader
+  let NODE_ENV = 'development';
+  if (typeof window !== 'undefined') {
+    NODE_ENV = (window as any).NODE_ENV;
+  }
+  if (NODE_ENV !== 'production' && local) {
```
stevejpurves, line 96: TODO probably for first pass, you need to remove your local key before deploying

PR #68, opened 2023-05-04
📄 Update article theme for single project, consume dataUrl

fwkoch:
This modifies `article-theme` so it is now intended for single-project-to-site configurations. Primarily it removes `projectSlug` from the url.

This also consumes `dataUrl` on links/xrefs so source mdast json data can be accessed directly from any provided link (rather than appending `.json` to the source url). This has nice implications around routing the index page for static HTML, i.e. the index page url can be simply `/` while the mdast is available at `/index.json` (or something equivalent...). This helps out with https://github.com/executablebooks/mystjs/issues/188


rowanc1:
Ok, this is tested to work for backwards compatible sites as well as for the next version -- going to merge and deploy the themes.

PR #70, opened 2023-05-06
`thebe-lite` integration into the themes

stevejpurves:
FvcMlqHXwAAuXZx


packages/site/src/pages/Root.tsx
```diff
@@ -59,6 +59,7 @@ export function Document({
           analytics_google={config?.analytics_google}
           analytics_plausible={config?.analytics_plausible}
         />
+        <script src="/lite/thebe-lite.min.js" async defer></script>
```
rowanc1, line 62: We probably want to load this on demand rather than on every page?

PR #76, opened 2023-05-11
[FIX] wrap gridded content in div for block display

tavin:
Hi @rowanc1 I really like what you've done for myst-based websites. The book theme looks great overall but I've run into a couple of quirks.

One is that paragraph margins are being doubled in the main content. The default margin is `1.25em` which seems reasonable but `2.5em` margins are actually displayed, which look enormous.

image

It's because margins are not collapsing since the HTML has been put directly inside a container with `display:grid`.

I offer a very simple fix, wrapping the content in a `div` which naturally has `display:block`.

### DIFF
--- a/packages/site/src/components/ContentBlocks.tsx
+++ b/packages/site/src/components/ContentBlocks.tsx
@@ -41,13 +41,15 @@ function Block({
         [subGrid]: !noSubGrid,
       })}
     >
-      {children}
-      {isACodeCell(node) && (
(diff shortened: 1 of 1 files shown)


rowanc1:
This has been bugging me for a while as well, thanks for the PR! I don't think the way you solved it is going to allow us to do some things we are planning in the next bit around styling and grid layouts. For example:

**Putting things in the margins:**
image

**Stretched out:**
image

**Full-bleed:**
image

Let me futz with css for a bit to try and keep that, while also doing margin-collapse.

PR #77, opened 2023-05-11
👩🏻‍🎨 Improve margins and spacing in grids

rowanc1:
Originally created by @tavin in #76

This tightens up margins and simulates margin collapse in a grid by eating the top margins. This works inside and outside of the grid, while also allowing us to position child content in the margins.

Thanks @tavin for getting the ball rolling here, sorry to step on your PR. I will get this out in the next day or so as a theme-release!

# Examples / Comparisons

image

image

### DIFF
--- a/packages/frontmatter/src/FrontmatterBlock.tsx
+++ b/packages/frontmatter/src/FrontmatterBlock.tsx
@@ -174,7 +174,7 @@ export function DoiBadge({ doi: possibleLink, className }: { doi?: string; class
   return (
     <div className={classNames('flex-none', className)} title="DOI (Digital Object Identifier)">
       <a
-        className="font-light hover:font-light no-underline text-inherit hover:text-inherit"
+        className="font-light hover:font-light no-underline hover:underline text-inherit hover:text-inherit"
         target="_blank"
         rel="noopener noreferrer"
         href={url}
@@ -289,7 +289,7 @@ export function FrontmatterBlock({
     subject || github || venue || biblio || open_access || license || hasExports || isJupyter;
   const hasDateOrDoi = doi || date;
   return (
-    <>
+    <div className="mb-8">
       {hasHeaders && (
         <div className="flex mt-3 mb-5 text-sm font-light items-center h-6">
           {subject && (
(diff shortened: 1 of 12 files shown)


styles/grid-system.css
```diff
@@ -9,6 +9,8 @@
   .article-grid > * {
     /* The default is spanning the body for any child component */
     @apply col-body;
+    /* Grids do not have margin-collapse, so each direct child needs to be addressed */
+    margin-top: 0 !important;
   }
```
rowanc1, line 14: @tavin, this is basically the fix. The rest is making sure that callouts/equations are balanced and have the same margins as paragraphs.

tavin: understood :+1:

packages/frontmatter/src/FrontmatterBlock.tsx
```diff
@@ -289,7 +289,7 @@ export function FrontmatterBlock({
     subject || github || venue || biblio || open_access || license || hasExports || isJupyter;
   const hasDateOrDoi = doi || date;
   return (
-    <>
+    <div className="mb-8">
```
tavin, line 292: May I just ask about this `mb-8` here? It's giving me `2rem` extra space that wasn't there before -- between the title and content, or else just pushing the content down when there's no title.

If I remove the `mb-8` class everything looks perfect.

rowanc1: Ah, yes, if you only have a title this isn't going to be good. I was testing with a full frontmatter, and that should be on conditionally (or something). Good catch!

We want these conditionally if there is frontmatter. Or maybe better is to override the `h1` here and have an explicit `mb-0` and then have some margin around the stuff that comes after it which is conditionally rendered.

Are you in a place to open a PR for this? If not, I can fix my bug shortly!

The two green things stack, and they should collapse if there isn't anything in between.

image

tavin: Hm I guess I need to think about it a bit. You may get there faster than me.

rowanc1: Yeah, I am not quite sure which solution is best. I will try and get to it by the end of today and tag you! Thanks again for pointing it out. :)

PR #79, opened 2023-05-11
💅 Improve the frontmatter spacing

rowanc1:
@tavin this should help out! Thanks again fro catching this. :)

### DIFF
--- a/.changeset/metal-insects-breathe.md
+++ b/.changeset/metal-insects-breathe.md
@@ -1,0 +1,5 @@
+---
+'@myst-theme/frontmatter': patch
+---
+
+Improve spacing under title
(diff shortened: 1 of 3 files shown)


tavin:
@rowanc1 I've tried your latest changes and everything's looking much better, thank you!

At some point you might wanna look at what happens when there's no frontmatter (and no H1) at all -- again, it results in some unwanted space on top. Or I may try to fix that myself.

But it's a minor concern, I think I will always have a frontmatter title or H1 in my current project.

rowanc1:
Awesome! Slowly we will get to this being 💯...

Would love a PR if you are up for it!

PR #87, opened 2023-05-13
👩🏽‍🏫 Initial support for proofs and theorems

rowanc1:
Based on `sphinx-proof`

See: https://github.com/executablebooks/mystjs/pull/384

### DIFF
--- a/packages/myst-to-react/src/proof.tsx
+++ b/packages/myst-to-react/src/proof.tsx
@@ -1,0 +1,213 @@
+import type { Admonition as AdmonitionSpec } from 'myst-spec';
+import React from 'react';
+import type { NodeRenderer } from '@myst-theme/providers';
+import ChevronRightIcon from '@heroicons/react/24/solid/ChevronRightIcon';
+import classNames from 'classnames';
+import { HashLink } from './heading';
+
+export enum ProofKind {
+  proof = 'proof',
+  axiom = 'axiom',
+  lemma = 'lemma',
+  definition = 'definition',
+  criterion = 'criterion',
+  remark = 'remark',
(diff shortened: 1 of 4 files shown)


packages/myst-to-react/src/proof.tsx
```diff
@@ -0,0 +1,213 @@
...
+        // 'border-l-2 border-red-500': color === 'red',
+      })}
+    >
+      <HeaderElement
+        dropdown={dropdown}
+        className={classNames(
+          'm-0 font-medium py-2 flex min-w-0',
+          'text-md',
+          '',
+          'border-y dark:border-y-0 dark:border-l-4 border-slate-400',
```
tavin, line 147: So (this is great btw) the `dark:border-l-4` here causes a small misalignment. To fix it either `dark:ml-3` is needed on L164 below, or else `dark:border-l-4` should be moved up to the WrapperElement. I prefer the latter solution for parity with dark admonitions.

I tried to do it myself but it results in the wrong border color and I'm not sure how to fix that.

rowanc1: Ah, ok -- first attempt is in #90, I misread this comment though and see now you were thinking proofs should be more like admonitions. I went back and forth on this in my first pass and decided to make them a bit different intentionally, but not a lot of reason behind that.

I will explore this now and post a screenshot of options.

rowanc1: Ok, I think I do like this better!

image

tavin: I like it better too, I think -- meaning w/ the long border. I am not sure whether the original border color was nicer... though not a big deal.

PR #90, opened 2023-05-14
🎨 Alignment of proof header

rowanc1:
Thanks @tavin for pointing it out!


rowanc1:
The screenshots are below. The dark mode is 4px more indented (but now aligned!), but I think it is more balanced than going to a smaller margin to account for the 4px border. I really like how the light-mode has turned out, and with the changes that @tavin suggested, I think that the dark mode is a bunch nicer now as well!

Curious @tavin if you have strong thoughts on direction here, I think this is an improvement. :)

@mmcky -- also if you have thoughts on color, I didn't give it to much thought beyond copying the general scheme you put together in sphinx proof.

image

image

tavin:
I like this although I find the red a bit on the "high contrast" side. But I have my own local tweaks anyway :)

mmcky:
thanks @rowanc1 and @tavin -- these look like nice changes. 

I will take note of this PR and try and push some of these improvements into `sphinx-proof` as well.

mmcky:
@rowanc1 

I tend to agree with this. What if the line in dark mode is the same as the background colour in light mode (for consistency?)

> I like this although I find the red a bit on the "high contrast" side.

rowanc1:
Knocking it back a bit, much better -- good call. :)

image

image

image

rowanc1:
Ok -- people on [twitter](https://twitter.com/rowancockett/status/1657890297815375874) also like the full border.


Merging this in!

PR #91, opened 2023-05-15
🔗 Add support for github issue previews

rowanc1:
- See https://github.com/executablebooks/mystjs/pull/389

github-previews

### DIFF
--- a/.changeset/cold-schools-drive.md
+++ b/.changeset/cold-schools-drive.md
@@ -1,0 +1,5 @@
+---
+'myst-to-react': patch
+---
+
+Add github issue and pull request previews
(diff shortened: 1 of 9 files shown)


rowanc1:
cc @choldgraf this is the work for the previews we were talking about! I think there is a bunch to improve (I have just skipped over github markdown parsing for the body text). That choice is mainly to keep the JS bundle small, and I think for a preview, it mostly looks ok.

I have picked up on the tags, and have a simple choice to show white or black text on it.

Going to push through and get this out the door and start testing it more! 🚀

@tavin if you have thoughts, always feel free to comment on the PRs and I/we will follow up and fix things (defaulting to [ship/show](https://martinfowler.com/articles/ship-show-ask.html) pattern, to move fast, but absolutely open to feedback and fixing things even though a PR is merged!!).

Issue #99, opened 2023-05-28
Single line formula always holds scrollbar when apply Katex style


GYHHAHA:
### Describe the bug

Now the Katex rendering will result to all single line formula hold scrollbar like
![28 05 2023_11 41 55_REC](https://github.com/executablebooks/myst-theme/assets/41546976/d76cd486-a66c-4656-a3e3-fc00ce39a936)
In order to fix this, I localize the katex.css and add one more overflow rule on that
```css
.katex-display {
  overflow-y: hidden;
}
```
This one has the desired behavior and I have verified that this won't break overflow-x-auto and multi-line formula rendering
![28 05 2023_11 43 52_REC](https://github.com/executablebooks/myst-theme/assets/41546976/6a8a534e-66b5-4be9-9a23-5a12ac3956ce)
But this is still kind of cumbersome for direct math component use, it better it add the `overflow-y-hidden` at here, which I think is a better solution.
https://github.com/executablebooks/myst-theme/blob/3420bb7562cca1d8a8cf906af2d576c00a2dfb5e/packages/myst-to-react/src/math.tsx#L58
If @rowanc1 is satisfied with this change, I will make a pr to fix.

### Reproduce the bug

/

rowanc1:
Thanks for the bug report, so strange! Is this a new bug or has it always been there?

Happy to have a PR for this! Thank you!!

GYHHAHA:
I'm not sure about when this starts because I didn't do much test on katex before. However, I can give you some more strange examples, that is not all the single line formulas have the same behavior:
```latex
$$
abc
$$

$$
\iint
$$

$$
\sum_{}
$$

$$
\frac{1}{2}
$$

$$
\sum_{i=1}
$$
```
You have a quick verify in the sandbox that the first 3 are good, but the last 2 don't work. @rowanc1

rowanc1:
🚀 I will release soon with a few of these fixes!! Thanks again for your help!

PR #100, opened 2023-05-28
🐛 Fix single line formula scrollbar issue

GYHHAHA:
see #99


rowanc1:
Thank you @GYHHAHA -- confirmed the bug on my side as well. So strange, but your fix works nicely! :)

GYHHAHA:
Maybe caused by some Katex black magic behind the scenes :joy:

PR #117, opened 2023-06-08
🆙 Updates to demo and various packages

rowanc1:


rowanc1:
@stevejpurves this is maybe going to break thebe and jupyter outputs, we need to circle back to that, but I am going to merge it in and we can figure it out on the other side.

PR #130, opened 2023-06-20
🎨 Add template option for logo in dark mode

eurunuela:
This PR stems from the conversation in [mystjs #155](https://github.com/executablebooks/mystjs/issues/155).

### DIFF
--- a/packages/site/src/components/Navigation/TopNav.tsx
+++ b/packages/site/src/components/Navigation/TopNav.tsx
@@ -204,10 +204,15 @@ function HomeLink({
     >
       {logo && (
         <div
-          className={classNames('p-1 mr-3', {
-            'bg-dark:bg-white': !logoDark,
-            'dark:rounded': !logoDark,
-          })}
+          className={classNames(
+            'p-1 mr-3',
+            {
+              'dark:bg-white': !!logoDark,
+            },
+            {
+              'dark:rounded': !!logoDark,
+            },
+          )}
         >
           <img src={logo} className="h-9 dark:hidden" alt={logoText || name} height="2.25rem"></img>
           <img
--- a/packages/site/src/components/Navigation/TopNav.tsx
+++ b/packages/site/src/components/Navigation/TopNav.tsx
@@ -57,7 +57,7 @@ function ExternalOrInternalLink({
 function NavItem({ item }: { item: SiteNavItem }) {
   if (!('children' in item)) {
     return (
-      <div className="relative grow-0 inline-block mx-2">
+      <div className="relative inline-block mx-2 grow-0">
         <ExternalOrInternalLink
           nav
           to={item.url ?? ''}
@@ -76,9 +76,9 @@ function NavItem({ item }: { item: SiteNavItem }) {
     );
   }
   return (
-    <Menu as="div" className="relative grow-0 inline-block mx-2">
+    <Menu as="div" className="relative inline-block mx-2 grow-0">
       <div className="inline-block">
-        <Menu.Button className="inline-flex items-center justify-center w-full mx-2 py-1 text-md font-medium text-stone-900 dark:text-white rounded-md focus:outline-none focus-visible:ring-2 focus-visible:ring-white focus-visible:ring-opacity-75">
+        <Menu.Button className="inline-flex items-center justify-center w-full py-1 mx-2 font-medium rounded-md text-md text-stone-900 dark:text-white focus:outline-none focus-visible:ring-2 focus-visible:ring-white focus-visible:ring-opacity-75">
           <span>{item.title}</span>
           <ChevronDownIcon className="w-5 h-5 ml-2 -mr-1 text-violet-200 hover:text-violet-100" />
         </Menu.Button>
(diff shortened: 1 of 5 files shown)


packages/site/src/components/Navigation/TopNav.tsx
```diff
@@ -57,7 +57,7 @@ function ExternalOrInternalLink({
 function NavItem({ item }: { item: SiteNavItem }) {
   if (!('children' in item)) {
     return (
-      <div className="relative grow-0 inline-block mx-2">
+      <div className="relative inline-block mx-2 grow-0">
```
rowanc1, line 60: What vscode extension do you use to order tailwind classes? (I am assuming?)

I was looking into/thinking about this today!

packages/site/src/components/Navigation/TopNav.tsx
```diff
@@ -182,19 +182,35 @@ function ActionMenu({ actions }: { actions?: SiteManifest['actions'] }) {
...
   const nothingSet = !logo && !logoText;
   return (
     <Link
-      className="flex items-center dark:text-white w-fit ml-3 md:ml-5 xl:ml-7"
+      className="flex items-center ml-3 dark:text-white w-fit md:ml-5 xl:ml-7"
       to={withBaseurl('/', baseurl)}
       prefetch="intent"
     >
       {logo && (
-        <div className="dark:bg-white dark:rounded p-1 mr-3">
```
rowanc1, line 196: I think we might want another branch/if-statement here for the existing behaviour when the dark logo is not defined. I think that should still have the white background?

rowanc1 (commented):
This is awesome, glad that you found everything! 🚀 

I will test it out in the next few days, but looks solid, I think we just want a bit more when the dark logo is not provided.

packages/site/src/components/Navigation/TopNav.tsx
eurunuela: I use [Headwind](https://marketplace.visualstudio.com/items?itemName=heybourn.headwind).

packages/site/src/components/Navigation/TopNav.tsx
eurunuela: Yeah, I was thinking about that too.

Probably something like `logoDark ?? <div className="..."></div>` will work.

rowanc1: The `dark:hidden` class name also needs to be added depending as well.

We have been using something like:

```jsx
<img className={classNames('shared-classes', { 'darkhidden': !!logoDark })} />
```

eurunuela: Oh, that looks much cleaner. Thanks Rowan!

eurunuela: What do you think of https://github.com/executablebooks/myst-theme/pull/130/commits/2e3b14fb79e55448fc9b50b2c518e8047cee912c

eurunuela:
I'm testing these changes but I'm getting this error:

`'options' extra key ignored: logo_dark (at myst.yml)`

Any idea?

packages/site/src/components/Navigation/TopNav.tsx
rowanc1: Doing this for the whole project in #135. I will merge after this PR goes in so there are no conflicts.

rowanc1:
Works great -- I updated the logic a bit after testing it out locally.

There is one bit of logic that I am not stoked about, the file URLs need to be updated dynamically and this should happen based on the template, rather than hard-coded. For now I think this is ok, and we can come back and fix this up later when we are copying more files in the template (e.g. custom styles or something).

dark-logo

I also added it to the article theme, and put in a changeset! Will release in the next few days!

rowanc1:
Thanks for your help with this @eurunuela!

eurunuela:
Thank you @rowanc1 for the help!

eurunuela:
> Also for testing:
> 
> ```yaml
> site:
>   template: ../path/to/myst-theme/book
> ```
> 
> This is a bit annoying for testing as even when running in `myst start --headless` it still reads the local template options (which doesn't have the `logo_dark`) to know to copy the file over, etc.

Either I'm doing something wrong or the snippet above doesn't work. I get this error:

```
Problem with template link "https://api.mystmd.org/templates//Users/eurunuela/GitHub/myst-theme": 404 Not Found

To list valid templates, try the command "myst templates list"
```

eurunuela:
Hey @rowanc1, I'm still having the issue with the logo:

`'options' extra key ignored: logo_dark (at myst.yml)`

I updated to the latest `mystmd` version, so I don't know if there is something in my setup that is making it fail. Any ideas?

rowanc1:
I think that you are on a previous template. Can you try:

```
myst clean --templates
```

And then re-run `myst start`?

eurunuela:
Agh cleaning did it! Thank you Rowan!

Maybe the cleaning command could happen when updating myst?

rowanc1:
Yeah, we need to document this better and have some indication that the templates are out of date. Right now we are skirting versioning on the templates and starting to be a bit of an issue in some cases.

Adding a quick note on this in [#467](https://github.com/executablebooks/mystmd/pull/467).

PR #133, opened 2023-06-21
👮🏻check for jq before commiting

stevejpurves:
otherwise you get a poor commit message which is auto pushed


rowanc1 (approved):
Nice, thank you!

PR #134, opened 2023-06-21
🛠 fix for empty blocks in notebooks

stevejpurves:
this caused a runtime error in notebook processing from mdast

### DIFF
--- a/packages/jupyter/src/providers.tsx
+++ b/packages/jupyter/src/providers.tsx
@@ -83,7 +83,6 @@ export function notebookFromMdast(
   //Object.assign(notebook.metadata, ipynb.metadata);
   notebook.cells = (mdast.children as GenericParent[]).map((block: GenericParent) => {
     if (block.type !== 'block') console.warn(`Unexpected block type ${block.type}`);
-    console.log(`BLOCK`, block);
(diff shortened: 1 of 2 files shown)


packages/jupyter/src/providers.tsx
```diff
@@ -105,7 +105,7 @@ export function notebookFromMdast(
       const cell = new core.ThebeNonExecutableCell(
         block.key,
         notebook.id,
-        block.children.reduce((acc, child) => acc + '\n' + (child.value ?? ''), ''),
+        block.children?.reduce((acc, child) => acc + '\n' + (child.value ?? ''), ''),
```
rowanc1, line 108: I think the if-statement will also fall through to here. This will mean that the content is undefined?

Is that fine? I am not sure what the `ThebeNonExecutableCell` does in this case!

stevejpurves: nice catch -- the no children condition only occurs with empty markdown cells, so we want to create the non executable cell with an empty source string in that case

rowanc1:
Nice -- glad that you found the source of this, makes sense I think, I will make an issue upstream to ensure that blocks have a children list always as well.

PR #136, opened 2023-06-23
feat/thebe v2 - interactive figures in articles

stevejpurves:
Current `thebe` integration only applies to notebooks, whereas we can include notebook outputs in any markdown file as a figure or embedding.

This PR will replace the `thebe` provider system with one that can support interactive figures anywhere in a myst site, in addition to still enabling the notebooks to be run and made interactive as before. 

Goals:
- [x] any notebook on a site is identified as "computable" and can be run
- [x] any article containing a notebook output is identified as "computable" and can be run
- [x] execution state/context is separate between pages on a site
- [x] a user can navigate around the site, without losing their session when returning to a page previously visited and run
- [x] simplified UI for notebooks
- [x] new compact UI for figures in markdown
- [x] UI feedback during server connections
- [x] connections/startups that are initiated will continue to progress even if the user navigates away
- [DEFERRED] Errors that occur in parts of a notebook that are not attached to the UI are surfaced - see #139


packages/site/src/components/Navigation/TableOfContents.tsx
```diff
@@ -163,7 +163,7 @@ export const TableOfContents = ({
     <div
       ref={tocRef}
       className={classNames(
-        'fixed xl:article-grid article-grid-gap xl:w-screen xl:pointer-events-none overflow-auto max-xl:min-w-[300px]',
+        'fixed xl:article-grid article-grid-gap xl:w-screen xl:pointer-events-none overflow-auto max-xl:min-w-[300px] z-20',
```
stevejpurves, line 166: @rowanc1 thoughts on this? it was to accomodate the z-10 being added to the compute controls and to prevent overlap on narrow screens which made the first TOC item unclickable

rowanc1:
I think this is good to go in after a few minor changes that I am making locally. Notes are for future PRs.

There is no error feedback when a `server: true` fails to connect.

Navigating back to a notebook does not reattach cells (I thought that was in scope?).

Navigating back to an article does not reattach the model correctly.
image

It would be nice to figure out a way to only install things on lite:
image

In lite, there are calls to `/pypi/all.json`, should this be the place that we put our dependencies?

packages/site/src/components/Navigation/TableOfContents.tsx
rowanc1: This will interfere with cross-reference popups.

packages/site/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "extends": "../tsconfig/react-library.json",
-  "include": ["."],
+  "include": [".", "../jupyter/src/execute", "../jupyter/src/controls"],
```
rowanc1, line 3: This is due to a drag-n-drop in vscode.

packages/site/src/components/Navigation/TableOfContents.tsx
rowanc1: Can we back this out and put pointer-events none on the part of the control that isn't active?

rowanc1: I have reverted. Given that the controls need another pass, I think we should have them not work rather than relying on the z-index pattern for the app (I had gotten away from that with the exception of the top-nav).

stevejpurves:
navigating back should reattach cells -- will investigate.

afaik the `/pypi/all.json` is more like a fake `pypi.org` main listing, rather than the list of wheels that should be loaded

packages/jupyter/src/controls/NotebookToolbar.tsx
```diff
@@ -41,8 +41,8 @@ export function NotebookToolbar({ showLaunch = false }: { showLaunch?: boolean }
 
   if (computable)
     return (
-      <div className="sticky top-[60px] flex justify-end w-full z-20">
-        <div className="flex p-1 m-1 space-x-1 border rounded-full shadow border-stone-300 bg-white/80 dark:bg-stone-900/80 backdrop-blur">
+      <div className="sticky top-[60px] flex justify-end w-full z-20 pointer-events-none">
+        <div className="flex p-1 m-1 space-x-1 border rounded-full shadow pointer-events-auto border-stone-300 bg-white/80 dark:bg-stone-900/80 backdrop-blur">
```
stevejpurves, line 45: guessing this was to stop the pointer showing up over blank space in the rest of the `div` yeah?

packages/site/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "extends": "../tsconfig/react-library.json",
-  "include": [".", "../jupyter/src/execute", "../jupyter/src/controls"],
+  "include": ["."],
```
stevejpurves: hmm wonder if this was a vscode auto thing, as i didnt; add these explicitly

rowanc1: Yeah, when you drag from one folder to another it happens!

packages/jupyter/src/controls/NotebookToolbar.tsx
rowanc1: Yep!

rowanc1 (commented):
Woops, didn't post my comments!

packages/site/src/components/Navigation/TableOfContents.tsx
```diff
@@ -163,7 +163,7 @@ export const TableOfContents = ({
     <div
       ref={tocRef}
       className={classNames(
-        'fixed xl:article-grid article-grid-gap xl:w-screen xl:pointer-events-none overflow-auto max-xl:min-w-[300px] z-20',
+        'fixed xl:article-grid article-grid-gap xl:w-screen xl:pointer-events-none overflow-auto max-xl:min-w-[300px]',
```
stevejpurves: @rowanc1 reverting this causes two issues: the menu items can be covered by the toolbar div and the inset black background no longer blacks out the compute toolbar. The toolbar is set to `z-10` to lift above content in the main body.

image

did the `z-20` here interfere with something else?

packages/jupyter/src/execute/actions.ts
```diff
@@ -0,0 +1,124 @@
...
+  type:
+    | 'NAVIGATE'
+    | 'REQUEST_BUILD'
+    | 'BUILD_STATUS'
+    | 'CLEAR_BUILD'
+    | 'ADD_MDAST'
+    | 'ADD_NOTEBOOK'
+    | 'ADD_SESSION'
+    | 'SET_FIRST_EXECUTION'
+    | 'SET_RENDERING_READY';
```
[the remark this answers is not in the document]
stevejpurves: yeah the typing is not great, so strengthening that is good. I'm getting turned off enums though as they're not types and trying to use the thebe event enums for example are painful.

packages/site/src/components/Navigation/TableOfContents.tsx
stevejpurves: or is there a better way to deal with this overall?

rowanc1: It breaks the cross-references which appear under the toc.

The pointer-events-none should allow the controls to be mostly fixed though, only the right hand side is active?

rowanc1: Left some other notes on this here:
https://github.com/executablebooks/myst-theme/pull/136#discussion_r1251169214

PR #137, opened 2023-06-27
⛺️ `thebe` async loading uses `BASE_URL`

stevejpurves:
This should fix https://github.com/executablebooks/mystjs/issues/441 by using the correct `BASE_URL`.

- [x] Tested locally both with and without the BASE_URL in normal operation (nr theme:book + headless content).
- [x] Tested locally with a static build (by copying over the deployable theme into a _build folder)
- [x] Tested on githuhbpages (https://github.com/stevejpurves/myst-online-media-testing, https://stevejpurves.github.io/myst-online-media-testing/)

I could not get a local static build or manual deployment to gh-pages working (the [docs](https://mystmd.org/guide/deployment#deploying-to-github-pages) could be clearer about where to set the env var in the non-action local build case... locally? or in an env var on github?)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36146,9 +36146,9 @@
       "dev": true
     },
     "node_modules/thebe-core": {
-      "version": "0.2.3",
-      "resolved": "https://registry.npmjs.org/thebe-core/-/thebe-core-0.2.3.tgz",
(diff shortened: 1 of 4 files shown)


rowanc1:
Tested with @stevejpurves this is up and going again with GithubPages!

PR #138, opened 2023-06-27
 🪧 UI Feedback on Jupyter Status

stevejpurves:
Improvements to UI feedback from the `thebe` events mechanism, mainly to provide better feedback during longer binder connections and when using jupyterlite.

eRFqSlDH45

ZDBmRVJI0q

### DIFF
--- a/packages/jupyter/src/ConnectionStatusTray.tsx
+++ b/packages/jupyter/src/ConnectionStatusTray.tsx
@@ -1,7 +1,7 @@
-import { useEffect, useRef, useState } from 'react';
-import { useThebeLoader, useThebeServer, useThebeSession } from 'thebe-react';
+import { useEffect, useState } from 'react';
+import { useThebeServer, useThebeSession } from 'thebe-react';
 import { useComputeOptions } from './providers';
-import { ThebeEventCb, ThebeEventData, ThebeEventType } from 'thebe-core';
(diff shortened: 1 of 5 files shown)


stevejpurves:
ok the binder message about a bad default ref will go once https://github.com/executablebooks/myst-theme/pull/137 comes in and we are on the latest thebe 👍

rowanc1:
Will test out! UI looks good from the screenshots!

packages/jupyter/src/ConnectionStatusTray.tsx
```diff
@@ -0,0 +1,89 @@
...
+    events.on('status' as ThebeEventType, handleStatus);
+  }, [events]);
+
+  useEffect(() => {
+    if (!thebe) return;
+    if (thebe?.useBinder || thebe?.useJupyterLite) {
+      if (busy || error) {
+        setShow(true);
+      } else if (ready) {
+        setTimeout(
```
rowanc1, line 33: I think if we modify this in the future we should move to using radix Toast:
https://www.radix-ui.com/docs/primitives/components/toast

There is a bunch of accessibility built into that, as well as timeouts and stacking, and it can handle notifications from the full application not just here!

rowanc1 (commented):
Making some modest changes. I think we should adopt radix here pretty quickly @stevejpurves, but that can come in the next PR.

Issue #140, opened 2023-07-03
z-index issues with Thebe controls


rowanc1:
![image](https://github.com/executablebooks/myst-theme/assets/913249/930c70df-8961-427f-98e0-e6bfce41f0d8)

stevejpurves:
the top toolbar we have control over, but the slider themselves are being rendered with ipywidgets styles. This is not limited to the sliders btw, other things in jupyter outputs areas can have varied z-indexes. The toolbar controls were pushed to a higher layer in order to keep them on top of various parts of outputs. pushing the menu itself to a higher z-index isn't an option?

stevejpurves:
consider only lifting the nav and gray background overlay up (z-10) when the menu is open on narrow screens

PR #142, opened 2023-07-04
🕹 Move controls to embed node

rowanc1:
This moves the in-article controls to the top of the embed node and also fixes the z-index issues we talked about in #140. I have also removed the article notebook toolbar, I think we could bring it back, but I didn't really think it was necessary? Linked notebooks still work together!

image

### DIFF
--- a/packages/jupyter/src/embed.tsx
+++ b/packages/jupyter/src/embed.tsx
@@ -1,0 +1,66 @@
+import { SourceFileKind, type GenericNode } from 'myst-common';
+import { useCellExecution } from './execute';
+import {
+  ArticleResetNotebook,
+  ArticleRunNotebook,
+  ArticleStatusBadge,
+} from './controls/ArticleCellControls';
+import { JupyterIcon } from '@scienceicons/react/24/solid';
+import { select } from 'unist-util-select';
+import { useLinkProvider } from '@myst-theme/providers';
+
+function EmbedWithControls({
+  outputKey,
+  children,
+  title = 'Jupyter Notebook',
+  url,
+}: {
+  outputKey: string;
+  children?: React.ReactNode;
+  title?: string;
+  url?: string;
+}) {
+  const { kind } = useCellExecution(outputKey);
+  const Link = useLinkProvider();
+  const showControls = kind === SourceFileKind.Article;
+
+  return (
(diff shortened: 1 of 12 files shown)


styles/hover.css
```diff
@@ -1,6 +1,7 @@
 .hover-card-content {
   animation-duration: 0.6s;
   animation-timing-function: cubic-bezier(0.16, 1, 0.3, 1);
+  z-index: 10;
```
rowanc1, line 4: This is the only real style change for z-index.

packages/site/src/pages/Article.tsx
```diff
@@ -40,7 +40,7 @@ export const ArticlePage = React.memo(function ({ article }: { article: PageLoad
             </div>
           )}
           {canCompute && article.kind === SourceFileKind.Notebook && <NotebookToolbar showLaunch />}
-          {canCompute && article.kind === SourceFileKind.Article && <NotebookToolbar />}
+          {/* {canCompute && article.kind === SourceFileKind.Article && <NotebookToolbar />} */}
```
rowanc1, line 43: I have commented this out, but we may want to bring something like it back.

packages/jupyter/src/embed.tsx
```diff
@@ -0,0 +1,66 @@
...
+import {
+  ArticleResetNotebook,
+  ArticleRunNotebook,
+  ArticleStatusBadge,
+} from './controls/ArticleCellControls';
+import { JupyterIcon } from '@scienceicons/react/24/solid';
+import { select } from 'unist-util-select';
+import { useLinkProvider } from '@myst-theme/providers';
+
+function EmbedWithControls({
```
rowanc1, line 12: I have moved the `embed -> output` node with controls here. This allows us to link to the notebook source.

I am not totally happy with it, but I think it is better than the controls being fully on the right.

rowanc1 (commented):
A few comments @stevejpurves, I think this is complete enough to come in.

packages/site/src/pages/Article.tsx
stevejpurves: i like the idea of making it figure oriented -- all we are losing here is the ability to "run all" in one go

PR #143, opened 2023-07-04
fix/thebe widgets restore

stevejpurves:
https://github.com/executablebooks/myst-theme/assets/1473646/0ac45c7c-574d-4f61-8b95-437b117ae3b2


rowanc1:
Works so well!!

Issue #148, opened 2023-07-04
When Thebe is not enabled, the controls are still active


rowanc1:
For example here:

https://mystmd.org/guide/cross-references#fig-altair-horsepower

PR #162, opened 2023-07-06
🔭 compute controls no longer shown when thebe not enabled

stevejpurves:
addresses #148 and https://github.com/executablebooks/mystmd/issues/461 with some concession to minimally altering the narritive/reading experience, while still keeping a reference link to the source notebook when thebe is not activated.

image

### DIFF
--- a/packages/jupyter/src/embed.tsx
+++ b/packages/jupyter/src/embed.tsx
@@ -20,14 +20,14 @@ function EmbedWithControls({
   title?: string;
   url?: string;
 }) {
-  const { kind } = useCellExecution(outputKey);
+  const { canCompute, kind } = useCellExecution(outputKey);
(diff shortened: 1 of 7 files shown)


packages/jupyter/src/embed.tsx
```diff
@@ -48,8 +48,29 @@ function EmbedWithControls({
...
-      <div className="mt-2">{children}</div>
+        <div className="mt-2">{children}</div>
+      </div>
+    );
+  }
+
+  // light
+  return (
+    <div>
+      <div className="sticky top-[60px] z-[2] w-full backdrop-blur py-1 px-2">
```
rowanc1, line 59: Probably can un-nest the divs and remove the sticky! Will take a look

PR #166, opened 2023-07-06
🪐☀️ Dark mode for Jupyter widgets

rowanc1:
This is minimum viable theme improvements for using thebe in dark mode. I think it does belong here rather than in Thebe? But maybe not @stevejpurves?

image

image


stevejpurves:
Yes best places here.

PR #167, opened 2023-07-07
📜 enable x-scrolling at figure level

stevejpurves:
This change should immediately allow some outputs to be scrollable (e.g. plain text rendered as pre), in combination with [this change in thebe](https://github.com/executablebooks/thebe/pull/662) it should also allow output contents like wide pandas tables to scroll again.

image


stevejpurves:
fixes #153

PR #171, opened 2023-07-07
👩‍💻 Simplifications to the code-copy button

rowanc1:
This adds a few improvements for accessibility when focused.

code-copy

### DIFF
--- a/packages/myst-to-react/src/code.tsx
+++ b/packages/myst-to-react/src/code.tsx
@@ -88,11 +88,7 @@ export function CodeBlock(props: Props) {
       >
         {value}
       </SyntaxHighlighter>
-      {showCopy && (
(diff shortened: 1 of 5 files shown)


packages/myst-to-react/src/code.tsx
```diff
@@ -88,11 +88,7 @@ export function CodeBlock(props: Props) {
       >
         {value}
       </SyntaxHighlighter>
-      {showCopy && (
-        <div className="absolute hidden top-1 right-1 group-hover:block">
-          <CopyIcon text={value} />
-        </div>
```
rowanc1, line 94: This was hiding the component, and inaccessbible to tab. Now this is tab-accessibile!

PR #172, opened 2023-07-11
🎨 Updates for article theme

rowanc1:
- banners
- static build


rowanc1:
NPM is having issues. I think we should go and fix this later if it is actually an issue.

image

PR #174, opened 2023-07-14
🙈 Hide outline and footer links

stevejpurves:
* adds site level `hide_toc`,`hide_outline` and `hide_footer_links` to book theme
* renamed `hide_footer` to `hide_footer_links` in article theme

### DIFF
--- a/packages/site/src/pages/Article.tsx
+++ b/packages/site/src/pages/Article.tsx
@@ -17,7 +17,13 @@ import {
 } from '@myst-theme/jupyter';
 import { FrontmatterBlock } from '@myst-theme/frontmatter';
 
-export const ArticlePage = React.memo(function ({ article }: { article: PageLoader }) {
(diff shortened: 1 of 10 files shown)


packages/site/src/pages/Article.tsx
```diff
@@ -17,7 +17,13 @@ import {
 } from '@myst-theme/jupyter';
 import { FrontmatterBlock } from '@myst-theme/frontmatter';
 
-export const ArticlePage = React.memo(function ({ article }: { article: PageLoader }) {
+export const ArticlePage = React.memo(function ({
+  article,
+  hide_all_footer_links,
+}: {
+  article: PageLoader;
+  hide_all_footer_links?: boolean;
```
stevejpurves, line 25: agonised over this a bit, setting on individual pages might be enough but the `frontmatter.design` code here seems to be prospective i.e. it's not possible to pass those options in on page frontmatter.

PR #181, opened 2023-07-17
‼️ Show errors in embedded notebooks

stevejpurves:
Add a component and logic that will catch and show "off page" notebook errors.

This PR adds an error tray using a jupyter output error to properly display the callback:

image


stevejpurves:
@rowanc1 currently this is displayed at the top of the article. With some more work tracking cell id's etc we could probably get it to display under the figure that triggered it, but perhaps it's better in a toaster/popup that gets the reader's attention but can be dismissed?

rowanc1:
Nice, will take a look!

rowanc1:
I can't get this to work? This is with thebe-lite:

image

stevejpurves:
do you this is is good to go as an increment, and we improve positioning in a separate PR?

rowanc1:
Yep! Just waiting for tests to go green!

PR #188, opened 2023-08-01
🔭 Add space under jupyter outputs

rowanc1:
@kolibril13 this is something that you mentioned, this should improve it slightly!

I am also moving figure --> div as many times outputs are not "figures".


rowanc1:
Theme isn't deployed yet! Sorry for the confusion. This PR is only going to solve (1) not the scrollbars.

PR #202, opened 2023-08-09
🐛 Guard outputs and fix embed children

rowanc1:
See https://github.com/orgs/executablebooks/discussions/1062#discussioncomment-6682400

### DIFF
--- a/packages/jupyter/src/embed.tsx
+++ b/packages/jupyter/src/embed.tsx
@@ -1,5 +1,4 @@
 import { type GenericNode } from 'myst-common';
-import { JupyterOutput } from './output';
 import { OutputDecoration } from './decoration';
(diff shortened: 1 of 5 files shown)


packages/jupyter/src/embed.tsx
```diff
@@ -13,12 +12,7 @@ export function Embed({ node }: { node: GenericNode }) {
       title={node.source?.title}
       url={node.source?.url}
     >
-      <JupyterOutput
-        outputId={output.id}
-        identifier={output.identifier}
-        align={node.align}
-        data={node.data}
-      />
+      <MyST ast={node.children} />
```
rowanc1, line 15: We were trying to render code + output as an output here. 💥

PR #213, opened 2023-08-11
🎨 Update icon sizes to be explicit

rowanc1:
This should be a no-op for the most part. However, when styles fail to load the experience is significantly better:

image

Before it was a giant icon as the first thing that you see.

### DIFF
--- a/packages/jupyter/src/controls/Spinner.tsx
+++ b/packages/jupyter/src/controls/Spinner.tsx
@@ -3,7 +3,9 @@ export function Spinner({ size }: { size: number }) {
     <div role="status">
       <svg
         aria-hidden="true"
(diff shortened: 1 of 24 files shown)


packages/jupyter/src/controls/Spinner.tsx
```diff
@@ -3,7 +3,9 @@ export function Spinner({ size }: { size: number }) {
     <div role="status">
       <svg
         aria-hidden="true"
-        className={`w-[${size}px] h-[${size}px] mr-2 text-gray-200 animate-spin dark:text-gray-600 fill-green-600`}
+        width={size}
+        height={size}
```
rowanc1, line 7: @stevejpurves, just FYI you can't have dynamic tailwind styles!

PR #224, opened 2023-08-21
📌 run notebooks on the correct path

stevejpurves:
The location of a notebook relative to the project root folder is important and we assume that code in the notebook is written relative to the project root.

This change uses the `location` field exposed in https://github.com/executablebooks/mystmd/pull/564 and not properly passed to `thebe` using the `kernelOptions.path` in place of `kernelOptions.name` (see https://github.com/executablebooks/thebe/pull/677)


packages/jupyter/src/execute/provider.tsx
```diff
@@ -49,11 +50,11 @@ function useScopeNavigate({
       return;
     }
 
-    const computables: Computable[] = selectAll('container > embed', mdast).map((node: any) => {
+    const computables: Computable[] = selectAll('container > output', mdast).map((node: any) => {
       const { key, label, source } = node;
       const output = selectAll('output', node);
-      if (output.length === 0) console.error(`embed must have exactly one output ${key}`);
-      if (output.length > 1) console.warn(`embed has more than one output block ${key}}`);
+      if (output.length === 0) console.error(`container must have exactly one output ${key}`);
```
fwkoch, line 56: We won't hit this any more, since we are only selecting containers with outputs now, right? Do we still want to error on other containers without outputs?

packages/jupyter/src/execute/types.ts
```diff
@@ -26,7 +26,8 @@ export interface ExecuteScopeState {
       slug: string;
       kind: SourceFileKind;
       computable: boolean;
-      dependencies: Dependency[];
+      location: string;
+      dependencies: (Dependency & { location: string })[]; // TODO remove location after release of myst-spec-ext
```
fwkoch, line 30: 👍

packages/jupyter/src/execute/leaf.tsx
```diff
@@ -119,22 +121,30 @@ export function SessionStarter({
...
       console.debug('Jupyter: running sessions', sessions);
-      const path = `/${pageSlug}-${notebookSlug}.ipynb`;
+
+      // the location gives us the correct path for the notebook including it's filename
+      // we need to replace the filename with one based on the page slug and notebook slug
+      // in order to allow for multiple independent sessions of the same notebook
+      let path = `/${pageSlug}-${notebookSlug}.ipynb`;
+      const match = location.match(/(.*)\/.*.ipynb$/);
+      if (match) {
+        path = `${match[1]}/${pageSlug}-${notebookSlug}.ipynb`;
```
fwkoch, line 135: Is this line pretty much the meat of the PR?

stevejpurves: yep! I think path could be `${match[1]}` but convention for jupyterlab is for it to include the filename, so thebe sticks with that.

packages/jupyter/src/execute/provider.tsx
stevejpurves: ok seems like we need to consider both the `container > output` and `embed > output` cases here

Issue #229, opened 2023-08-29
TOC is not correctly in tab order for keyboard access


stevejpurves:
### Describe the bug

**context**
When I do initially load the page and navigate via the keyboard (tab)

**expectation**
I expected to be able to tab through all navigation elements before reaching the article frontmatter an content

**bug**
But instead I the TOC is reached the end of the article, or by SHIFT-TAB fro the top nav which is not a natural flow.

```console
$ jupyter-book build mybook
ERROR ...
```

**problem**
This is a problem for people using keyboard navigation as they cannot eaily find the TOC navigation which should be early.

### Reproduce the bug

![eTh2isjlgE](https://github.com/executablebooks/myst-theme/assets/1473646/29492262-2e02-489c-adc7-89445dc44a69)

PR #230, opened 2023-08-29
⌨️ fixing tab ordering for TOC

stevejpurves:
Addresses #229 

8cIJVkbmu5


fwkoch (approved):
These seem like pretty safe ordering changes? I don't quite know the implications of un-nesting `TopNav` and `TabStateProvider` outside of `Navigation`, but my assumption is the nesting doesn't really matter for these components.

stevejpurves:
yes, nesting in nav seems completely unnecessary

Issue #231, opened 2023-08-29
MyST book and article theme have no Skip Navigation feature


stevejpurves:
### Describe the bug

**context**
When I do load the page and press tab there is no way to skip to the main content

**expectation**
I expected a skip navigation link(s) to be available allowing a keyboard user to skip all of the navigation and arrive at the main content.

**bug**
But instead there is no skip navigation link and a kb user needs ot tab through 3 sets of menus (in book theme) to arrive at the main content.

```console
$ jupyter-book build mybook
ERROR ...
```

**problem**
This is a problem for navigating using the kb and screen readers as it takes many presses to arrive at the main content

### Reproduce the bug

1.load a page
2. press `tab`

PR #232, opened 2023-08-29
⌨️ Skip-to-content keyboard accessibility

stevejpurves:
Addresses #231 

This can still be improved a lot but fulfills the basic need, a keyboard focus jumps ahead of nav and frontmatter.

lazSNANApT

bu49m8EYV0

There are some remaining issues:

* all themes - the page does not scroll as desired on use of the links on first page load, maybe fighting with other scroll restore behaviour?
* article theme - skip to supporting documents would be a good addition

### DIFF
--- a/packages/site/src/components/Abstract.tsx
+++ b/packages/site/src/components/Abstract.tsx
@@ -1,14 +1,16 @@
 import type { GenericParent } from 'myst-common';
 import { ContentBlocks } from './ContentBlocks';
 import classNames from 'classnames';
+import { HashLink } from 'myst-to-react';
 
 export function Abstract({ content }: { content: GenericParent }) {
   if (!content) return <div className="hidden" aria-label="this article has no abstract" />;
   return (
     <>
-      <span className="mb-3 font-semibold" tabIndex={0}>
+      <h2 id="abstract" className="mb-3 text-base font-semibold group">
         Abstract
-      </span>
+        <HashLink id="abstract" title="Link to Abstract" hover className="ml-2" />
+      </h2>
       <div className="px-6 py-1 mb-3 rounded-sm bg-slate-50 dark:bg-slate-800">
         <ContentBlocks mdast={content} className="col-body" />
       </div>
@@ -26,7 +28,7 @@ export function Keywords({
   if (hideKeywords || !keywords || keywords.length === 0)
     return <div className="hidden" aria-label="this article has no keywords" />;
   return (
-    <div className="mb-10" tabIndex={0}>
+    <div className="mb-10 group">
       <span className="mr-2 font-semibold">Keywords:</span>
       {keywords.map((k, i) => (
         <span
@@ -38,6 +40,7 @@ export function Keywords({
           {k}
         </span>
       ))}
+      <HashLink id="keywords" title="Link to Keywords" hover className="ml-2" />
     </div>
   );
 }
--- a/packages/site/src/components/Abstract.tsx
+++ b/packages/site/src/components/Abstract.tsx
@@ -3,10 +3,12 @@ import { ContentBlocks } from './ContentBlocks';
 import classNames from 'classnames';
 
 export function Abstract({ content }: { content: GenericParent }) {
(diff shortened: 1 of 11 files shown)


fwkoch (approved):
This is great - the PR makes sense to me in general - add hidden navigational links and targets with relevant aria labels.

stevejpurves:
the only thing I am not sure on is added in the `tabindex` to non clickable items (abstract, keywords) but I think the focus and visual cue is useful. I decided to leave that in pending more learning / feedback on accessibility best practice

packages/site/src/components/SkipToArticle.tsx
```diff
@@ -0,0 +1,52 @@
...
+  article?: boolean;
+}) {
+  const fm = 'skip-to-frontmatter';
+  const art = 'skip-to-article';
+
+  const frontmatterHander = useCallback(() => makeSkipClickHander(fm), [frontmatter]);
+  const articleHandler = useCallback(() => makeSkipClickHander(art), [article]);
+  return (
+    <div
+      className="fixed top-1 left-1 h-[0px] w-[0px] focus-within:z-40 focus-within:h-auto focus-within:w-auto bg-white overflow-hidden focus-within:p-2 focus-within:ring-1"
```
stevejpurves, line 29: need to look at using `focus-visible` here so that these focus states only apply to keyboard navigation

rowanc1: You can use the `sr-only` class here! That picks up a bunch of these I think (the top/h/w at least)?

packages/site/src/components/Abstract.tsx
```diff
@@ -3,10 +3,12 @@ import { ContentBlocks } from './ContentBlocks';
...
 
 export function Abstract({ content }: { content: GenericParent }) {
-  if (!content) return null;
+  if (!content) return <div className="hidden" aria-label="this article has no abstract" />;
   return (
     <>
-      <span className="mb-3 font-semibold">Abstract</span>
+      <span className="mb-3 font-semibold" tabIndex={0}>
+        Abstract
+      </span>
```
rowanc1, line 11: There is a good walkthrough video of ar5iv here that argues for abstract to be a (styled) h2 element.

https://github.com/brucemiller/LaTeXML/issues/2066#issuecomment-1696389753

By Patrick Smyth from the Iota School on YouTube:
https://youtu.be/bHM6mes51nE

Was worth a watch for me, I think this PR is the major thing to implement.

rowanc1:
For the abstract/keywords anchors, we should use the `HashLink` component to make them clickable and set the URL etc.

packages/site/src/components/Abstract.tsx
rowanc1: I changed to an h2!

packages/site/src/components/SkipToArticle.tsx
rowanc1: Ok, I think in this case yours is better because we are changing things!

stevejpurves: skip nav is for keyboard users rather than screen readers only

stevejpurves:
> For the abstract/keywords anchors, we should use the `HashLink` component to make them clickable and set the URL etc.

overall I didn't like the div anchor approach here but couldn't find a better / common way to do this

packages/site/src/components/Abstract.tsx
```diff
@@ -1,14 +1,16 @@
...
 import { ContentBlocks } from './ContentBlocks';
 import classNames from 'classnames';
+import { HashLink } from 'myst-to-react';
 
 export function Abstract({ content }: { content: GenericParent }) {
   if (!content) return <div className="hidden" aria-label="this article has no abstract" />;
   return (
     <>
-      <span className="mb-3 font-semibold" tabIndex={0}>
+      <h2 id="abstract" className="mb-3 text-base font-semibold group">
```
stevejpurves, line 10: 👍🏼

PR #239, opened 2023-09-27
launch binder for articles

stevejpurves:
* uses a site's `thebe/jupyter` settings
* add a "Launch Binder" button to the 
* once a binder server is available, "Launch Binder" will change to "Open in Binder" and clicking that will open a new tab, with a new instance of jupyterlab but with 

This is dependent on recent changes to thebe enabling different "providers" to be configured.


themes/article/app/routes/$.tsx
```diff
@@ -133,6 +140,11 @@ export function Article({
   const tree = copyNode(article.mdast);
   const abstract = extractPart(tree, 'abstract');
   const { title, subtitle } = article.frontmatter;
+
+  const thebe = useThebeOptions();
+  const canCompute = !!thebe && (article.frontmatter as any)?.thebe !== false;
+  // TODO in lieu of extended frontmatter or theme options
+  const enable_notebook_toolbar = false;
```
stevejpurves, line 147: @rowanc1 do we have a way of bringing in theme options yet for the web theme? or any thoughts on where to do that?

I am wondering if there is already a mechanism there via the site config + template options?

packages/jupyter/src/providers.tsx
```diff
@@ -27,7 +27,8 @@ function makeThebeOptions(
     binderBadgeUrl,
   );
 
-  const options = thebeFrontmatter ? optionsFromFrontmatter : undefined;
+  const options = optionsOverrideFn(thebeFrontmatter ? optionsFromFrontmatter : undefined);
```
stevejpurves, line 30: a parent theme is able to fully manipulate the options, guarding against undesirable options (e..g connecting to a foreign binder) coming in from the myst site, or just being able to patch certain options appropriately for where there parent theme is used/deployed.

stevejpurves:
Confirmed also that from the book-theme when we hit the existing "Launch in Binder" link on the Notebook Toolbar, this successfully opens the jupyterlab on the server.

What we are missing in the book-theme experience is a place to add the Launch in Binder link.
We also don't have a location for the other downloadable assets.

Should we be considering adding something to the header section of the index article?
image



or some type of top level nav dropdown?

image

rowanc1:
Couldn't do the rebase well with so many conflicts. The merge I think is correct!

Se should be sure to squash-merge this when ready.

PR #242, opened 2023-10-02
📦 Move to ESM only

rowanc1:
Need to
- [x] figure out `Ansi`
- [x] update package.json
- [x] add changesets


packages/frontmatter/src/downloads.tsx
```diff
@@ -1,6 +1,5 @@
 import { Menu } from '@headlessui/react';
-import DocumentIcon from '@heroicons/react/24/outline/DocumentIcon';
-import ArrowDownTrayIcon from '@heroicons/react/24/outline/ArrowDownTrayIcon';
+import { DocumentIcon, ArrowDownTrayIcon } from '@heroicons/react/24/outline';
```
rowanc1, line 2: This is the one potential problem, we may be bringing in all icons, however, I think that ESM should fix this.

rowanc1: I can confirm that this is fixed with ESM:
image

rowanc1:
This is going to cause conflicts with your branch @stevejpurves, but I have a few more questions. I will fix it up.

PR #245, opened 2023-10-04
Add missing dependency for myst-to-react

Carreau:
I think ? I'm not a Js person, but I have the impression this is needed.

### DIFF
--- a/.changeset/orange-papayas-chew.md
+++ b/.changeset/orange-papayas-chew.md
@@ -1,0 +1,6 @@
+---
+'myst-to-react': patch
+'@myst-theme/site': patch
+---
+
+Update missing dependencies
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,7 +22,7 @@
         "@types/react": "^18.2.24",
         "@types/react-dom": "^18.2.8",
         "concurrently": "^8.2.0",
-        "eslint-config-curvenote": "^0.0.3",
+        "eslint-config-curvenote": "^0.0.4",
         "npm-run-all": "^4.1.5",
         "patch-package": "^6.5.1",
         "prettier": "latest",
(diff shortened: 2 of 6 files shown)


rowanc1:
Thank you - our dependencies in the mono-repo fall back to the entire repo, so we sometimes miss these. After some investigation I am going to turn this on so this doesn't happen again:
https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-extraneous-dependencies.md

I will update and fix this PR. Thanks for reporting.

Is there any help/pointers you need in getting started?

Carreau:
I'll trust you, the js ecosytem is stull a mystery for me. I'm still trying to understand a bit how react/npm and all that works.

rowanc1:
Let me know if I can be of any help!

Carreau:
I'll ask maybe dumb questions/notes from trying to get things to work.

I tried to make myst-to-react work with https://parceljs.org/getting-started/webapp/

It doesn't work and complain that `@parcel/resolver-default: Cannot load file './index' from module`.

This appear to happen only for all myst-related packages. 
From trial and error I looks like adding `"main": "./dist/index.js"`, in `node_modules/<packages>/packages.json` works. And I find the docs about main [here](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#main).

Now I don't know anything about the js ecosytem, and I guess parcel is maybe not something widely us, but maybe `main` should be set on all the packages ?

rowanc1:
Hi @Carreau -- I have put together a demo here:

**Repo** https://github.com/rowanc1/myst-papyri
**Video:** https://www.loom.com/share/6e3d652883de42c985aa5127f57befdb

That demo is using `create-react-app` which uses webpack I believe (similar to JupyterLab). It looks like there may be some issues with ESM and parcel (a similar upgrade from python 2.7 -> 3, and just the worst). 

https://github.com/parcel-bundler/parcel/issues/7823

It looks like in the last few months they did (I think?) add support for ESM packages ([blog](https://parceljs.org/blog/v2-9-0/#new-resolver)), are you using that version, and maybe add this?

```json
{
  "@parcel/resolver-default": {
    "packageExports": true
  }
}
```

A few folks on that open thread suggesting esbuild, which also just generally works in my experience. I have put together a demo here for esbuild:

**Repo:** https://github.com/rowanc1/myst-playground
**Video:** https://www.loom.com/share/33957fa64d5340c389ca6e25f6d357f8

If parcel is a hard requirement for you, I can do some more digging. As a quick comparison here are some bundlers:
https://npmtrends.com/esbuild-vs-next-vs-parcel-vs-webpack

image

For the `main` in the json, I am following the advice here: https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c#how-can-i-make-my-typescript-project-output-esm
Which says to remove it. 🙈 We have recently moved fully over to ESM in all of the packages (a similar effort is happening in JupyterLab at the moment I believe). Hopefully the `packageExports` config flag in parcel will work for you!

Carreau:
Thanks, I did manage to make it work with create-react-app yesterday but using [this so answer](https://stackoverflow.com/questions/70964723/webpack-5-in-ceate-react-app-cant-resolve-not-fully-specified-routes/75109686#75109686). 

> It looks like in the last few months they did (I think?) add support for ESM packages ([blog](https://parceljs.org/blog/v2-9-0/#new-resolver)), are you using that version, and maybe add this?

Yep that works. I guess I don't know ESM and all these stuff so it made it hard for me to find the info. 

I just tried parcel as the person next to me at my coworking space was just suggesting it , and I just reported issues, you probably now best. Npm/yarn whatever is fine, I just need to get used to it.

THanks !

PR #248, opened 2023-10-09
🎆 Support inline property on image nodes

fwkoch:
This consumes the image `inline: true` property added here https://github.com/executablebooks/mystmd/pull/641 and adds `style="display:inline"` if that property is present on image nodes.

### DIFF
--- a/.changeset/clean-frogs-fail.md
+++ b/.changeset/clean-frogs-fail.md
@@ -1,0 +1,5 @@
+---
+'myst-to-react': patch
+---
+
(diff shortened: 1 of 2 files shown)


stevejpurves:
@rowanc1 @fwkoch what are the next steps on this one?

agoose77:
We should try and close this! :)

rowanc1:
Ultimately fixed here:
https://github.com/jupyter-book/myst-theme/pull/710

PR #261, opened 2023-11-10
👩‍🎨 Nest design under options

fwkoch:
I'm not sure how `design` is coming into article frontmatter, so I'm not sure if these changes are correct. However, with the introduction of `siteConfig.options`, `siteConfig.design` is now moved to `siteConfig.options.design`...

### DIFF
--- a/themes/book/app/routes/$.tsx
+++ b/themes/book/app/routes/$.tsx
@@ -96,9 +96,9 @@ export default function Page() {
   const { container, outline } = useOutlineHeight();
   const top = useThemeTop();
   const article = useLoaderData<PageLoader>() as PageLoader;
-  const pageDesign: BookThemeTemplateOptions = (article.frontmatter as any)?.options?.design ?? {};
+  const pageDesign: BookThemeTemplateOptions = (article.frontmatter as any)?.options ?? {};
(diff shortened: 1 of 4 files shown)


themes/book/app/routes/$.tsx
```diff
@@ -96,9 +96,9 @@ export default function Page() {
   const { container, outline } = useOutlineHeight();
   const top = useThemeTop();
   const article = useLoaderData<PageLoader>() as PageLoader;
-  const pageDesign: BookThemeTemplateOptions = (article.frontmatter as any)?.design ?? {};
+  const pageDesign: BookThemeTemplateOptions = (article.frontmatter as any)?.options?.design ?? {};
   const siteDesign: BookThemeTemplateOptions =
-    (useSiteManifest() as SiteManifest & BookThemeTemplateOptions) ?? {};
```
fwkoch, line 101: I would have expected this to be `(...).design` - maybe there is something else going on here that isn't just simple frontmatter reading.

rowanc1:
I think these are just in options and we remove "design".

Can you update for that!?

fwkoch:
Oh, nice, yes that makes sense, and they are already options in the theme `template.yml`.

PR #262, opened 2023-11-13
💻  Notebook toolbar always on but not functional

### LINKED ISSUE #255: Thebe notebook bar shown on every notebook
In the book theme we are showing thebe on every notebook regardless of if it is on in the theme or now. I think that this is because we are looking to the article frontmatter (where thebe will *never* be defined):
https://github.com/executablebooks/myst-theme/blob/toc/nest/packages/site/src/pages/Article.tsx#L37

I believe that we have also taken out `thebe !== false`, so this is also no longer in the frontmatter.
https://github.com/executablebooks/myst-theme/blob/toc/nest/packages/jupyter/src/providers.tsx#L82

I think the default for thebe should be off, with the option to explicitly turn it on. But for sure putting `thebe: false` in the project frontmatter should remove the toolbars.

stevejpurves:
👍 canCompute now returns false on falsy `thebe|jupyter` key

Fixes #255


rowanc1:
Will take a look now!

PR #269, opened 2023-11-14
📦 Upgrade MyST dependencies

rowanc1:


packages/jupyter/src/execute/provider.tsx
```diff
@@ -32,7 +31,6 @@ type ArticleContents = {
   mdast: GenericParent;
   location?: string;
   dependencies?: Dependency[];
-  frontmatter: { thebe?: boolean | Thebe };
```
rowanc1, line 35: @stevejpurves have removed this type here. We don't use it, and it is now no longer defined on the page - only the project.

packages/jupyter/src/utils.ts
```diff
@@ -73,46 +68,15 @@ export function thebeFrontmatterToOptions(
...
 
-  // handle thebe.local.*
-  // as local OVERRIDES other binder and server settings, handle these first
-  // TODO need to expose NODE_ENV somehow via a loader
-  let NODE_ENV = 'development';
-  if (typeof window !== 'undefined') {
-    NODE_ENV = (window as any).NODE_ENV;
-  }
-  if (NODE_ENV !== 'production' && local) {
-    if (isObject(local)) {
```
rowanc1, line 101: @stevejpurves -- removing `JupyterLocalOptions`.

- See executablebooks/mystmd#750

rowanc1 (commented):
Bumping packages and fixing up a few types.

PR #288, opened 2023-11-27
🆙 update thebe

stevejpurves:
update thebe to the latest version, which in turn updates the latest jupyterlite and jupyterlab dependencies.

There are large lockfile changes here mainly due to the thebe dependencies. Thebe core and lite are dynamically loaded but thebe-react still brings in these dependencies.


stevejpurves:
tested on book and article theme

PR #290, opened 2023-11-28
🕺split out compute options provider and add feature flags

stevejpurves:
Bundling the ThebeOptions provider in beside the ThebeServer really hurts flexiblity. The ThebeOptions could be much further up the render tree and early analysis of the thebe option can be used to setup the site, e.g. like knowing whether to load thebe lite or not.

Providers are split out here and we also add flags to enable different parts of the compute ui separately.


stevejpurves:
Once this comes in we're ready for a theme release which will enable lite, latest juptyer components and should improve a bunch of things

PR #295, opened 2023-11-30
Make Default MyST to act renderer configurable

### LINKED ISSUE #294: Ability to set a custom "DefaultComponent" for Myst-to-react renderer
### Context

It seem that by default if there is no node type for a renderer,  Myst-to-react will fallback to DefaultComponent (see MyST.tsx). The default is to wrap in div/span depending on wether the component contains children (or not), and recurse.

Unfortunately this does not seem easy to configure (it might be doable, but I'm nto familiar enough with react/js to figure out how)

### Proposal

Make the DefaultComponent hook configurable, either a separate configuration, or maybe a specific type in renderers ? 

```
renderers[node.type] ?? renderers['default'] ?? DefaultComponent
```

Happy to do it, but i'm not yet versed in JS/TSX, so you would have to bear with my crappy code.

### Tasks and updates

- [ ] decide if this need to be done
- [ ] how to have it done
- [ ] document

Carreau:
My attempt at trying to fix #294 and write some docs.

I'm still figuring out JS/TS, so this is untested as I don't know yet how to use a dev version locally.

### DIFF
--- a/packages/myst-to-react/src/MyST.tsx
+++ b/packages/myst-to-react/src/MyST.tsx
@@ -14,7 +14,7 @@ export function MyST({ ast }: { ast?: GenericNode | GenericNode[] }) {
   const renderers = useNodeRenderers();
   if (!ast || ast.length === 0) return null;
   if (!Array.isArray(ast)) {
-    const Component = renderers[ast.type] ?? renderers['default'] ?? DefaultComponent;
+    const Component = renderers[ast.type] ?? renderers['DefaultComponent'] ?? DefaultComponent;
     return <Component key={ast.key} node={ast} />;
   }
   return (
--- a/packages/myst-to-react/src/MyST.tsx
+++ b/packages/myst-to-react/src/MyST.tsx
@@ -14,7 +14,7 @@ export function MyST({ ast }: { ast?: GenericNode | GenericNode[] }) {
   const renderers = useNodeRenderers();
   if (!ast || ast.length === 0) return null;
   if (!Array.isArray(ast)) {
-    const Component = renderers[ast.type] ?? DefaultComponent;
+    const Component = renderers[ast.type] ?? renderers['default'] ?? DefaultComponent;
     return <Component key={ast.key} node={ast} />;
   }
   return (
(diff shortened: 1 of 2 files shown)


rowanc1 (changes requested):
Looks good. Minor hesitation with using the key `default` here as that is a reserved word in javascript -- I think it is ok as long as you don't name your component that though!

If you could look at the formatting change for the doc string + lang, then we can get this in and do a release later today! (we also some changes coming with thebe from @stevejpurves).

rowanc1:
Actually, do you think we can rename this to `DefaultComponent` in the renderer dict?

image

Using `default` as a const name (raises all sorts of weird errors) -- I know that I have been in a debug loop on this before and finally realized that I couldn't name something `default`.

Carreau:
Thanks, 

> Minor hesitation with using the key `default` here as that is a reserved word in javascript

Yeah, that's what I figured, and that it can be it's own provider, I'm just not the good person to judge the technical merits of one vs the other.

packages/myst-to-react/src/MyST.tsx
```diff
@@ -14,7 +14,7 @@ export function MyST({ ast }: { ast?: GenericNode | GenericNode[] }) {
   const renderers = useNodeRenderers();
   if (!ast || ast.length === 0) return null;
   if (!Array.isArray(ast)) {
-    const Component = renderers[ast.type] ?? DefaultComponent;
+    const Component = renderers[ast.type] ?? renderers['default'] ?? DefaultComponent;
```
Carreau, line 17:
```suggestion
    const Component = renderers[ast.type] ?? renderers['DefaultComponent'] ?? DefaultComponent;
```

rowanc1 (commented):
I think this is it for changing the name!?

rowanc1 (approved):
Thank you! Will see if linting passes and then get this in!

Carreau:
Thanks, I don't know if you want to add a warning to use this with caution, as technically you don't need it, it just make it easier to prototype.

rowanc1:
Considering it a real and supported feature now. :) I think this is good anyways to have and is helpful beyond debugging as well.

Thinking back to something you said, you could now raise an error or log in that default component (maybe that is what you had in mind)!

Carreau:
> Thinking back to something you said, you could now raise an error or log in that default component (maybe that is what you had in mind)!

Yes, among other, right now, I just want to wrap some thing in a colored border to say "not completely implemented", right now it was just rendering children.

PR #297, opened 2023-12-01
Also dispatch to custom component if node is an array.

Carreau:
I just came across that while re-reading the code. I think that's what we meant yesterday.

Also, that make me think that maybe we should revisit that, and have 

```
if (!Array.isArray(ast)){
  ast = [ast]
}
/// unconditionally call the map code ?
```

That might avoid code duplication ?

### DIFF
--- a/packages/myst-to-react/src/MyST.tsx
+++ b/packages/myst-to-react/src/MyST.tsx
@@ -20,7 +20,7 @@ export function MyST({ ast }: { ast?: GenericNode | GenericNode[] }) {
   return (
     <>
       {ast?.map((node) => {
-        const Component = renderers[node.type] ?? DefaultComponent;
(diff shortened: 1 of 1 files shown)


stevejpurves:
@Carreau sorry we left this one hanging!
Is there a test case for this one - or from you comment, is this not the implementation we want?

Carreau:
Oh, I need to re-dive into this.

I think the patch we did before works only when you passed a single node, not when you pass a list of node.

agoose77:
@Carreau I'm going to close this (I've not consulted the other team members) as it looks like it needs more work, and I'm trying to get on top of our PRs. Please do re-open it if you think it should remain open :)

Carreau:
Yeah, sorry, thanks.  I haven't been able to work on that for the past few month.

PR #309, opened 2024-01-09
feat: support the align attribute in tablecell

agoose77:
Hey @fwkoch, I pinged you for review of this because I need to be better at pinging "not rowan"! Let me know if I should instead ask Rowan for review on this!

This PR simply passes through the `align` attribute of `TableCell` to the underlying `td`/`th` element.

As a PR in myst-theme, this doesn't address the problem of alignment in HTML / LaTeX exports.

### DIFF
--- a/packages/myst-to-react/src/basic.tsx
+++ b/packages/myst-to-react/src/basic.tsx
@@ -282,14 +282,19 @@ const BASIC_RENDERERS: BasicNodeRenderers = {
       rowSpan: ifGreaterThanOne(node.rowspan),
       colSpan: ifGreaterThanOne(node.colspan),
     };
+    const align = {
+      'text-left': node.align === 'left',
+      'text-right': node.align === 'right',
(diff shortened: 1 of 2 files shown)


agoose77:
@fwkoch would you mind giving this an eye? :)

packages/myst-to-react/src/basic.tsx
```diff
@@ -284,12 +284,12 @@ const BASIC_RENDERERS: BasicNodeRenderers = {
     };
     if (node.header)
       return (
-        <th className={node.class} style={node.style} {...attrs}>
+        <th className={node.class} style={node.style} align={node.align} {...attrs}>
```
agoose77, line 287: We should use tailwind to do this.

stevejpurves:
@agoose77 worked with my local test case, if you can check and test with yours -- let's merge and get it into the imminent theme release

PR #314, opened 2024-02-12
🎨 Write class name to admonition

agoose77:
Fixes executablebooks/jupyterlab-myst#201 by exposing classname to DOM.

### DIFF
--- a/packages/myst-to-react/src/admonitions.tsx
+++ b/packages/myst-to-react/src/admonitions.tsx
@@ -150,6 +150,7 @@ export function Admonition({
   dropdown,
   children,
   hideIcon,
+  classes
 }: {
   title?: React.ReactNode;
   color?: Color;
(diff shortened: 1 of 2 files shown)


packages/myst-to-react/src/admonitions.tsx
```diff
@@ -150,6 +150,7 @@ export function Admonition({
   dropdown,
   children,
   hideIcon,
+  classes
```
agoose77, line 153:
```suggestion
  className
```

packages/myst-to-react/src/admonitions.tsx
```diff
@@ -247,6 +250,7 @@ export const AdmonitionRenderer: NodeRenderer<AdmonitionSpec> = ({ node }) => {
       dropdown={isDropdown}
       simple={isSimple}
       hideIcon={hideIcon}
+      className={classNames(classes)}
```
agoose77, line 253:
```suggestion
      className={classNames(className)}
```

rowanc1:
Nice - changes look good. Can you add a changeset and we can get this in and released!

PR #317, opened 2024-02-20
🔧 added official mimetype for svg when filtering output bundles

stevejpurves:
This fixes https://github.com/executablebooks/mystmd/issues/916 in local testing

`image/svg-xml` is the correct and offical (IANA) mimetype identifier for SVG images, so we should be filtering based on that. 

`image/svg` is not an official mimetype at all, but I've left it in place - hesistating here in case the original code was covering some edge case that's out there ins the wild.

Happy to remove it though if that's the concensus.

### DIFF
--- a/packages/jupyter/src/convertImages.ts
+++ b/packages/jupyter/src/convertImages.ts
@@ -34,7 +34,6 @@ export async function fetchAndEncodeOutputImages(
     outputs.map(async (output) => {
       if (!('data' in output)) return output;
 
-      console.log('output', output);
       const imageMimetypes = Object.keys(output.data as MinifiedMimeBundle).filter(
         (mimetype) =>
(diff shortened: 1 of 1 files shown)


stevejpurves:
there are also maybe further issues here. If we look at the ggplot output we have various image types all base64 encoded 

image

but the function `requestImageAsBase64String` only encodes to base64 `image/png` data urls (https://github.com/executablebooks/myst-theme/blob/91d884fbd3c448ec1263c5412e284f078ae75b49/packages/jupyter/src/convertImages.ts#L20) could this mean that we end up with incorrectly encoded for images when these have been minified and are fetched? (e.g. we end up with an object with a `content_type` of `image/jpeg` but `content` that is encoded as an `image/png`?)

PR #318, opened 2024-02-26
interim support for plotly

stevejpurves:
Plotly plots are not working in `myst-theme` as the rendermime factory is not being registered.
Recent changes to plotly (as compared to pre-juptyerlab 4 extensions) make it more tricky to bring in the plotly js renderer from the extension package.

For anything more than a short term fix here we need to move the myst-theme towards making proper use of the jupyterlab extension mechanism via federated modules, so that we can configure in myst/environment which extensions to load and have those included in the website build appropriately. As discussed with @agoose77 @rowanc1 

As a short term solution, this PR makes a small patch to the plotly jupyterlab extension module in order to load this dynamically.

In passive mode, the module is only loaded when a plotly mimetype is seen in an `IOutput[]`, whereas in active mode (i.e. after connecting to a kernel) we always load the plotly library.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2433,6 +2433,22 @@
         "url": "https://github.com/prettier/prettier?sponsor=1"
       }
     },
+    "node_modules/@choojs/findup": {
+      "version": "0.2.1",
+      "resolved": "https://registry.npmjs.org/@choojs/findup/-/findup-0.2.1.tgz",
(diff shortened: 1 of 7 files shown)


stevejpurves:
note the alternative to this is to attempt to load the plotly scripts from the notebook outputs onto page along with supporting div tags and have the scripts execute, we'd need to do this reliably within react on each page load. That feels like it's work in the wrong direction though (i.e. no loading the extension module)

PR #323, opened 2024-03-07
🛠 fix notebook build process

### LINKED ISSUE #322: fatal error when executing a notebook
### Describe the bug

**context**
When I press the "power" button on a Notebook a fatal error orccurs. Note that interactive figure in markdown based pages work ok.

**expectation**
I expected the notebook to connect to a session and become executable.

**bug**
But instead I get:

image

**problem**
This is a problem as notebook execution is broken and deployment of the latest myst-theme is blocked.

### Reproduce the bug

Run myst-theme in development connecting to a local server

### List your environment

myst-theme 0.6.0

stevejpurves:
Recent changes to plotly introduced a problem for noteobok based pages at execution time, this PR fixes that ensuring that plotly is loaded before the notebook is deemed built. This is due to the fact that we expect a rendermime registry to be constructed during the notebook build phase, which in turn needs to wait on plotly.

fixes #322

### DIFF
--- a/packages/jupyter/src/plotly.ts
+++ b/packages/jupyter/src/plotly.ts
@@ -30,9 +30,9 @@ export function isPlotly(outputs: IOutput[]) {
 }
 
 export function usePlotlyPassively(rendermime: IRenderMimeRegistry, outputs: IOutput[]) {
-  const [loaded, setLoaded] = useState(false);
-
   const isPlotlyOutput = isPlotly(outputs);
+  // skip loading for non plotly outputs
+  const [loaded, setLoaded] = useState(!isPlotlyOutput);
 
   useEffect(() => {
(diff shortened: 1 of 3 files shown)


packages/jupyter/src/plotly.ts
```diff
@@ -30,9 +30,9 @@ export function isPlotly(outputs: IOutput[]) {
 }
 
 export function usePlotlyPassively(rendermime: IRenderMimeRegistry, outputs: IOutput[]) {
-  const [loaded, setLoaded] = useState(false);
-
   const isPlotlyOutput = isPlotly(outputs);
```
stevejpurves, line 33: fixed additional issue found during testing, also introduced by the recent plotly change

PR #325, opened 2024-03-07
🏷 Pull frontmatter from proper key

agoose77:
In executablebooks/myst-theme#332, @choldgraf observed that opengraph metadata is missing from the static MyST sites. It turns out it's also missing from the dynamic routes too.

I believe the root cause is fixed by this PR, though I have not tested the static export!

Fixes executablebooks/myst-theme#332

### DIFF
--- a/packages/common/src/utils.ts
+++ b/packages/common/src/utils.ts
@@ -17,8 +17,9 @@ export function getProject(
   projectSlug?: string,
 ): ManifestProject | undefined {
   if (!config) return undefined;
-  if (!projectSlug) return config.projects?.[0];
-  const project = config.projects?.find((p) => p.slug === projectSlug);
+  if (!config.projects || config.projects.length === 0) return undefined;
+  if (!projectSlug) return config.projects[0];
+  const project = config.projects?.find((p) => p.slug === projectSlug) ?? config.projects[0];
   return project;
 }
 
--- a/.changeset/modern-weeks-repeat.md
+++ b/.changeset/modern-weeks-repeat.md
@@ -1,0 +1,5 @@
+---
+'@myst-theme/book': patch
+---
+
+Migrated to V2 Remix meta function pattern
--- a/.changeset/ninety-planes-pump.md
+++ b/.changeset/ninety-planes-pump.md
@@ -1,0 +1,6 @@
+---
+'@myst-theme/common': minor
+'@myst-theme/site': minor
+---
+
+Migrated to meta helper functions to the Remix V2 pattern, old meta helpers functions are still available with a `_v1` suffix.
--- a/packages/site/src/seo/meta.ts
+++ b/packages/site/src/seo/meta.ts
@@ -1,7 +1,8 @@
-import type { HtmlMetaDescriptor } from '@remix-run/react';
+import type { HtmlMetaDescriptor, V2_MetaDescriptor } from '@remix-run/react';
 
 type SocialSite = {
   title: string;
+  description?: string;
   twitter?: string;
 };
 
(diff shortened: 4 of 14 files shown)


stevejpurves:
I don't think this is quite right yet and doesn't address the `_index` route either.

I think the documentation is a bit unclear here: https://mystmd.org/guide/seo-and-social as in, where are those fields defined? in project, page or site sections or all?

as @agoose77 points out there are `page` and `project` frontmatter that are not directly merged, i.e. are treated independently by `mystmd`. 

However there is also [SiteFrontmatter](https://github.com/executablebooks/mystmd/blob/main/packages/myst-frontmatter/src/site/types.ts) that also has title, description and thumbnail fields, (this is returned by the meta function in root.tsx).

I think we should agree on the behaviour we want for the meta responses for the `_index.tsx` and `$.tsx` routes, which may be slightly different(?). 

Here is a proposal for fixing `book-theme` (typescripty psuedocode):

### _index.tsx

Motivation here is that we are on the main page of the "book" and so `site` and `project` frontmatterare used.

| field | proposed value |
|-|-|
| title | myst.yml#site.title ? "myst.yml#site.title - myst.yml#project.title" : "myst.yml#project.title" |
| description | myst.yml#site.description ?? myst.yml#project.title |
| thumbnail | myst.yml#project.thumbnail ?? myst.yml#site.thumbnail |

(@fwkoch **note that the SiteFrontmatter type does have a .thumbnail field in the type definition, however when setting a thumbnail in myst.yml#site.thumbnail, it seems like that's not picked up, optimised or added to the site config - is that a bug?)

### $.tsx

Here `page` frontmatter takes precedence

| field | proposed value |
|-|-|
| title | "page.title - {myst.yml#project.title ?? myst.yml#site.title}" |
| description | "page.description ?? myst.yml#project.description ?? myst.yml#site.description |
| thumbnail | page.thumbnail ?? myst.yml#project.thumbnail ?? myst.yml#site.thumbnail |


** NOTE: it **is** proposed that `thumbnailOptimized` ahead of thumbnail but that is omitted from the above for brevity

agoose77:
Ah yes, I meant to add a note about the suspicion that inheritance wasn't applied here. Thanks for catching that!

stevejpurves:
What I am not clear on is the role of the `myst.yml#site.*` fields, are they intended as overrides or fallbacks? the current proposal above treats them as fallbacks (cc @rowanc1)

agoose77:
My understanding is that the `site:` key in `myst.yml` is `SiteConfig`, which is an intersection type of `SiteFrontMatter` and other config fields. So, I would expect to be able to add site-specific frontmatter there.

The documentation [somewhat disagrees with this](https://mystmd.org/guide/frontmatter#in-a-myst-yml-file), suggesting that if you only have `site:` and want to add frontmatter, you should add `project:` to `myst.yml`. 

It's not clear to me what the inheritance order should be either -- in my view there's no point in having `site:` frontmatter unless you want to change something w.r.t `project:` for a site deployment. Or, it has a purpose, which is to separate web config from non-web config. I don't think we want to go down that path; better to define all the config in one place and pick-and-choose what makes sense for different exports.

With MyST, I think we're at a stage where things exist for a reason, so I welcome Rowan's thoughts on why we have site frontmatter in addition to project. With my level of knowledge, I'd be tempted to remove frontmatter from `site:` altogether, though I am no doubt missing something.

agoose77:
Pinging @rowanc1 as this probably needs a wider discussion.

stevejpurves:
yes, that is why I was thinking that site is perhaps useful as an override -- it's worth just saying that although myst sites are single project by default, and certainly the documentation is all pointed there, they do support multiple projects -- this is case where the site frontmatter starts to make more sense maybe.

👍🏼 on wider dicussion

choldgraf:
Just quickly noting that, from my perspective (a user with no knowledge of how the implementation is done), my expectations were:

1. Anything defined in `myst.yml` applies to every page in my documentation.
2. Anything defined on a page overrides anything in `myst.yml`.
3. The `project` and `site` keys are disjoint (AKA they share none of the same keys)
4. If (3) is an incorrect assumption, my assumption would be that `site` keys will over-ride `project` keys, and that page metadata will override both.

stevejpurves:
The `_index` route serves back meta based on the `project` but overridden by the `site`
image


Other routes serve back `page` meta falling back to `project` meta, in the case of title they are merged.
image

packages/common/src/utils.ts
```diff
@@ -17,8 +17,9 @@ export function getProject(
   projectSlug?: string,
 ): ManifestProject | undefined {
   if (!config) return undefined;
-  if (!projectSlug) return config.projects?.[0];
-  const project = config.projects?.find((p) => p.slug === projectSlug);
+  if (!config.projects || config.projects.length === 0) return undefined;
+  if (!projectSlug) return config.projects[0];
+  const project = config.projects?.find((p) => p.slug === projectSlug) ?? config.projects[0];
```
stevejpurves, line 22: always return a project

fwkoch (approved):
I think this is good to get in - big move to `v2` metadata, it will be easy to make tweaks to each page's meta on top of this.

fwkoch:
Pulled some of the site frontmatter discussion over to this issue: https://github.com/executablebooks/mystmd/issues/1103

agoose77:
@stevejpurves can we merge this now? :)

PR #334, opened 2024-03-15
⌨️ Add `keyboard` renderer

agoose77:
This PR is the sibling to executablebooks/mystmd#989, which adds a new `kbd`/`keyboard` role to MyST.

### DIFF
--- a/packages/myst-to-react/src/basic.tsx
+++ b/packages/myst-to-react/src/basic.tsx
@@ -46,6 +46,10 @@ type Aside = {
   type: 'aside';
 };
 
+type Keystrokes = {
+  type: 'keystrokes';
+};
+
 type BasicNodeRenderers = {
   text: NodeRenderer<spec.Text>;
   span: NodeRenderer<GenericNode>;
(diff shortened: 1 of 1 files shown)


agoose77:
Is it reasonable to implicitly assume that the `keyboard` node has a single, `Text`, child like this?

packages/myst-to-react/src/basic.tsx
```diff
@@ -363,6 +368,10 @@ const BASIC_RENDERERS: BasicNodeRenderers = {
       </dd>
     );
   },
+  keyboard({ node }) {
+    const valueNode = (node.children as GenericNode[])[0];
+    return <kbd>{valueNode.value}</kbd>;
```
rowanc1, line 373: I think we should use `<Myst children>` explicitly, which does that, but won't break if it isn't the case.

rowanc1, line 373:
```suggestion
    return <kbd><MyST ast={node.children} /></kbd>;
```

PR #335, opened 2024-03-15
🔧 Update myst dependencies

stevejpurves:
update myst dependencies to the latest.


packages/frontmatter/package.json
```diff
@@ -25,8 +25,8 @@
     "@radix-ui/react-popover": "^1.0.6",
     "@scienceicons/react": "^0.0.6",
     "classnames": "^2.3.2",
-    "myst-common": "*",
```
fwkoch, line 28: Is there a reason these were `"*"`?

packages/myst-demo/src/index.tsx
```diff
@@ -110,17 +110,14 @@ async function parse(
     cite: { order: [], data: {} },
     footnotes: {},
   };
-  const { frontmatter: frontmatterRaw } = getFrontmatter(file, mdast, {
-    removeYaml: true,
```
fwkoch, line 114: This is the default behaviour - removing `removeYaml` makes no difference.

packages/myst-demo/src/index.tsx
```diff
@@ -110,17 +110,14 @@ async function parse(
...
-    removeHeading: options?.removeHeading ?? false,
```
fwkoch, line 115: This is no longer an option in MyST, the heading is always removed if it matches `frontmatter.title`

For myst demo, this will cause funny behaviour in the specific case of matching `frontmatter.title` and first H1...

I think if we want to fix this, we should add the node back explicitly here in `myst-demo`.

themes/article/package.json
```diff
@@ -25,8 +25,8 @@
     "@remix-run/node": "~1.17.0",
     "@remix-run/react": "~1.17.0",
     "@remix-run/vercel": "~1.17.0",
-    "myst-common": "^1.1.18",
```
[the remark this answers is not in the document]
rowanc1: Yes, these should be bumped to the latest versions when we do this!

rowanc1:
When testing this, can you ensure that the epigraph work that @agoose77 completed shows up in the mystmd docs?

packages/myst-demo/src/index.tsx
rowanc1: I think just testing it on any `{myst}` directives in the docs, I suppose that anything starting with an H1 is going to be a problem, if so, we will need to put it back in with any labels as well.

There are a few places we use frontmatter:
- https://github.com/executablebooks/mystjs/blob/main/docs/glossaries-and-terms.md#L53
- https://github.com/executablebooks/mystjs/blob/main/docs/math.md#L212
- https://github.com/executablebooks/mystjs/blob/main/docs/cross-references.md#L135

It might also be easier to just duplicate the logic of pulling the first yaml code block off?

packages/frontmatter/package.json
fwkoch: Ok - left this as permissive

packages/myst-demo/src/index.tsx
fwkoch: Ugh - ok, took your advice and duplicated the `getFrontmatter` logic directly in `myst-demo`. Might be worth adding more flexibility back to the other `getFrontmatter` version at some point...

rowanc1:
Updated to the latest, quote and kbd roles are working. Tested the new frontmatter fixes, they also work.

image

PR #338, opened 2024-03-19
FIX: bump myst-spec

agoose77:
I'm not sure if this is the right solution: my intention is to avoid installing two copies of myst-spec when solving in a busy environment. We could also use
```json
"myst-spec": "^0.0.4 || ^0.0.5"
```

My assertion is that we're compatible with both myst-spec `0.0.4` and `0.0.5`.

This is part of work to unblock `jupyterlab-myst`


rowanc1:
This is taken over by #335.

PR #340, opened 2024-03-19
📖  Add links to the MyST Theme documentation

choldgraf:
Adds a few links to the theme documentation in the README, in the hopes that this makes them easier to discover.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,9 @@
 [![MIT License](https://img.shields.io/badge/license-MIT-blue.svg)](https://github.com/executablebooks/myst-theme/blob/main/LICENSE)
 [![CI](https://github.com/executablebooks/myst-theme/workflows/CI/badge.svg)](https://github.com/executablebooks/myst-theme/actions)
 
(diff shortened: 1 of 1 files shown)


rowanc1:
🚀

choldgraf:
@rowanc1 I tagged @agoose77 just because he was the one affected by this initially, but I think it's fine if you just want to merge!

Probably best to have a future team conversation about this, but personally I'd prefer that we adopt a liberal merge policy like:

> anybody should merge a PR if they think it's ready to go unless explicitly stated otherwise by the author

In my opinion it's more important that we can fluidly iterate than if we get the changes perfect the first time. I know others might want more mechanisms to ensure quality control, but just stating where I'm at there :-)

agoose77:
@choldgraf I'd be tempted to scope that to docs, rather than all PRs given the amount of domain-specific knowledge that I know I need to continue to develop.

rowanc1:
Yeah - I held off merging because you tagged @agoose77, I think we can just keep channels open, use our best discretion, and aim to get things merged as fast as we can.

Issue #344, opened 2024-03-27
Author information needs higher z-index


rowanc1:
When clicking on the author info (possible other hovers as well?) the interaction with the table of contents is not the right ordering.

![image](https://github.com/executablebooks/myst-theme/assets/913249/4d088f4e-48f2-4dfa-8348-aebf4bb2b38c)

stevejpurves:
The Table of contents has a z-index of 10 (https://github.com/executablebooks/myst-theme/blob/e549832055fa51faacc1f3966201e659a1365560/packages/site/src/components/Navigation/TableOfContents.tsx#L70) when it's in non-mobile mode, A search for `z-10` reveals a number of containers being set to z-10, including all content blocks and the document outline -- so a z-0 on the table of content means it's unclickable as soon as you scroll past the header.

if `z-10` for content, doc outline is correct/intended, then we need to apply the same to the author popover. If that `z-10` is not desirable then we can try moving things back to `z-0`, but I'm not clear on why we have `z-10` in place -- it could well be to do with a jupyter rendered output?

rowanc1:
Fixed in #345

PR #345, opened 2024-04-01
🛟 lift author popover to z-10

stevejpurves:
possible fix for #344


rowanc1 (approved):
Changed to `hover-card-content` which also brings the animations with it!

Issue #342, opened 2024-03-23
HTML served with `<img src=http://localhost:3100> even with `HOST=10.0.0.1 myst start --keep-host`


drewlio:
### Describe the bug

When I have a fresh install with the mystmd-quickstart, and do a `HOST=10.0.0.1 myst start --keep-host`,

I expect to remotely access `http://10.0.01:3000` and have it serve the entire web page,

but instead, while most assets are loaded from relative links and work fine, `<img src>` tags point to an absolute URL of `http://localhost:3100/...` (see picture below of incorrect links in the client browser)

![image](https://github.com/executablebooks/myst-theme/assets/13006720/b65ff0b4-8f6e-4bf4-91af-3db109035e5f)

However, the server is binding to 10.0.0.1 and will serve the asset file. (see picture below of directly accessing the asset URL with the correct domain instead of `localhost`)

![image](https://github.com/executablebooks/myst-theme/assets/13006720/9b4a4cfc-dbbb-4628-afbd-935d7d6de7cb)

This is a problem for people trying to use mystmd through a SSH connection to the big server where they do their computational work and where their Jupyter notebooks live and run.

To try to help, could it be related to the hardcoded `localhost` below? It seems like this would be fixed if this was a relative path like the other assets.
https://github.com/executablebooks/myst-theme/blob/e549832055fa51faacc1f3966201e659a1365560/themes/book/app/utils/loaders.server.ts#L14

### Reproduce the bug

1. install mystmd
2. download the quickstart
3. set your HOST env variable
4. `myst start --keep-host`

### List your environment

$ myst --version
v1.1.48

$ myst templates list --site book-theme
ID: site/myst/book-theme
Version: 1.0.0

drewlio:
Looks like article-theme does the same thing
https://github.com/executablebooks/myst-theme/blob/e549832055fa51faacc1f3966201e659a1365560/themes/article/app/utils/loaders.server.ts#L14

drewlio:
On the idea of making it a relative path, [RFC3986 Section 5.2.2](https://www.rfc-editor.org/rfc/rfc3986#section-5.2.2) says you can't change the port in a relative path. If you need to change the port, you have to specify the entire "authority" (host:port).

LecrisUT:
[...]

PR #346, opened 2024-04-01
🚥 allow CONTENT_CDN host to be spec'ed

stevejpurves:
fixing #342 

This allows the hostname used in the `CONTENT_CDN` to be specified independent of `PORT` or `HOST`. 

Note trying to use `HOST` for this purpose is problematic as the express development server attempts to bind to this and fails, we really just want to influence rewriting on urls to static assets, so we are adding `CONTENT_CDN_HOST` in a similar vein to how `CONTENT_CDN_PORT` is used.

A parallel change in `mystmd` is needed in order to add `CONTENT_CDN_HOST` based on `HOST` when starting the dev server (potentially only when `--keep-host` is used?)

### DIFF
--- a/themes/article/app/utils/loaders.server.ts
+++ b/themes/article/app/utils/loaders.server.ts
@@ -38,7 +38,6 @@ function updateLink(url: string) {
   if (process.env.MODE === 'static') {
     return `/myst_assets_folder${url}`;
   }
-  console.log('rewriting:', url, `${CONTENT_CDN}${url}`);
   return `${CONTENT_CDN}${url}`;
 }
 
--- a/themes/article/app/utils/loaders.server.ts
+++ b/themes/article/app/utils/loaders.server.ts
@@ -10,8 +10,10 @@ import {
 } from '@myst-theme/common';
 import { responseNoArticle, responseNoSite, getDomainFromRequest } from '@myst-theme/site';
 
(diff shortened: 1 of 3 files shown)


stevejpurves:
tested this by setting a name on `/etc/hosts` and then `HOST=http://noplacelikehome myst start --keep-host` with the output from `make build-book` copied into `_build/templates` folder

image

themes/article/app/utils/loaders.server.ts
```diff
@@ -34,6 +38,7 @@ function updateLink(url: string) {
   if (process.env.MODE === 'static') {
     return `/myst_assets_folder${url}`;
   }
+  console.log('rewriting:', url, `${CONTENT_CDN}${url}`);
```
rowanc1, line 41:
```suggestion
```

themes/book/app/utils/loaders.server.ts
```diff
@@ -34,6 +38,7 @@ function updateLink(url: string) {
   if (process.env.MODE === 'static') {
     return `/myst_assets_folder${url}`;
   }
+  console.log('rewriting:', url, `${CONTENT_CDN}${url}`);
```
rowanc1, line 41:
```suggestion
```

rowanc1:
I think the approach in #468 is a bit cleaner. Allows to be set without a port. I think this also is a solution for #342.

PR #351, opened 2024-04-03
📥 Prefer frontmatter downloads over exports

fwkoch:
Here's a start to consume the new `downloads` available in project/page frontmatter (https://github.com/executablebooks/mystmd/pull/1049)

Where we previously built a download list from `exports` we now look to see if `downloads` is defined. If so, the theme uses that instead. If `downloads` is not defined, it still falls back to `exports` for backwards compatibility.

This also makes a couple tweaks so download `format` may be undefined - since we allow downloads to be of arbitrary format, we need to handle unknown formats.


themes/article/app/components/Downloads.tsx
```diff
@@ -28,9 +32,10 @@ export function Downloads() {
             key={action.url}
             href={action.url}
             className="inline-block mr-2 no-underline lg:mr-0 lg:block"
+            target={action.internal ? '_self' : '_blank'}
```
fwkoch, line 35: There's probably a better way to do this...?

fwkoch: This also isn't actually respected for relative urls.

stevejpurves:
This PR so far addresses the `article-theme`, how should the `book-theme` be handling these given they are defined on the project?

themes/article/app/components/Downloads.tsx
stevejpurves: changed this up significantly

themes/article/app/components/Downloads.tsx
```diff
@@ -15,24 +23,83 @@ function formatToTitle(format: string) {
...
-          </a>
-        ))}
+          );
+
+          return (
+            <a
+              key={item.url}
+              href={item.url}
+              className="inline-block mr-2 no-underline lg:mr-0 lg:block"
+              download={item.static ? item.filename : undefined}
```
fwkoch, line 91: @fwkoch  Change these to be based on `filename`, not `static`

fwkoch:
I've aligned `DownloadsDropdown` in `frontmatter` with your changes to the article theme - this addresses book theme since it uses that component.

The only difference is `frontmatter` does not depend on `@remix-run/react` so it does not use `Link` for the internal links. Not sure if there's any implications to adding that dep to `frontmatter`?

stevejpurves:
@fwkoch yeah let's keep `Link` out of frontmatter and just use `a` there

stevejpurves (approved):
tested both themes out locally, looks great @fwkoch!

PR #369, opened 2024-04-23
:bookmark_tabs: Add document outline mobile view

### LINKED ISSUE #363: Show "on this page" navigation menu for in-page navigation on mobile
We have an "In this article" menu for in-page navigation to the right of each page. However, on mobile it disappears, and there is no user interface element that lets you discover it.

Being on mobile makes it particularly hard to navigate longer pages, so it would benefit from the ability to quickly see the page's main structure and jump around within it.

Example of the in-page navigation disappearing:

CleanShot 2024-04-16 at 12 18 22

### Example of how this is handled in Jupyter Book

In the pydata and book themes, the right sidebar pops into a drawer on the right. You can open it with a button, similarly to the left drawer:

CleanShot 2024-04-16 at 12 21 26

### Example of how this is handled in Docusaurus

In this case they add a dropdown to the top of each page with in-page links and content.

CleanShot 2024-04-17 at 09 35 55@2x

agoose77:
Fixes #363 by moving the document outline to the main body of the page.

image

- I've added a new `types.ts` file that declares the template option interface.
- I've moved `ArticlePage` from `@myst-theme/site` to `themes/book` (to mirror the one already copied to `themes/article`.
- I've added `hide_outline` and `outline_maxdepth` to the `article` theme options

I need to understand why the `article` theme has its own implementation of `ArticlePage`, instead of re-using that from `@myst-theme/site`. Perhaps @stevejpurves could shed some light? :)

### DIFF
--- a/themes/book/app/types.ts
+++ b/themes/book/app/types.ts
@@ -1,0 +1,6 @@
+export interface TemplateOptions {
+  hide_toc?: boolean;
+  hide_outline?: boolean;
+  hide_footer_links?: boolean;
+  outline_maxdepth?: number;
+}
--- a/themes/book/app/types.ts
+++ b/themes/book/app/types.ts
@@ -3,4 +3,5 @@ export interface TemplateOptions {
   hide_outline?: boolean;
   hide_footer_links?: boolean;
   outline_maxdepth?: number;
+  hide_title_block?: boolean;
 }
--- a/.changeset/great-weeks-sit.md
+++ b/.changeset/great-weeks-sit.md
@@ -1,0 +1,5 @@
+---
+'@myst-theme/site': patch
+---
+
+Mark ArticlePage as deprecated
--- a/.changeset/healthy-waves-sell.md
+++ b/.changeset/healthy-waves-sell.md
@@ -1,0 +1,6 @@
+---
+'@myst-theme/article': patch
+'@myst-theme/book': patch
+---
+
+Support inline navigation
--- a/packages/site/src/pages/Article.tsx
+++ b/packages/site/src/pages/Article.tsx
@@ -6,6 +6,7 @@ import {
   FooterLinksBlock,
   FrontmatterParts,
   BackmatterParts,
+  DocumentOutline,
 } from '../components/index.js';
 import { ErrorDocumentNotFound } from './ErrorDocumentNotFound.js';
 import { ErrorProjectNotFound } from './ErrorProjectNotFound.js';
@@ -53,7 +54,8 @@ export const ArticlePage = React.memo(function ({
   const manifest = useProjectManifest();
   const compute = useComputeOptions();
 
-  const { hide_title_block, hide_footer_links } = (article.frontmatter as any)?.options ?? {};
+  const { hide_title_block, hide_footer_links, hide_outline, outline_maxdepth } =
+    (article.frontmatter as any)?.options ?? {};
   const downloads = combineDownloads(manifest?.downloads, article.frontmatter);
   const tree = copyNode(article.mdast);
   const keywords = article.frontmatter?.keywords ?? [];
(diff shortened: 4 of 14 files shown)


agoose77:
@stevejpurves I've replicated the `DocumentOutline` logic for both `article` and `book` themes; now, only the article differs in that it has supporting downloads too.

It feels like `Article` is a fairly high-level component that is theme specific. I wonder if we should just move `Article` from `@myst-theme/site` into the book theme (such that the `article` and `book` themes have their own independent versions). What do you think about that suggestion? Is there something I'm missing?

agoose77:
OK, I decided to make that change; I assume that as we built out the themes, it became clear that the article theme needed its own `ArticlePage`, and from there we have been left with the type in `@myst-theme`.

agoose77:
I've seen that this is a place where we have project, site, and page frontmatter. I haven't gone through these changes super carefully, so any reviewer would be welcome to cast an eye.

rowanc1:
We have 3-4 themes that are dependent on that export from site, anything in book/article cannot be exported.

agoose77:
Hmm, OK. @rowanc1 can you nudge on whether the core site export should include this change, or whether I should fork it for the book export?

agoose77:
Sure, that's an idea! It might just be that the difference between the two is the addition of downloads.

agoose77:
@rowanc1 I'm actually thinking that we shouldn't try and have an `ArticlePage` in the site package. These article pages are built from published components, so users can already re-implement them fairly easily. I think the differences between articles and book pages should better be described by distinct components instead of a "meta" component that encodes the logic for both variants.

Are you OK with that change; dropping the `Article` from `@myst-theme/site`?

stevejpurves:
@agoose77 I think that suggestion was not to have a single ArticlePage meta component that encoded both sets of changes but indeed to have two distinct page components available in `site` by moving the `article-theme` version up to the package too?

That's leaves us with some naming to sort out but then provides two jumping off points for existing page patterns to be used in downstream themes...?

themes/book/app/types.ts
```diff
@@ -0,0 +1,7 @@
+export interface TemplateOptions {
+  hide_toc?: boolean;
+  hide_outline?: boolean;
+  hide_footer_links?: boolean;
+  outline_maxdepth?: number;
+  hide_title_block?: boolean;
```
fwkoch, line 6: Now that these are a nice clear type, it's easy to compare to the `template.yml` - and in doing so, I noticed `hide_title_block` is not declared there as an option. You mind adding that? (I see you already added a couple options to the article `template.yml`, so I assume it's easy.)

fwkoch (changes requested):
Tried this out locally and the mobile view works great, outline appears at the top as expected and the various options seem to be picked up correctly.

Just summarizing from the discussion here, though, I think we need to keep the existing `ArticlePage`, etc, exports from `@myst-theme/site`, so we don't break dependencies from other external themes. @agoose77 - if you think the correct direction is for each theme to define it's own `ArticlePage`, I think it's fine to leave your new `book/.../ArticlePage.tsx` so book theme no longer needs to import it from `@myst-theme/site`. But there isn't any reason to delete the old version from there (can even mark it as `deprecated`?).

This means we won't be breaking dependencies for external themes; we will just be setting an example of how to do things better, and they can upgrade as they see fit.

Other than that and the template option I mentioned in my other comment, all this needs is a changeset, then should be good to go!

agoose77:
@fwkoch this is a really helpful review, thank you (and also to @rowanc1).

Could you perhaps help nudge this over the line with the deprecation logic?

fwkoch (approved):
@agoose77 - thanks for the updates! This is all I was thinking regarding deprecation: https://github.com/executablebooks/myst-theme/blob/agoose77/wip-outline-inline/packages/site/src/pages/Article.tsx#L45

Should all be good to go now, as far as I'm concerned.

agoose77:
Are you comfortable merging this? Or would you like a separate review e.g. from @stevejpurves?

PR #372, opened 2024-05-04
🔗 Add support for myst.xref.json

fwkoch:
Cross references for external MyST projects is added here: https://github.com/executablebooks/mystmd/pull/1180 - this PR finesses the themes so those changes work well:
- `myst.xref.json` is served by the theme, similar to `objects.inv`
- page json responses have permissive cors response to allow external sites to load the data
- non-existent `/api/lookup` is no longer used for external links


themes/article/app/routes/$slug[.json].tsx
```diff
@@ -14,7 +14,17 @@ function api404(message = 'No API route found at this URL') {
...
+    const xref = await getMystXrefJson();
+    if (!xref) return new Response('MyST Xref not found', { status: 404 });
+    return json(xref);
+  }
   const data = await getPage(request, { slug }).catch(() => null);
   if (!data) return api404('No page found at this URL.');
-  return json(data);
+  return json(data, {
+    headers: {
+      'Access-Control-Allow-Origin': '*',
```
Comment (fwkoch) [inline themes/article/app/routes/$slug[.json].tsx:27]: This covers the basic case of loading cross-reference tool-tips from external sites. It won't work for github pages...

PR #373, opened 2024-05-06
🔍 Make document outline collapsible

### LINKED ISSUE #170: Hide TOC when it intersects with margin content
image

There is some code @choldgraf has pointed us to here:

https://github.com/executablebooks/sphinx-book-theme/blob/a9794b73cec766e01bee35a96f7d44e22851f278/src/sphinx_book_theme/assets/scripts/index.js#L66

rowanc1:
> Adds new collapsible control to hide document outline when elements intersect with the margin and are on screen.

-- @agoose77

https://github.com/user-attachments/assets/6056b370-06cb-445e-8647-39d980d03ce0

Fixes #170


stevejpurves:
👍🏼 cool - I like the use of radix-ui in here! I guess the next step for this is to hook it up to an intersection observer or something to auto-collapse on margin content?

agoose77:
I've done some general refactoring work here too, which touches a reasonable amount of this code. I'm growing in my comfort with React perf specifics, so I might have made a regression here. Would reviewers be able to try this out with that in mind?

packages/site/src/components/DocumentOutline.tsx
```diff
@@ -62,11 +63,7 @@ const Headings = ({ headings, activeId, highlight, selector }: Props) => (
             e.preventDefault();
             const el = document.querySelector(`#${heading.id}`);
             if (!el) return;
-            getHeaders(selector).forEach((h) => {
-              h.classList.remove(HIGHLIGHT_CLASS);
-            });
-            el.classList.add(HIGHLIGHT_CLASS);
-            highlight?.();
```
agoose77, line 69: I have removed this in favour of letting the intersection observer handle it. Wouldn't be surprised if this was a specific bugfix -- I didn't check the blame.

themes/book/app/components/ArticlePage.tsx
```diff
@@ -75,7 +76,7 @@ export const ArticlePage = React.memo(function ({
   const tree = copyNode(article.mdast);
   const keywords = article.frontmatter?.keywords ?? [];
   const parts = extractKnownParts(tree);
-
+  const isOutlineMargin = useMediaQuery('(min-width: 1024px)');
```
agoose77, line 79: It would be nice not to need this, but I think it's the cleanest approach.

choldgraf:
Could you add a GIF of what this looks like to the top comment? I can't figure out how to build previews locally and we don't have a PR-based preview available, so it's hard to know what the UX is like.

agoose77:
@choldgraf done!

choldgraf (approved):
This looks great to me from an end functionality perspective! As long as we aren't introducing any bugs or weird edge case regressions I'd be +1 on merging if somebody gives a +1 on the tech implementation

agoose77:
@stevejpurves I think the total set of class names that intersect with the margin is:
```
col-margin-right-inset
col-gutter-outset-right
col-screen-right
col-screen-inset-right
col-page-right
col-page-inset-right
col-body-outset-right
col-gutter-page-right
```

Based upon https://jupyter-book.github.io/myst-theme/?path=/docs/components-grid-system--docs. Do you agree?

stevejpurves:
@agoose77 if you don't include col-screen, col-page, col-page-inset and col-body-outset could, for example, a full width figure spanning out from the main body content use one of those or is it expected that col-page-inset-right, col-body-outset-right would cover those cases?

PR #390, opened 2024-05-16
🎨 Add H5 and H6 styles

### LINKED ISSUE #389: Headers five and six are not styled
Headers 5 and 6 don't have any kind of styling, we should at least **bold** them so that it's clear they're headers.

CleanShot 2024-05-15 at 11 49 11@2x

choldgraf:
This adds CSS styles for h4/h5/h6 so that they look like headers in the text.

It also consolidates one CSS rule, mostly so that I can validate with @stevejpurves that this JS-based configuration lets us write the same kind of structures that work in SCSS/SASS. Is that right Steve?

- closes #389

### DIFF
--- a/styles/index.js
+++ b/styles/index.js
@@ -84,9 +84,13 @@ const themeExtensions = {
       css: {
         code: {
           fontWeight: '400',
-          '&::before, &::after': {
-            content: '',
-          }
+        },
+        // These code before/after elements are hard coded to remove the backticks, "`", that are in by default.
+        'code::before': {
+          content: '',
+        },
+        'code::after': {
+          content: '',
         },
         'blockquote p:first-of-type::before': { content: 'none' },
         'blockquote p:first-of-type::after': { content: 'none' },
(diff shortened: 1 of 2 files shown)


styles/index.js
```diff
@@ -84,12 +84,9 @@ const themeExtensions = {
       css: {
         code: {
           fontWeight: '400',
-        },
-        'code::before': {
-          content: '',
-        },
-        'code::after': {
-          content: '',
+          '&::before, &::after': {
```
rowanc1, line 87: Note that this might be written like this previously to override those parts of the tailwind typography, not 100% sure though. Please test before merging this code style before/after change!

rowanc1: Changing this does introduce changes to the code styling:

image

H4 also already has styles! Updating.

rowanc1:
Merging as at least an iterative improvement. Happy to take more style directions.

I was trying to find my typography reference guide that had essentially a take-down piece on h5 and h6. But the web is awash with SEO and I can't find anything very useful at the moment... :(

PR #397, opened 2024-06-05
🦶🎵 Add footnotes to bottom of page

rowanc1:
See: executablebooks/mystmd#926

image

Can click on the footnotes back and forward to go from the bottom of the page to the reference. Hover still works. I have refactored some of the scroll logic to enable us to do push state rather than replace in the future, however that didn't work in my first pass, leaving that for another iteration!


rowanc1:
There is going to be additional logic to find a footnote that is hidden (e.g. in a different footnote) currently not tackling that!

PR #408, opened 2024-06-20
:gear: More preparation for Remix v2 (1)

agoose77:
More work on preparing for v2 in Remix, fixing `CatchBoundary` and some other small changes.

PR #409, opened 2024-06-20
:arrow_up: Use Remix V2

agoose77:
This PR builds upon #408 and actually switches to Remix V2, using the classic remix compiler. A future PR would need to be used to pull in the Vite compiler.


rowanc1:
If you npx serve the built html, landing on any page would go into an infinite reload loop. It is very obvious.

It may only be the case if you change the base path?

E.g. BASE_URL=/_build/html

And then serve from the project directory.

The json that remix sends us seen to be invalid, because we changed it, and then it goes into a loop. Possible that that has been fixed as we are skipping many versions.

agoose77:
I can confirm that I'm seeing the continuous loop behaviour that you reported for 1.1.18 in version 2.x. It looks like we might need to consider the other routes to static exports (like SSG SPA that I linked previously).

rowanc1:
Very unfortunate. :(

rowanc1:
Maybe? I think that would be just changing the request coming in (i.e. _build/html/slug) and then in the loader ensuring that matches.

stevejpurves:
btw @agoose77 I think the best future for this would be to probably skip `v2` and head straight to `Remix_v3`/`ReactRouter_v7`

agoose77:
@stevejpurves lol are you following me!?

I was getting this particular PR up to speed before the summit today (I'm jetlagged) in order to try and get MyST serving at a different base path. It feels a lot less painful to just serve the site under `/foo/bar` rather than doing URL rewriting, proxying, etc, to get everything in the right place.

I hadn't seen the v3 news -- that's interesting. Definitely going to look at the 2->3 upgrade instead of the other PR now.

agoose77:
I'm closing this, and going to start a new react-router PR.

PR #410, opened 2024-06-21
:fast_forward: Use vite instead of Remix Classic Compiler (3)

agoose77:
This PR works towards using Vite instead of the Remix Classic Compiler. From reading, it seems like only Vite supports changing the base path, which I think is a required solution to the v2 redirect problem (actually serve the theme from the base path it will be served from)

As this is an iterative PR, here are some findings:
- Switch to Vite configuration
- Split server and client loaders
- Drop LiveReload (Vite has HMR)


agoose77:
So it looks like the v1-v2 routing thing was maybe a fluke? I sat back down to the branch today and it broke the navigation.

Without that change, navigation and link-intent from non-index pages works, but the index seems to trigger a reload.

agoose77:
Looks like the redirect phenomenon is actually `@remix-run/react` triggering a reload when it fails to dynamically import the route module.

agoose77:
I'm closing this, and going to start a new react-router PR.

PR #417, opened 2024-07-01
:wrench: Fix positioning of secondary sidebar when scrolling

agoose77:
@rowanc1 brought this to my attention somewhere.

### DIFF
--- a/.changeset/afraid-plants-kiss.md
+++ b/.changeset/afraid-plants-kiss.md
@@ -1,0 +1,6 @@
+---
+'@myst-theme/article': patch
+'@myst-theme/book': patch
+---
+
+Fix position of secondary sidebar during scrolling
--- a/themes/article/app/components/Article.tsx
+++ b/themes/article/app/components/Article.tsx
@@ -17,6 +17,8 @@ import { copyNode } from 'myst-common';
 import { BusyScopeProvider, ConnectionStatusTray, ExecuteScopeProvider } from '@myst-theme/jupyter';
 import { SourceFileKind } from 'myst-spec-ext';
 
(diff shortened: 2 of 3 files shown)


agoose77:
Huh, testing this was difficult. I'm not sure why setting `top` via the JSX `style` attribute behaves as expected, but setting the same value via tailwind classes does not. I assume it's a priority thing. In any case, I noticed that we lost the `useThemeTop` logic. I didn't see why we added that to begin with, but I assume it's all in the effort to be more consistent.

I've just eye-balled the vertical position thereafter; the TOC is using flex, whereas the secondary nav container is block display. So, it's approximate.

stevejpurves:
@agoose77 how were you setting the tailwind class exactly?

agoose77:
> @agoose77 how were you setting the tailwind class exactly?

In the `className` attribute of the JSX.

stevejpurves:
@agoose77 ok I just wondered if you were using an arbitrary value pattern? like:

```{`p-2 some-class top-[${top}px] border ...`}```

As tailwind cannot compile that to a class name at build time, in those circumstance using style is best

stevejpurves:
book-theme looks great 👍🏼 

article-theme looks worse though?

image

It's not aligned with content top or anything else.

while previously it looks good alignment wise

image

agoose77:
Hmm, I see your point. Maybe let's revert that; I'll check the rationale when next at my pc.

stevejpurves:
Latest:

book-theme
image

article-theme
image

stevejpurves (approved):
some tweaks and more eyeballing but looks ok!

PR #421, opened 2024-07-03
🔧 fix lg breakpoint for center-grid

stevejpurves:
fixes a issue with the grid and makes spacing better 🤩 

image

### DIFF
--- a/styles/grid-system.css
+++ b/styles/grid-system.css
@@ -29,7 +29,7 @@
   }
   .col-margin-right,
   .col-margin {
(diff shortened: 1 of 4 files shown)


styles/grid-system.css
```diff
@@ -29,7 +29,7 @@
   }
   .col-margin-right,
   .col-margin {
-    @apply col-body lg:col-margin-right-lg;
+    @apply col-body xl:col-margin-right-lg;
```
stevejpurves, line 32: this is great fo center-grid but breaks everything else!

PR #431, opened 2024-07-20
:mage: Render new nodetypes for Sphinx autodoc

agoose77:
This PR adds some basic styling for mdast nodes emitted by https://github.com/executablebooks/sphinx-ext-mystmd

### DIFF
--- a/packages/myst-to-react/src/autodoc.tsx
+++ b/packages/myst-to-react/src/autodoc.tsx
@@ -85,12 +85,7 @@ const AUTODOC_RENDERERS = {
   },
   descSignature({ node }: { node: HasChildren }) {
     return (
-      <dt
-        style={{
-          fontWeight: 'lighter',
-          fontFamily: "Menlo, Consolas, 'DejaVu Sans Mono', monospace",
-        }}
-      >
+      <dt className="font-mono font-light">
         <MyST ast={node.children} />
       </dt>
     );
--- a/packages/myst-to-react/src/autodoc.tsx
+++ b/packages/myst-to-react/src/autodoc.tsx
@@ -47,12 +47,7 @@ const AUTODOC_RENDERERS = {
   descName({ node }: { node: HasTextChildren }) {
     const [child] = node.children;
     return (
-      <span
-        style={{
-          fontWeight: 'bold',
-          whiteSpace: 'pre',
-        }}
-      >
+      <span className="whitespace-pre font-bold">
         {child.value}
       </span>
     ); //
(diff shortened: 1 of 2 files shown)


packages/myst-to-react/src/autodoc.tsx
```diff
@@ -0,0 +1,134 @@
...
+};
+
+const preParentRenderer = ({ node }: { node: HasTextChildren }) => {
+  const [child] = node.children;
+  return (
+    <span
+      style={{
+        whiteSpace: 'pre',
+      }}
+    >
```
rowanc1, line 22:
```suggestion
    <span className="whitespace-pre">
```

packages/myst-to-react/src/autodoc.tsx
```diff
@@ -0,0 +1,134 @@
...
+  },
+  descName({ node }: { node: HasTextChildren }) {
+    const [child] = node.children;
+    return (
+      <span
+        style={{
+          fontWeight: 'bold',
+          whiteSpace: 'pre',
+        }}
+      >
```
rowanc1, line 55:
```suggestion
      <span className="whitespace-pre font-bold">
```

packages/myst-to-react/src/autodoc.tsx
```diff
@@ -0,0 +1,134 @@
...
+    );
+  },
+  descSignature({ node }: { node: HasChildren }) {
+    return (
+      <dt
+        style={{
+          fontWeight: 'lighter',
+          fontFamily: "Menlo, Consolas, 'DejaVu Sans Mono', monospace",
+        }}
+      >
```
rowanc1, line 102:
```suggestion
      <dt className="font-mono font-light">
```

packages/myst-to-react/src/autodoc.tsx
```diff
@@ -0,0 +1,134 @@
...
+        <MyST ast={node.children} />
+      </dd>
+    );
+  },
+  descAnnotation: renderChildren,
+  descParameter: renderChildren,
+  descReturns({ node }: { node: HasChildren }) {
+    return (
+      <>
+        <span>&#x2192; </span>
```
rowanc1, line 119: As this is decoration, probably better in css? You can do this in a `:before` node and set the content to an arrow.

rowanc1 (changes requested):
As these are specific for sphinx, are not used in scientific , and the AST might change. I think these should be their own package. Ideally we have them as an export out of `myst-ext-autodoc/react`?

Doing that makes it a bit harder to do the auto reload, but I think it is maybe worth it?

If we did that, then I still think these should come in by default in the themes we have, but other themes can make different choices.

rowanc1:
In conversation today: we should see how far we can get on doing a pre-transform on the AST to use `kind` to differentiate, e.g., a field list, on a `definitionList`.

agoose77:
Closed in favour of #443

PR #441, opened 2024-07-31
📏 Allow full-length TOC when sidebar is useable

agoose77:
Presently, the TOC is always trimmed to the page length, in full-width OR narrow displays:

- Wide:   
  image
- Narrow  
  image

This PR tweaks the computation to detect when the "wide" media query is true/false:

- Wide:  
  image
- Narrow:  
  image

I think there might be some overflow problems, but I don't *think* they're caused by this PR.

### DIFF
--- a/packages/providers/src/ui.tsx
+++ b/packages/providers/src/ui.tsx
@@ -30,7 +30,7 @@ export function useNavOpen(): [boolean, (open: boolean) => void] {
   return [state?.isNavOpen ?? false, setOpen];
 }
 
-export function isWide(): boolean {
+export function useIsWide(): boolean {
   const [state, _] = useContext(UiContext) ?? [];
   return state?.isWide ?? false;
 }
--- a/packages/providers/src/ui.tsx
+++ b/packages/providers/src/ui.tsx
@@ -3,11 +3,12 @@ import { useMediaQuery } from './hooks.js';
 
 type UiState = {
   isNavOpen?: boolean;
+  isWide?: boolean;
 };
 
 type UiContextType = [UiState, (state: UiState) => void];
 
-const UiContext = createContext<UiContextType | undefined>(undefined);
+export const UiContext = createContext<UiContextType | undefined>(undefined);
 
 // Create a provider for components to consume and subscribe to changes
 export function UiStateProvider({ children }: { children: React.ReactNode }) {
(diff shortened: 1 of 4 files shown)


agoose77:
This PR bumps each package as a `patch` change. Is that right? Technically we're now exporting two new things from `ui/providers`.

agoose77:
Note: I want to generalise the TOC sidebar into a primary sidebar, but that's more work.

themes/book/app/routes/$.tsx
```diff
@@ -102,10 +102,32 @@ export function ArticlePageAndNavigation({
...
+}) {
+  return (
+    <UiStateProvider>
+      <ArticlePageAndNavigationInternal
+        children={children}
+        hide_toc={hide_toc}
+        projectSlug={projectSlug}
+        inset={inset}
+      />
+    </UiStateProvider>
```
agoose77, line 128: Is there a nicer way to access the result of a `useContext` _inside_ the component tree?

rowanc1:
Where is the footer coming from in the UI?

agoose77:
> Where is the footer coming from in the UI?

Oh, haha, I appended it to the dom during testing. It only matters presently for the mystmd website.

packages/providers/src/ui.tsx
```diff
@@ -28,3 +29,8 @@ export function useNavOpen(): [boolean, (open: boolean) => void] {
   };
   return [state?.isNavOpen ?? false, setOpen];
 }
+
+export function isWide(): boolean {
```
rowanc1, line 33: can we rename this to `useIsWide()`

packages/providers/src/ui.tsx
```diff
@@ -28,3 +29,8 @@ export function useNavOpen(): [boolean, (open: boolean) => void] {
...
+  const [state, _] = useContext(UiContext) ?? [];
```
rowanc1, line 34:
```suggestion
  const [state] = useContext(UiContext) ?? [];
```

agoose77:
This is now done!

themes/book/app/routes/$.tsx
stevejpurves: do you mean that you want to access the UIStateProvider context within this same function, rather than creating a new wrapper component? you can do something like this but it's not nicer...

```
return (
    <UiStateProvider>
        {(() => {
            const uiState = useUiState();
            return (
                <ArticlePageAndNavigationInternal
                    children={children}
                    hide_toc={hide_toc}
                    projectSlug={projectSlug}
                    inset={inset}
                />);
            })
        }
    </UiStateProvider>
```

agoose77: Oh yep, that's one way!

agoose77:
@stevejpurves I rebased this on main, would you be able to confirm it still behaves as expected on mystmd.org?

stevejpurves:
rebased again, review in progress locally... having to update mystmd.org on some orthogonal changes that have happened whilst this PR has been alive...

stevejpurves:
all looks good, sorry for the lengthy review cycle!

PR #443, opened 2024-08-01
:cat2: Add built-in styling for Sphinx-annotated ASTs

agoose77:
In #431 I explored adding pass-through rendering support for Sphinx AST nodes (generated from a separate extension for Sphinx). This PR instead looks for annotations on a canonical MyST AST to make the styling required to improve Sphinx rendering. Currently this is only focused on autodoc nodes.

There are some limitations in our AST right now that make this a bit fiddly, namely that many node types don't support custom classes. Consequently, some of the selectors are more like heuristics, but they are all gated on a parent class starting with `sphinx-`.

1722518444-wayshot
1722518425-wayshot
1722518401-wayshot


agoose77:
This styling isn't perfect, but it's simple enough to start with.

Issue #402, opened 2024-06-13
Top header navigation falls out of top bar in thin views


agoose77:
- Wide: ![image](https://github.com/user-attachments/assets/53cd0917-c482-42a6-a629-e1722e2b0fc1)
- Narrow: ![image](https://github.com/user-attachments/assets/3d900d13-4182-4fcf-8b99-b4b0d14974b6)

It can be seen that the navbar entry "2i2c links" is missing.

rowanc1:
This is something that @stevejpurves is working on currently. :)

parmentelat:
I'd like to outline that the hamburger button in jb1 can be used to slide the toc away
and this does not seem to work under mystmd

not quite sure what the plan originally was, but as far as I am concerned the ability to slide the toc away is super useful
the toc is helpful to navigate the book, but once a page has been located it helps to be able to use the real estate in full !

PR #444, opened 2024-08-05
:compass: Move collapsed top-bar to primary sidebar on small screens

agoose77:
This PR reworks our `TableOfContents` as a `PrimarySidebar`. In doing so, it adds a new navigation section that is visible below the `lg` selector (the breakpoint at which the top-nav disappears).

- `>lg`: image
- `<lg`: image

This uses the same Radix components as the TOC, so it feels on brand.

Fixes https://github.com/jupyter-book/myst-theme/issues/402


choldgraf:
From a UX perspective I think this is an improvement over what we've got right now (which is that they just disappear). It also matches what the pydata theme does:

CleanShot 2024-08-06 at 11 12 28

The PST takes it a step further and puts *all* header items into the top section of the primary sidebar (except for a dedicated search button). But I think that's a decision that could be made in the future (e.g. whether to put them all in the sidebar, or have them split between primary and secondary sidebars, or have some other way to access header items on narrow screens).

agoose77:
@choldgraf Yes, PST is a really good prior-art reference for what we might want to happen here. With all of the open PRs merged we should be in a reasonably good spot for small viewports. Folding in other components like the `logo` and/or `logo_text` might be something to think about too. But, I also feel that the best practices there also mean it's less of an issue; small logos with text, or big logos without text seem like the best approach in the top bar. 

I'm hopeful we'll get all this in, and then do another round of UX evaluations.

stevejpurves:
@agoose77 made some initial changes here with the intent of keeping the components reusable in other themes including composed themes like mystmd.org which doesn't yet work out of the box with these changes. Basically this is lifting props and letting components higher up the render tree (ideally in the themes) deal with getting the right SiteManifest etc...

This PR is looking good but I'd like to get a parallel PR up on mystmd.org consuming these before merging. Doing that next.

stevejpurves:
We've also lost alignment after this PR

BEFORE:
alignment-book-current


AFTER:
alignment-book-new

agoose77:
@stevejpurves I don't *think* it's this PR that broke alignment; I think something else that merged into main already did this.

stevejpurves:
@agoose77 I've fixed alignment here in any case as doing so requires moving between this repo and other dependent repos - so to avoid duplicating work, and as the changes required are in the PrimarySidebar as well as a couple of other places, doing it in this pass.

stevejpurves:
OK this is good to go!!!

Made some changes to scroll behaviour as the 2 independent scrolling units in the sidebar does not work well.


https://github.com/user-attachments/assets/9cb35e49-4a19-4204-9c9a-feb84e7561b6

Intending to bring this in ASAP to unblock PR's elsewhere

stevejpurves (approved):

PR #445, opened 2024-08-05
:cookie: Use `localStorage` instead of cookie for static builds' theme

### LINKED ISSUE #319: Light / dark mode does not persist between pages in static export
## Description

The light and dark mode changes back to the default when you navigate between pages. This is on an HTML build and can be demoed here:

https://2i2c.org/report-czi-2021/

For example:

CleanShot 2024-03-01 at 18 00 46

## Proposed solution

Persist the status of light/dark by using an HTML browser window key variable.

agoose77:
This PR doesn't remove the cookie pathway, but rather augments it for static builds. We shouldn't end up with a cookie in these contexts.

Fixes #319 
Fixes https://github.com/jupyter-book/mystmd/issues/585

Also fixes bug in which system preference is ignored.

### Live MyST server:

https://github.com/user-attachments/assets/6725c283-e640-4199-b298-d056ce18b60a

### Static MyST HTML build:

https://github.com/user-attachments/assets/9d6d9f3e-8d57-4211-b4fe-404f6ad92f77


agoose77:
I took great inspiration from https://www.mattstobbs.com/remix-dark-mode/#2-initial-state-from-user-preferences

agoose77:
Wow, that was a bit of a journey. I think I've learned a lot more about React and SSR.

The challenging part of this puzzle was dealing with client-side state (`matchMedia` and `localStorage`) in a way that avoids a [FOUC](https://en.wikipedia.org/wiki/Flash_of_unstyled_content) due to a client-only re-render. 

The way that [the above guide](https://www.mattstobbs.com/remix-dark-mode/#2-initial-state-from-user-preferences) handles dark-mode is to defer responsibility for choosing the theme to the client in cases where we don't have an SSR_available theme. When the client hydrates, it can read the client-side APIs to determine the proper value of `theme`. However, this means that the SSR result may not correctly hydrate with the intended state on the client. There are two ways to deal with this:
1. Ensure the initial hydration matches SSR and trigger a re-render (e.g. using `useEffect`)
2. Modify the DOM in-place before hydration such that it hydrates properly

(2) is the approach taken in the core theme logic that sets the tailwind class on the `html` node. We only need to do this if we don't know the SSR-determined theme, which occurs whenever a route is loaded without a set cookie, i.e. for _every_ static build page load, and _any_ dynamic SPA load whereby the user has not toggled the theme before.

We also need to contend with the theme selector which has to reconcile the same problem. I opted to remove any theme-dependence of the DOM here, so the tooltips, aria labels, and SVGs are invariant, and tailwind is left in charge of handling appearance.

Another

agoose77:
One point to note; this means that the `useTheme` hook potentially is not very useful in our application; we need to handle the initial state loading on the client in a special manner. I don't know if there is a way to indicate this in our API so that we don't accidentally use `const { theme } = useTheme()` and find ourselves surprised that it doesn't work as expected for static builds.

packages/providers/src/theme.tsx
```diff
@@ -56,48 +56,102 @@ type ThemeContextType = {
...
+const PREFERS_LIGHT_MQ = '(prefers-color-scheme: light)';
+
+const THEME_KEY = 'myst:theme';
+
+/**
+ * A blocking element that runs on the client before hydration to update the <html> preferred class
+ * This ensures that the hydrated state matches the non-hydrated state (by updating the DOM on the
+ * client between SSR on the server and hydration on the client)
+ */
+export function BlockingThemeLoader({ useLocalStorage }: { useLocalStorage: boolean }) {
```
rowanc1, line 68: I think that we should be making the providers code a lot _smaller_ and removing this logic/components, and bringing it into the site instead.

For example, when instantiating the providers, put the loading/setting functions in at that stage (allowing different themes to make different choices). E.g. setting a cookie, modifying the dom, etc. are part of the site implementation, where the providers are just the interface to the state.

packages/site/src/loaders/theme.server.ts
```diff
@@ -1,11 +1,12 @@
...
-import { isTheme, Theme } from '@myst-theme/providers';
+import { isTheme } from '@myst-theme/providers';
+import type { Theme } from '@myst-theme/providers';
 import type { ActionFunction } from '@remix-run/node';
 
 export const themeStorage = createCookieSessionStorage({
   cookie: {
     name: 'theme',
-    secure: true,
+    secure: false,
```
rowanc1, line 9: What is the reason for this change?

rowanc1 (dismissed):
These changes are going to conflict with the choices made in downstream themes. I think there are a few small tweaks that we can make to ensure the logic/components for these choices is made in `site`, not `providers`.

Let me know if that makes sense, or I can expand on what I mean further..!

packages/site/src/loaders/theme.server.ts
agoose77: For me (Firefox), a local MyST http server can't use cookies without disabling this.

agoose77:
@rowanc1 I've reworked the logic such that the providers are very thin, and all of the business logic is really in a new `hooks` subdirectory in `site`. Does this work better for you?

rowanc1:
Thank you yes, that looks solid - I will take a look/test this afternoon and aim to get it in!

packages/site/src/loaders/theme.server.ts
rowanc1: Can we pop this in another PR/issue? I can handle it in my testing pass.

agoose77, line 9:
```suggestion
    secure: true,
```

agoose77 (commented):
Removing change to cookie

rowanc1:
I will fix up the conflicts and merge on Monday. Sorry for my delay.

stevejpurves:
did some final regression checks on non-static sites, looks good, merging!

PR #447, opened 2024-08-07
📅 Fix representation of MyST date strings

agoose77:
This PR is the counterpart to https://github.com/jupyter-book/mystmd/pull/1426 which ensures that the validated frontmatter contains `yyyy-mm-dd` formatted dates as wall-clock dates (i.e. not tied to a timezone).

At the moment, myst-theme incorrectly builds a UTC object, such that the result is TZ sensitive. This PR fixes the date handling such that we're not displaying dates in any timezone; what you put in is what you put out.
- [ ] Fix locale

### DIFF
--- a/packages/frontmatter/src/FrontmatterBlock.tsx
+++ b/packages/frontmatter/src/FrontmatterBlock.tsx
@@ -74,14 +74,16 @@ export function DateString({
   spacer?: boolean;
 }) {
   if (!date) return null;
-  // Parse the date, either using date's intrinsic timezone, or local timezone
-  const d = new Date(date);
-  // Rebuild the timezone aware date in UTC without applying the TZ shift
-  const utcDate = new Date(Date.UTC(d.getFullYear(), d.getMonth(), d.getDate()));
+  // Parse the date
+  // As this is a YYYY-MM-DD form, the parser interprets this as a UTC date
+  // (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#the_epoch_timestamps_and_invalid_date)
+  const utcDate = new Date(date);
 
+  // Now cast our UTC-date into the local timezone
+  const localDate = new Date(utcDate.getUTCFullYear(), utcDate.getUTCMonth(), utcDate.getUTCDate());
 
-  // Now format as human-readable
-  const dateString = utcDate.toLocaleDateString('en-US', format);
+  // Then format as human-readable in the local timezone.
+  const dateString = localDate.toLocaleDateString('en-US', format);
   return (
     <time dateTime={date} className={classNames({ 'text-spacer': spacer })}>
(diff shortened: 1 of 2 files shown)


packages/frontmatter/src/FrontmatterBlock.tsx
```diff
@@ -74,11 +74,20 @@ export function DateString({
   spacer?: boolean;
 }) {
   if (!date) return null;
-  const d = new Date(date); // This is in the users timezone
-  const utcDate = new Date(d.getUTCFullYear(), d.getUTCMonth(), d.getUTCDate());
+  // Parse the date, either using date's intrinsic timezone, or local timezone
+  const d = new Date(date);
+  // Rebuild the timezone aware date in UTC without applying the TZ shift
+  const utcDate = new Date(Date.UTC(d.getFullYear(), d.getMonth(), d.getDate()));
```
agoose77, line 80: We might not even need to do this now; https://github.com/jupyter-book/mystmd/pull/1426 ensures that the validated date has `dddd-mm-yy` format, so we could just parse it as a string and then build `utcDate` from the components.

stevejpurves: we should just pass yyyy-mm-dd into the `<time>` element https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time it supports it.

packages/frontmatter/src/FrontmatterBlock.tsx
```diff
@@ -74,11 +74,20 @@ export function DateString({
...
+
+  // Format the date as a machine-readable date
+  const isoString = utcDate.toISOString();
+  const match = isoString.match(/^\d+-\d+-\d+/);
+  const isoDateString = match ? match[0] : date;
+
+  // Now format as human-readable
   const dateString = utcDate.toLocaleDateString('en-US', format);
```
stevejpurves, line 88: This will still cause a timezone conversion out of UTC to local and could cause the date to shift

stevejpurves (dismissed):
@agoose77 I don't think this is there yet, have suggested some changes - testing timezones is always tricky

packages/frontmatter/src/FrontmatterBlock.tsx
agoose77: Hmm, it looks like you're right - I missed this description in the api spec.

I think we can probably change tack here and circumvent the utc usage altogether.

agoose77:
OK, setting the timezone to PST and ingesting a UTC date (YYYY-MM-DD) indeed formats improperly:
```node
> process.env.TZ = "America/Anchorage"
'PST'
> (new Date("2024-06-06")).toLocaleDateString("en-GB")
'05/06/2024'
```
Fixed in 78061a4

rowanc1:
This is now just docs!

PR #450, opened 2024-08-21
📏 Fix spacing in themes

agoose77:
I noticed that there are a couple of spacing problems right now in our themes:

1. The `abstract` heading is padded:  
  image
2. The article title is too low in book theme, and the document outline too:
  image

I think the `abstract` heading should be `not-prose`, as we don't want to pad it. Do you agree @stevejpurves? 

As for `useThemeTop`, it looks like a mistake that it's used in the `main` body. So, I've removed it.

### DIFF
--- a/packages/site/src/components/Abstract.tsx
+++ b/packages/site/src/components/Abstract.tsx
@@ -16,7 +16,7 @@ export function Abstract({
   if (!content) return null;
   return (
     <div className={className}>
-      <h2 id={id} className="mb-3 text-base font-semibold group">
+      <h2 id={id} className="mb-3 text-base font-semibold group not-prose">
         {title}
         <HashLink id={id} title={`Link to ${title}`} hover className="ml-2" />
       </h2>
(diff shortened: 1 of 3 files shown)


stevejpurves:
some of this may collide with changes i've already made in #444, maybe re-assess this now that is in @agoose77?

stevejpurves:
on the abstract heading... that probably also applies to any FrontmatterPart not just Abstract and I think we might want to cut the padding but keep other styling. The article-theme is where I really saw the mismatch.

packages/site/src/components/Abstract.tsx
```diff
@@ -16,7 +16,7 @@ export function Abstract({
   if (!content) return null;
   return (
     <div className={className}>
-      <h2 id={id} className="mb-3 text-base font-semibold group">
+      <h2 id={id} className="mb-3 text-base font-semibold group not-prose">
```
rowanc1, line 19: Making this not-prose will change the font which is not ideal. Can we just add `mt-0`?

stevejpurves:
I think that we already have good alignment without this PR and could close @agoose77, I took care of alignment in with the toc/nav work

PR #451, opened 2024-08-21
0️⃣ Support `null` dates

agoose77:
Fixes rendering of "now" dates (`null`)

### DIFF
--- a/.changeset/tame-kangaroos-train.md
+++ b/.changeset/tame-kangaroos-train.md
@@ -1,0 +1,5 @@
+---
+'@myst-theme/frontmatter': patch
+---
+
+Handle null dates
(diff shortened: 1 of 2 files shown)


rowanc1:
What is the use case for this? Can we send through "today" instead? `null`-ish to me means _no_ date.

I can (maybe) see that a preprint/blog gets a `today` string, that is recognized by the CLI (not the theme) and then serialized to the frontend as the date it was built on. That is pretty dangerous though - it means all your blog posts are always written "today" whenever you rerender them (which is probably not what people want).

I don't know if I understand the use case where we want the frontend to show today's date whenever you refresh .. ?

agoose77:
We discuss this over at mystmd, but the high level is that implicit 'today' is a bad thing. I'm closing this pr.

PR #460, opened 2024-08-31
⏩ Extend renderers to allow for kind/subtypes

rowanc1:
This allows for multiple renderers to be attached to a single node, based on a match of the kind.

For example you can now add a specific renderer for a `link[kind=github]`. This allows for inheriting these renderers and then extending or overriding them independently.

### DIFF
--- a/packages/myst-to-react/src/MyST.tsx
+++ b/packages/myst-to-react/src/MyST.tsx
@@ -1,3 +1,5 @@
+import { matches } from 'unist-util-select';
+import type { NodeRenderersValidated } from '@myst-theme/providers';
 import { useNodeRenderers } from '@myst-theme/providers';
 import type { GenericNode } from 'myst-common';
 
@@ -10,17 +12,25 @@ function DefaultComponent({ node }: { node: GenericNode }) {
   );
 }
 
+export function selectRenderer(renderers: NodeRenderersValidated, node: GenericNode) {
+  const componentRenderers = renderers[node.type] ?? renderers['DefaultComponent'];
+  const SpecificComponent = Object.entries(componentRenderers).find(
+    ([selector]) => selector !== 'base' && matches(selector, node),
+  )?.[1];
+  return SpecificComponent ?? componentRenderers.base ?? DefaultComponent;
+}
+
 export function MyST({ ast }: { ast?: GenericNode | GenericNode[] }) {
   const renderers = useNodeRenderers();
   if (!ast || ast.length === 0) return null;
   if (!Array.isArray(ast)) {
-    const Component = renderers[ast.type] ?? renderers['DefaultComponent'] ?? DefaultComponent;
+    const Component = selectRenderer(renderers, ast);
     return <Component key={ast.key} node={ast} />;
   }
   return (
     <>
       {ast?.map((node) => {
-        const Component = renderers[node.type] ?? DefaultComponent;
+        const Component = selectRenderer(renderers, node);
         return <Component key={node.key} node={node} />;
       })}
     </>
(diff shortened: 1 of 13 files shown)


packages/myst-to-react/src/links/index.tsx
```diff
@@ -134,8 +137,20 @@ export const linkBlock: NodeRenderer<TransformedLink> = ({ node }) => {
...
+const LINK_RENDERERS: NodeRenderers = {
+  link: {
+    base: SimpleLink,
+    // Then duplicate the renderers for protocols
+    'link[protocol=github]': GithubLinkRenderer,
+    'link[protocol=wiki]': WikiLinkRenderer,
+    'link[protocol=rrid]': RRIDLinkRenderer,
+    'link[protocol=ror]': RORLinkRenderer,
+    // Put the kinds last as they will match first in the future
+    'link[kind=github]': GithubLinkRenderer,
```
rowanc1, line 149: Prepares for:

https://github.com/jupyter-book/mystmd/issues/1417

packages/myst-to-react/src/MyST.tsx
```diff
@@ -10,17 +12,25 @@ function DefaultComponent({ node }: { node: GenericNode }) {
   );
 }
 
+export function selectRenderer(renderers: NodeRenderersValidated, node: GenericNode) {
+  const componentRenderers = renderers[node.type] ?? renderers['DefaultComponent'];
+  const SpecificComponent = Object.entries(componentRenderers)
+    .reverse()
```
fwkoch, line 18: Why `reverse`?

packages/providers/src/renderers.tsx
```diff
@@ -0,0 +1,52 @@
...
+      throw new Error(
+        `Renderer for "${key}" must be either a function or an object containing a "base" renderer.`,
+      );
+    }
+  }
+
+  return validatedRenderers;
+}
+
+export function mergeRenderers(renderers: NodeRenderers[], validate: true): NodeRenderersValidated;
```
fwkoch, line 34: A few more docstrings - particularly for the functions in this file - would be helpful.

packages/myst-to-react/src/MyST.tsx
```diff
@@ -10,17 +12,25 @@ function DefaultComponent({ node }: { node: GenericNode }) {
   );
 }
 
+export function selectRenderer(renderers: NodeRenderersValidated, node: GenericNode) {
+  const componentRenderers = renderers[node.type] ?? renderers['DefaultComponent'];
+  const SpecificComponent = Object.entries(componentRenderers)
```
fwkoch, line 17:
```suggestion
  const specificComponent = Object.entries(componentRenderers)
```

packages/myst-to-react/src/links/index.tsx
```diff
@@ -134,8 +137,20 @@ export const linkBlock: NodeRenderer<TransformedLink> = ({ node }) => {
   );
 };
 
-const LINK_RENDERERS = {
-  link,
+const LINK_RENDERERS: NodeRenderers = {
+  link: {
```
fwkoch, line 141: I don't quite see why we need this nested structure as opposed to:
```ts
{
  link: SimpleLink,
  'link[protocol=github]': GithubLinkRenderer,
  'link[protocol=wiki]': WikiLinkRenderer,
  ...
}
```

I guess the nesting lets us first narrow down based on `type` so we are not doing `selector` checks against _all_ the `renderers`?

fwkoch: My next question is: would this work the same if we did something like:
```ts
{
  link: {
    base: SimpleLink,
    '[protocol=github]': GithubLinkRenderer,
    '[protocol=wiki]': WikiLinkRenderer,
    ...
  }
}
```

i.e. we don't need to repeat `"link"` since we are only dealing with `link` nodes at this point?

fwkoch (commented):
I think this makes sense. Now, instead of a simple `type: renderer` lookup, we have a more complex `type: { renderers }` structure where the basic renderer is under `base` and other renderers are under `selector` keys.

Keying off `selector` strings seems a little complicated (as opposed to, say, just `kind` or something) but it is quite flexible and powerful. Some of this logic is a little hard to parse without a bit more docstring-level documentation, though.

PR #477, opened 2024-09-24
 🤏 Improve theme for narrow screens

agoose77:
This PR sets some flex styling to avoid shrinkage causing issues at intermediate widths.
This PR does not address the need to perhaps tweak the menu breakpoints / move the theme button into the ellipsis menu.

Existing:
- Unnecessarily wide search bar
- Stretched theme icon
- Warped action button(s)

image

New: 
- Narrower search bar
- Non-growing/shrinking (square) theme icon
- Non-shrinking action button(s)

image

### DIFF
--- a/.changeset/tasty-books-applaud.md
+++ b/.changeset/tasty-books-applaud.md
@@ -1,0 +1,5 @@
+---
+'@myst-theme/site': patch
+---
+
+Fix small screens with search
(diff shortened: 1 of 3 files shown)


rowanc1:
Can you put screenshots in?!

stevejpurves:
@agoose77 is wide screen behaviour unchanged? even when there are a good list of nav items in the topbar?

agoose77:
@stevejpurves I've done some more testing on that, and I think it needs some work.

I think we need to start applying more conditional styling so that if users have nav items, we prefer compactness at intermediate sizes. Does anyone have any thoughts here? It's a bit of a feedback loop between nav and search at the moment.

agoose77:
I'm closing this for now -- we'll need to revisit.

PR #480, opened 2024-09-26
👩‍💻 Fix SSR styling of code-blocks

agoose77:
Recent theme work moved theme logic entirely client-side[^note], so that we can support client-only theme configuration (`localStorage`).

This PR is one "hacky" fix which duplicates each `<pre>` tag, and sets the dark and light styles respectively. Visibility is then class-driven. I don't immediately know if this is a performance nightmare.

[^note]: Technically SSR can render the proper state server-side, but for first-visits, we have to wait for the client to compute the proper state (and do some hydration-hacks to resolve the state).

### DIFF
--- a/packages/myst-to-react/src/code.tsx
+++ b/packages/myst-to-react/src/code.tsx
@@ -1,5 +1,4 @@
 import type { NodeRenderer } from '@myst-theme/providers';
-import { useThemeSwitcher } from '@myst-theme/providers';
 import { LightAsync as SyntaxHighlighter } from 'react-syntax-highlighter';
 import light from 'react-syntax-highlighter/dist/esm/styles/hljs/xcode.js';
 import dark from 'react-syntax-highlighter/dist/esm/styles/hljs/vs2015.js';
@@ -7,6 +6,8 @@ import { DocumentIcon } from '@heroicons/react/24/outline';
 import classNames from 'classnames';
 import { CopyIcon } from './components/index.js';
 import { MyST } from './MyST.js';
+import { useMemo, cloneElement } from 'react';
+import type { ComponentProps, CSSProperties } from 'react';
 
 type Props = {
   value: string;
@@ -32,8 +33,9 @@ function normalizeLanguage(lang?: string): string | undefined {
   }
 }
 
+type HighlightPropsType = ComponentProps<typeof SyntaxHighlighter>;
+
 export function CodeBlock(props: Props) {
-  const { isLight } = useThemeSwitcher();
   const {
     value,
     lang,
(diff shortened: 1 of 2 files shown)


packages/myst-to-react/src/code.tsx
```diff
@@ -1,12 +1,13 @@
 import type { NodeRenderer } from '@myst-theme/providers';
-import { useThemeSwitcher } from '@myst-theme/providers';
 import { LightAsync as SyntaxHighlighter } from 'react-syntax-highlighter';
 import light from 'react-syntax-highlighter/dist/esm/styles/hljs/xcode.js';
 import dark from 'react-syntax-highlighter/dist/esm/styles/hljs/vs2015.js';
 import { DocumentIcon } from '@heroicons/react/24/outline';
 import classNames from 'classnames';
 import { CopyIcon } from './components/index.js';
 import { MyST } from './MyST.js';
+import { useMemo, cloneElement } from 'react';
```
rowanc1, line 9:
```suggestion
import { useMemo } from 'react';
```

packages/myst-to-react/src/code.tsx
```diff
@@ -32,8 +33,9 @@ function normalizeLanguage(lang?: string): string | undefined {
   }
 }
 
+type HighlightPropsType = ComponentProps<typeof SyntaxHighlighter>;
+
 export function CodeBlock(props: Props) {
-  const { isLight } = useThemeSwitcher();
```
rowanc1, line 36: Is there a way to make this work?

I don't love having two of every code element in the DOM (and all of the additional nodes for syntax highlighting). Especially for notebooks. In many of the existing sites, we are all ready getting lighthouse warnings that there are many DOM nodes and that starts to kill performance.

Perhaps re-trigger this in the static builds, and live with the flicker for code.

agoose77: Yeah, I don't love the way it's needed either.

I think it should be possible to return a single DOM node when SSR if the cookie is set. That only helps subsequent visits.

What we probably need to do is look into pushing this state into CSS variables. We already do this for our tailwind styling, but I didn't immediately see a way to do it here. Perhaps we need a custom theme for the code block.

agoose77:
@rowanc1 I've just lifted the CSS from highlightjs directly into our styling. This means that we don't need to duplicate any nodes. Are you comfortable with this change?

packages/myst-to-react/src/code.tsx
```diff
@@ -1,13 +1,11 @@
 import type { NodeRenderer } from '@myst-theme/providers';
 import { LightAsync as SyntaxHighlighter } from 'react-syntax-highlighter';
-import light from 'react-syntax-highlighter/dist/esm/styles/hljs/xcode.js';
-import dark from 'react-syntax-highlighter/dist/esm/styles/hljs/vs2015.js';
```
agoose77, line 5: :vomiting_face:

PR #481, opened 2024-09-27
🎹 Improve keyboard accessibility

rowanc1:
This changes the focus to the element, rather than just scrolling to it.

### DIFF
--- a/packages/site/src/components/DocumentOutline.tsx
+++ b/packages/site/src/components/DocumentOutline.tsx
@@ -60,11 +60,14 @@ const Headings = ({ headings, activeId }: Props) => (
           href={`#${heading.id}`}
           onClick={(e) => {
             e.preventDefault();
-            const el = document.querySelector(`#${heading.id}`);
+            const el = document.querySelector(`#${heading.id}`) as HTMLElement | undefined;
             if (!el) return;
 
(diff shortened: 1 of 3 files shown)


packages/site/src/components/DocumentOutline.tsx
```diff
@@ -60,11 +60,14 @@ const Headings = ({ headings, activeId }: Props) => (
...
           onClick={(e) => {
             e.preventDefault();
-            const el = document.querySelector(`#${heading.id}`);
+            const el = document.querySelector(`#${heading.id}`) as HTMLElement | undefined;
             if (!el) return;
 
             el.scrollIntoView({ behavior: 'smooth' });
             history.replaceState(undefined, '', `#${heading.id}`);
+            // Changes keyboard tab-index location
+            if (el.tabIndex === -1) el.tabIndex = -1;
```
agoose77, line 69: @rowanc1 did you want to change the tab index?

rowanc1: This actually does, -1 is the `get` default, but it isn't actually set. When you set it to -1 that actually changes the behaviour in the browser!

agoose77: Ah, I see!

PR #489, opened 2024-10-23
🗂 Support url slugs with multiple path segments

fwkoch:
This PR includes the theme updates required to support nested folder structures in URLs. See https://github.com/jupyter-book/mystmd/issues/670

If MyST site content is built with the option `url_folders` (see https://github.com/jupyter-book/mystmd/pull/1601), the folder structure will be reflected as dot-delimited parts in the page slugs. With this PR, the themes will load the correct page based on slash-delimited URLs.

For example, the file `folder1/folder2/page.md` will be given the slug `folder1.folder2.page` and will be available at URL `/folder1/folder2/page`.


packages/site/src/components/Navigation/TableOfContentsItems.tsx
```diff
@@ -35,10 +35,16 @@ function nestToc(toc: Heading[]): NestedHeading[] {
   return items;
 }
 
+function pathnameMatchesHeading(pathname: string, heading: Heading, baseurl?: string) {
+  const headingPath = withBaseurl(heading.path, baseurl);
+  if (pathname && headingPath === `${pathname}/index`) return true;
```
fwkoch, line 40: There's a few small changes in here that allow `index` pages to be accessible at the folder-path url.

For example, the file `folder1/folder2/index.md` is available at `/folder1/folder2` (and `/folder1/folder2/index` redirects back to `/folder1/folder2`).

themes/article/app/routes/$.tsx
```diff
@@ -47,14 +47,15 @@ export const meta: V2_MetaFunction<typeof loader> = ({ data, matches, location }
 export const links: LinksFunction = () => [KatexCSS];
 
 export const loader: LoaderFunction = async ({ params, request }) => {
-  const [first, second] = new URL(request.url).pathname.slice(1).split('/');
-  const projectName = second ? first : undefined;
-  const slug = second || first;
+  const [first, ...rest] = new URL(request.url).pathname.slice(1).split('/');
   const config = await getConfig();
-  const project = getProject(config, projectName ?? slug);
+  const project = getProject(config, first);
```
fwkoch, line 52: I learned about some strangeness in the `article` theme while making these changes: the article theme does not work correctly with project slugs. I think it is ok if project slugs just do not work with `article` theme, but we should make that more clear. For example, here we should not even try to load the project based on the URL; it should always just be the default single project.

However, for this PR, I tried as best as I could to keep parity with the old functionality.

themes/article/app/routes/($a).($b).($c).($d).$slug[.json].tsx
```diff
@@ -13,23 +13,27 @@ function api404(message = 'No API route found at this URL') {
...
-  else if (slug === 'myst.search') {
+  else if (rest.length === 0 && first === 'myst.search') {
     const search = await getMystSearchJson();
     if (!search) {
       return json({ message: 'myst.search.json not found', status: 404 }, { status: 404 });
     }
     return json(search);
   }
+  const slug = [first, ...rest].join('.');
   const data = await getPage(request, { slug }).catch(() => null);
```
Comment (fwkoch) [inline themes/article/app/routes/($a).($b).($c).($d).$slug[.json].tsx:37]: (Look at this for example - on this route, we do not mention anything about `project`. The other route should probably be the same way.)

themes/book/app/routes/($project)_.($a).($b).($c).($d).$slug[.json].tsx
```diff
@@ -33,10 +36,10 @@ export const loader: LoaderFunction = async ({ request, params }) => {
   }
   const config = await getConfig();
   const flat = isFlatSite(config);
-  const data = await getPage(request, {
-    project: flat ? project : (project ?? slug),
-    slug: flat ? slug : project ? slug : undefined,
-  });
+  const project = flat ? undefined : first;
```
Comment (fwkoch) [inline themes/book/app/routes/($project)_.($a).($b).($c).($d).$slug[.json].tsx:39]: In the `book` theme, all the project slug stuff works correctly.

stevejpurves:
@fwkoch has this been tested with notebooks in the mix, and specifically with notebooks in a subfolder being pages on the site and one or more outputs having been included as a figure? I'm curious about how any changes to the `slug` and `location` fields might impact that.

fwkoch:
@stevejpurves - I have not tested these with notebook compute. I'll do that. `location` has not been changed at all. `slug` is now dot-delimited in JSON files, but if you follow urls using that slug, they should resolve in the same way as if the dots were slashes. Hopefully that means everything still works correctly...? But I'll confirm.

rowanc1:
We tested that the nested notebooks load at the correct jupyter path!

packages/common/src/utils.ts
```diff
@@ -40,9 +40,10 @@ export function getProjectHeadings(
     },
     ...project.pages.map((p) => {
       if (!('slug' in p)) return p;
+      const slug = p.slug?.replace(/\.index$/, '').replaceAll('.', '/');
```
rowanc1, line 43: Lets get this into an option in myst-common

rowanc1 (approved):

themes/article/app/utils/loaders.server.ts
```diff
@@ -62,11 +62,19 @@ export async function getPage(
...
-    return redirect(projectName ? `/${projectName}` : '/');
+    throw redirect(projectName ? `/${projectName}` : '/');
+  }
+  if (opts.slug?.endsWith('.index') && opts.redirect) {
+    const newSlug = opts.slug.slice(0, -6);
+    throw redirect(projectName ? `/${projectName}/${newSlug}` : `/${newSlug}`);
+  }
+  let slug = opts.loadIndexPage || opts.slug == null ? project.index : opts.slug;
+  let loader = await getStaticContent(projectName, slug).catch(() => null);
+  if (!loader) {
```
rowanc1, line 73:
```suggestion
  if (!loader) {
    // If you haven't loaded the first time, try the `.index`
```

PR #492, opened 2024-10-30
🗂 Update slugToUrl in article theme

fwkoch:
Most of the URL folder structure changes were landed in https://github.com/jupyter-book/myst-theme/pull/489

This just adds `slugToUrl` in a couple places that were missed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,10 @@
     "changeset": "changeset",
     "version": "changeset version && npm install",
     "publish": "npm run clean && npm run build -- --force && changeset publish && git push --follow-tags",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -38,6 +41,7 @@
     "concurrently": "^8.2.0",
     "eslint-config-curvenote": "^0.0.4",
     "npm-run-all": "^4.1.5",
+    "npm-upgrade-monorepo": "^1.2.6",
```
fwkoch, line 44: This is nice for upgrading all myst dependencies. Is it too much to have here though? Causes package-lock to grow a lot...

PR #500, opened 2024-11-19
🪄 Add more `class` support to directives

agoose77:
This PR should increase the number of directives that support `class` and `identifier` in rendering. I doubt I've got them all, but I created a test case from the directives list: https://mystmd.org/guide/directives

<details><summary>Details</summary>

````markdown
:::{aside} An Optional Title
:class: red

This is an aside. It is not entirely relevant to the main article.
:::

:::{tip}
:class: red
Try changing `tip` to `warning`!
:::

:::{bibliography}
:class: red

:::

:::{blockquote}
:name: quote
:class: red

Hello

-- world
:::

:::{code} python
:class: red

x + y
:::

:::{div}
:class: red

div
:::

:::{dropdown}
:class: red

dropdown
:::

:::{embed} #quote
:class: red

:::

:::{figure} https://picsum.photos/512
:class: red

:::

:::{glossary}
:class: red

Foo
: Bar

:::

:::{iframe} https://google.com
:class: red
:::

:::{include} test.md
:class: red

:::

:::{mermaid}
:class: red

graph LR;
dot --> dash;
:::

```{myst}
:class: red

Try this!
```

```{table}
:class: red

| Month    | Savings |
| -------- | ------- |
| January  | $250    |
| February | $80     |
| March    | $420    |

```


````

</details>

### DIFF
--- a/.changeset/tidy-ducks-breathe.md
+++ b/.changeset/tidy-ducks-breathe.md
@@ -1,0 +1,7 @@
+---
+'@myst-theme/diagrams': patch
+'myst-demo': patch
+'myst-to-react': patch
+---
+
+Add more support for class and identifier
--- a/packages/diagrams/src/index.tsx
+++ b/packages/diagrams/src/index.tsx
@@ -10,7 +10,15 @@ async function parse(id: string, text: string): Promise<string> {
   });
 }
 
-export function MermaidRenderer({ id, value }: { value: string; id: string }) {
+export function MermaidRenderer({
+  id,
+  value,
+  className,
+}: {
+  value: string;
+  id: string;
+  className?: string;
+}) {
   const key = useId();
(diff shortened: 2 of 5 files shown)


agoose77:
Could someone clarify whether we expect AST nodes to always have `html_id`, or whether the current fallback pattern
`node.html_id || node.identifier` is expected? I mentioned this somewhere else (probably Discord), but I don't think we touched upon it.

rowanc1:
If there is an identifier, there should be an html_id always. If there isn't that is a bug. Those in the theme and there for legacy likely., or if we need the node to be identified, prefer the I'd and fall back to the key in the order: html_id, identifier, key.

fwkoch (approved):
Nice, all pretty simple stuff, propagating classes, identifiers, etc in more places.

PR #502, opened 2024-11-25
➖ Hide scrollbar in PrimarySidebar

jnywong:
Ref https://github.com/jupyter-book/mystmd/issues/1672

### DIFF
--- a/packages/site/src/components/Navigation/PrimarySidebar.tsx
+++ b/packages/site/src/components/Navigation/PrimarySidebar.tsx
@@ -174,7 +174,7 @@ export const PrimarySidebar = ({
           },
         )}
       >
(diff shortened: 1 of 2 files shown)


agoose77:
Thanks for this PR @jnywong!

I need to do some due diligence to think through what the A11y ramifications of making this change will be. I'll circle back round once I know more!

jnywong:
Good shout regarding a11y principles! Happy to help with thinking around that.

jnywong:
Okay yeah, this change is not very a11y-friendly!

Happy to repurpose this into making the scrollbar theme change when switching to dark mode, but need some pointers on where to start because the light/dark mode theme switcher looks a little more complicated to contribute to at first glance.

PR #508, opened 2024-12-17
🎨 Style scrollbar

jnywong:
Closes https://github.com/jupyter-book/mystmd/issues/1672.

Here is what the changes look like in both light and dark mode for a Chromium-based browser:

Screenshot 2024-12-17 at 09 44 37
Screenshot 2024-12-17 at 09 44 41

No changes were needed for Firefox, hence the use of `-webkit-scrollbar` only.

Screenshot 2024-12-17 at 09 48 19

### DIFF
--- a/packages/site/src/components/Navigation/PrimarySidebar.tsx
+++ b/packages/site/src/components/Navigation/PrimarySidebar.tsx
@@ -174,7 +174,7 @@ export const PrimarySidebar = ({
           },
         )}
       >
-        <div className="flex-grow py-6 overflow-y-auto">
+        <div className="flex-grow py-6 overflow-y-auto primary-scrollbar">
           {nav && (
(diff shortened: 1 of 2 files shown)


rowanc1 (approved):
🚀

agoose77:
I think the Chromium styling is perhaps better than the Firefox after this PR w.r.t. the background colour of the scrollbar. Should we follow up and touch that too?

choldgraf:
Jenny your website looks awesome!

jnywong:
> I think the Chromium styling is perhaps better than the Firefox after this PR w.r.t. the background colour of the scrollbar. Should we follow up and touch that too?

I can easily go back in and take a look at that. I initially decided against it because I wanted to make CSS changes as minimal as possible.

Issue #513, opened 2025-01-07
Support return navigation for footnotes on mobile


agoose77:
We should make it possible to return from the navigation point for footnotes, if possible, on mobile displays.

> I've really had a pet peeve about footnotes on mobile and this does the crime. If you click a footnote and it jumps very far away and doesn't have a return navigation, I have left your page usually after the second time I see a footnote and didn't note my exact scroll. A small thing but pretty please don't make me have to hunt for where I was...
> -- https://news.ycombinator.com/item?id=42573388#42604612

rowanc1:
![Image](https://github.com/user-attachments/assets/93bf8cc7-0395-4671-84dd-6a845738f1d9)

That is possible, and is hooked up to keyboard navigation for accessibility.

We should just make these back links visible on mobile.

agoose77:
@rowanc1 you star! I didn't notice the back-link (and didn't look at myst-theme). I'll rework the title to reflect the true missing feature.

PR #516, opened 2025-01-08
👈 Disable `hover` modifier for mobile in `HashLink`

agoose77:
This simple PR adds a `[@media(hover:hover)]` modifier to the class-names for HashLink instances that use `hover`. The following components are affected:

- `Heading`
- `SolutionRenderer`
- `References`
- `Footnotes`
- `Abstract`
- `Bibliography`
- `Backmatter`
- `Keywords`

This PR was motivated by #513, but I think the same logic holds -- our mobile site should not lose interactivity. It does change the aesthetics, though:

<details><summary>Screenshots</summary>

</details> 

Forgive the HiDPI images -- I was already playing around with this PR on mobile.

### DIFF
--- a/.changeset/heavy-geese-flow.md
+++ b/.changeset/heavy-geese-flow.md
@@ -1,0 +1,5 @@
+---
+'myst-to-react': patch
+---
+
+Always show hover HashLink on mobile
--- a/packages/myst-to-react/src/hashLink.tsx
+++ b/packages/myst-to-react/src/hashLink.tsx
@@ -61,7 +61,7 @@ export function HashLink({
   id?: string;
   kind?: string;
   title?: string;
-  hover?: boolean;
+  hover?: boolean | 'desktop';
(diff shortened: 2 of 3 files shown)


rowanc1:
Not sure I totally love the pilcrows showing up everywhere, I think they do take away from the readability?

The footnotes are a loss of functionality for sure. The hashes on titles etc are a much more advanced feature (i.e. you are wanting to share a link to that section from mobile).

Thoughts on not enabling them?

At the least, we should have an additional option in the component or `hover: boolean | 'desktop'` that enables the current behaviour.

agoose77:
I wasn't a huge fan either, but I felt on balance it was a useful feature. However, I realised that mobile browsers are usually smart enough to treat a touch as a persistent hover, so this is _only_ needed for footnotes!

I'll make a change in line with your suggestion, which should only materially change footnote handling on mobile.

PR #521, opened 2025-01-13
🔘 Add support for buttons

jnywong:
Ref: https://github.com/2i2c-org/infrastructure/issues/5346

c.f. jupyter-book/mystmd#1773

image

Co-authored-by: Angus Hollands <angus@oieltd.com>

### References

- Related PR: https://github.com/jupyter-book/mystmd/pull/1773
- closes https://github.com/jupyter-book/mystmd/issues/1654


packages/myst-to-react/src/crossReference.tsx
```diff
@@ -150,6 +153,7 @@ export function CrossReferenceHover({
     const el = document.getElementById(htmlId);
     scrollToElement(el, { htmlId });
   };
+  const isButtonLike = !!className?.includes('button');
```
rowanc1, line 156: Do we need to do something that parses the class list a bit more carefully here?

agoose77: Yes, actually. I don't like this at all! 

(coffee brain)

rowanc1: e.g. a string with `border shadow not-a-button` will turn on behaviour.

There is code for this in the admonitions I think.

packages/myst-to-react/src/links/index.tsx
```diff
@@ -12,6 +12,7 @@ import { RRIDLink } from './rrid.js';
 import { RORLink } from './ror.js';
 import { GithubLink } from './github.js';
 import { MyST } from '../MyST.js';
+import classNames from 'classnames';
```
rowanc1, line 15:
```suggestion
```

rowanc1 (approved):

choldgraf:
+1 on getting these in and released so we can kick the tires on them and iterate from there. We can put these into next.jupyterbook.org to see how that goes.

agoose77:
Sounds good! @jnywong is going to write up some small docs, and then she can merge!

agoose77:
@rowanc1 this is blocked until we make a new first of `myst-ext-button` (it's needed by the demo component). Any suggestions on procedure for unblocking?

rowanc1:
This is now published and `ebp-bot` has access to work in CI.

agoose77:
Kicking CI

PR #522, opened 2025-01-14
Book theme options to specify theme and disable theme toggle button

gvarnavi:
Adds the following two options to the book-theme:
```yaml
site:
  template: book-theme
  options:
    theme: 'dark'
    hide_theme_toggle: true
```
- `theme` can either be 'light' or 'dark' and forces the initial theme (as opposed to using local storage)
- `hide_theme_toggle` is a boolean and disables the top-right theme toggle

### DIFF
--- a/packages/site/src/components/Navigation/TopNav.tsx
+++ b/packages/site/src/components/Navigation/TopNav.tsx
@@ -103,7 +103,7 @@ export function NavItems({ nav }: { nav?: SiteManifest['nav'] }) {
   );
 }
 
-export function TopNav({ hideToc, hideSearch }: { hideToc?: boolean; hideSearch?: boolean }) {
+export function TopNav({ hideToc, hideSearch, hideThemeToggle }: { hideToc?: boolean; hideSearch?: boolean; hideThemeToggle?: boolean }) {
   const [open, setOpen] = useNavOpen();
   const config = useSiteManifest();
   const { title, nav, actions } = config ?? {};
@@ -131,7 +131,7 @@ export function TopNav({ hideToc, hideSearch }: { hideToc?: boolean; hideSearch?
           <NavItems nav={nav} />
           <div className="flex-grow block"></div>
           {!hideSearch && <Search />}
-          <ThemeButton />
+	  {!hideThemeToggle && <ThemeButton />}
           <div className="block sm:hidden">
             <ActionMenu actions={actions} />
           </div>
--- a/themes/book/app/root.tsx
+++ b/themes/book/app/root.tsx
@@ -3,6 +3,7 @@ import tailwind from '~/styles/app.css';
 import thebeCoreCss from 'thebe-core/dist/lib/thebe-core.css';
 import { getConfig } from '~/utils/loaders.server';
 import type { SiteLoader } from '@myst-theme/common';
+import { Theme } from '@myst-theme/common';
 import {
   Document,
   responseNoSite,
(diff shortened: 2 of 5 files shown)


agoose77:
Hi @gvarnavi,

Theme selectors are an important part of accessibility (though there is nuance to that point). Is there a particular reason that you want to disable this preference for your users of the theme?

agoose77:
Thanks for the quick reply @gvarnavi. It took me a bit longer to circle round here than usual.

The question around removing accessibility from the user is tricky, and I don't have all the answers. I can see why it is useful for your use case! I'll bring this up at our developers meeting.

gvarnavi:
I wanted to quickly follow-up on this. I just gave a [conference talk](https://gvarnavides.com/2025apmc-transfer-of-information/parallax-imaging) using a slightly-modified book-theme found here: https://github.com/gvarnavi/custom-myst-book-theme.

It was very well received and I'd be happy to prototype this further if you think that would be useful to contribute a `presentation-theme` template, in which case we can close this PR. @rowanc1

fperez:
Thanks @gvarnavi for this PR and your Discord help!

To add my 👍 to at least the default selector option for the theme(s). I just started building [my website](https://fernandoperez.org) with MyST and would very much want it to default to light mode.

I think as long as the toggle is visible by default, having an option to disable it on occasion is reasonable and should be available to authors. Someone may well build a site so customized that it only makes sense in one mode, and that seems to me a valid option to offer authors.

choldgraf:
Copied over from Discord:

I think what Fernando is saying is:

- In an ideal world we'd all have light/dark accessible sites
- Many people have other site artifacts (images, logos, etc) that clash with one or another theme
- They don't have the capacity to fix all those artifacts themselves (e.g. create a "dark theme version" of all their relevant images)
- So they'd like the ability to force a certain theme in order to sustainably achieve the visual aesthetic they want

IMO, this is a choice between "forcing authors to allow users to experience their site in a way that the author doesn't want" vs. "allowing an author to force a reader to experience a theme they don't prefer".

My feeling is that we should focus on empowering authors to create the reading experience they want. MyST should have **sensible defaults**, but err on the side of empowering authors to create the reading experience they want, as long as it is sustainable to maintain.

agoose77:
> they'd like the ability to force a certain theme in order to sustainably achieve the visual aesthetic they want

How do we feel about a compromise here, which is to choose an author-preferred default _but_ maintain the ability for users to toggle that theme? I can see a counterargument being that the presence of a theme button might tacitly imply "support" for both, but on balance I wonder if that holds much less weight than the argument for "accessibility" (in quotes, because it's not well defined for dark-light).

choldgraf:
What if we added the ability to set the default theme, and for "removing" the button, we documented the following solution:

1. Create custom.css, put this in it:

    ```css
    button.theme {
      display: none;
    }
    ```
2. Put this in your myst.yml file:
    ```yaml
    site:

      options:
        style: custom.css
    ```

fperez:
I'm 👍 on @agoose77's compromise - I think this gives both ends of the deal (authors and readers) enough options to tug things in their preferred direction. Readers who really want a mode an author didn't optimize for can just live with the odd logo/image/whatever not being too readable, but authors can't fully "impose" their choice on all readers.

Also 👍 to @choldgraf's docs indication.

Thanks all!

stefanv:
We will soon have `myst-*` classes on the various elements, making it even easier to hide elements. See https://github.com/jupyter-book/myst-theme/pull/667

PR #528, opened 2025-02-04
🚀 Add `awesomebar`-like UX to launch button

### LINKED ISSUE #507: Detect Hub vs Binder in launch button
We can programmatically detect the presence of a Hub vs Binder from querying API endpoints in the URL.

agoose77:
This PR closes #507 by changing the launch button UI to detect the target. 

It is motivated by https://github.com/2i2c-org/infrastructure/issues/5348 in 2i2c's sprint planning.

This feature is intended to bring us closer to a time in which the distinction between JupyterHub and BinderHub becomes increasingly less relevant. 

There are some nuances encoded in this UX; we assume that the JupyterHub uses `nbgitpuller`. In future, we (2i2c) would like to unify the features of nbgitpuller and mybinder so that the language for addressing contents can be drawn closer together.

https://github.com/user-attachments/assets/623c7b09-573a-4442-8130-51883883dd1d

### To Do
- [x] Update scienceicons with JupyterHub and/or BinderHub icons https://github.com/continuous-foundation/scienceicons/pull/6
- [x] Use new icons in launch button
- [x] Remove special casing of hub/binder names (besides https://mybinder.org)

In future, this feature should probably use a reducer — the business logic already feels more complicated than I would like.

### DIFF
--- a/.changeset/famous-wasps-bow.md
+++ b/.changeset/famous-wasps-bow.md
@@ -1,0 +1,5 @@
+---
+'@myst-theme/frontmatter': patch
+---
+
+Add note about JHub requirements
--- a/.changeset/wild-ghosts-tease.md
+++ b/.changeset/wild-ghosts-tease.md
@@ -1,0 +1,5 @@
+---
+'@myst-theme/frontmatter': patch
+---
+
+Add awesomebar support to launch button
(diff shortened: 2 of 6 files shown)


choldgraf:
That is so slick!

PR #530, opened 2025-02-06
Experimental: support disabling pretty-urls (for readthedocs)

ebolyen:
A potential solution to the assumption of URLs without extensions (pretty urls) being served from a bare html file noted in https://github.com/jupyter-book/mystmd/issues/984.

**Limitations:**
 - First time working with this theme, so I probably missed something.
 - No tests yet (want to get feedback on the approach)
 - I've only changed the "book" theme.
 - Doesn't handle the flyout menu in readthedocs, I think that may only need a react upgrade, which I will probably test out soon.

**Approach**
 - Add a site option for `pretty_urls`, it is by default interpreted to be `true`. (I'm not sure if there's a better way to cause a default, but it didn't seem like it since we're relying on the remix loader which just fetches the config.)
 - When false, the book theme will set a React context which various components will look at.
 - I looked at each reference to base_url and worked out if it was relevant. (There is one section that uses base_url in a way that I think might be a bug, but I'm not familiar enough with what "internal" means to this code.)

**Example**
https://genome-sampler.readthedocs.io/en/latest/

### DIFF
--- a/packages/myst-to-react/src/card.tsx
+++ b/packages/myst-to-react/src/card.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import type { NodeRenderer } from '@myst-theme/providers';
 import classNames from 'classnames';
-import { useLinkProvider, useBaseurl, withBaseurl } from '@myst-theme/providers';
+import { useLinkProvider, useBaseurl, withBaseurl, usePrettyUrl, withPrettyUrl } from '@myst-theme/providers';
 import { MyST } from './MyST.js';
 import type { GenericNode } from 'myst-common';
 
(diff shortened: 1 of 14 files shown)


ebolyen:
Thanks @changeset-bot, I'll switch this to a draft.

ebolyen:
Is it intended that regular markdown links do not produce a hover?

This code: 
https://github.com/jupyter-book/myst-theme/blob/2d9882099eebcebb3dbd205a773efca689406402/packages/myst-to-react/src/links/index.tsx#L18-L24

Does not generally find internal pages, since the "path" doesn't include the project so it never finds a default project to then root through the pages.

I can't tell if this behavior is just a bit wrong or if its intended for a non-book (article only) audience.

choldgraf:
I believe that this is now fixed, we have readthedocs working and jupyterbook.org now serves from that!

PR #535, opened 2025-02-12
🏷️ Extend MyST component to accept  `className` argument

agoose77:
This PR is both useful, and required to enable us to drop the `ContentBlocks` component.


rowanc1 (commented):
Worth doing identifier at the same time?

agoose77:
This only allows JSX to pass in className — I deliberately haven't done a pass to pull all `node.className` up too. With that in mind, do you think we need to be able to set `id` as well (via JSX)?

rowanc1:
Yeah, I think we should extend to id/class on everything. Can also do this in a few steps, but the places you just are going into now are exactly the same.

I certainly like breaking this up into a few small and quick PRs.

packages/myst-to-react/src/MyST.tsx
```diff
@@ -20,18 +20,24 @@ export function selectRenderer(renderers: NodeRenderersValidated, node: GenericN
...
     const Component = selectRenderer(renderers, ast);
-    return <Component key={ast.key} node={ast} />;
+    return <Component key={ast.key} node={ast} className={className} />;
   }
   return (
     <>
       {ast?.map((node) => {
         const Component = selectRenderer(renderers, node);
-        return <Component key={node.key} node={node} />;
+        return <Component key={node.key} node={node} className={className} />;
```
rowanc1, line 40: Interesting. You are moving this from on the node to also coming in through a prop.

Maybe ID doesn't make sense.

rowanc1 (approved):
Looks good.

Issue #474, opened 2024-09-17
Landing Pages with widget layouts


choldgraf:
Landing pages serve a different purpose than most other documentation pages, and thus they have a different structure than the content-heavy structure that makes up most MyST sites.

For example, they tend to be more UI-heavy, and have elements like Hero boxes, splash pages, etc. They also often get rid of the left/right sidebars so that users focus on the content and don't get distracted.

A good example of this is the MySTMD landing page itself, which has this "blockier" and UI-heavy structure:

<img src="https://github.com/user-attachments/assets/746022f4-8c2d-4430-bf5a-a5d8e6a33375" width=400 />

### Suggestion

- We document a workflow for landing pages using the building blocks that are already there (e.g. grids, cards, etc) to show off some best practices.
- We create a list of UI components to develop that add different landing page-like functionality (e.g. hero cards, sliding galleries, etc)

Then we could use this as the tracking issue around a "landing page" use-case and chip away at the problem with a combination of new development + documentation.
 

### Implement with tailwind components?

The [Tailwind component library](https://tailwindui.com/components) might be the easiest way to enable this in a flexible-enough manner. Could we build a small library of directives that basically turn themselves into Tailwind components?

For example the [hero component of tailwind](https://tailwindui.com/components/marketing/sections/heroes) looks like this:

![Image](https://github.com/user-attachments/assets/21f5b8c2-3f54-4746-a82f-1321dc16f819)

And is enabled with the following HTML.

```html
      <div class="hidden sm:mb-8 sm:flex sm:justify-center">
        <div class="relative rounded-full px-3 py-1 text-sm/6 text-gray-600 ring-1 ring-gray-900/10 hover:ring-gray-900/20">
          Announcing our next round of funding. <a href="#" class="font-semibold text-indigo-600"><span class="absolute inset-0" aria-hidden="true"></span>Read more <span aria-hidden="true">&rarr;</span></a>
        </div>
      </div>
      <div class="text-center">
        <h1 class="text-balance text-5xl font-semibold tracking-tight text-gray-900 sm:text-7xl">Data to enrich your online business</h1>
        <p class="mt-8 text-pretty text-lg font-medium text-gray-500 sm:text-xl/8">Anim aute id magna aliqua ad ad non deserunt sunt. Qui irure qui lorem cupidatat commodo. Elit sunt amet fugiat veniam occaecat.</p>
        <div class="mt-10 flex items-center justify-center gap-x-6">
[...]

PR #536, opened 2025-02-12
🧱 Drop `ContentBlocks` and move functionality to `MyST`

agoose77:
This PR removes the special-casing of `block` nodes, and replaces it with the `<MyST ast={...}/>` component. The intention here is to make it possible to override / extend this in other PRs, such as #474.


packages/myst-to-react/src/block.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { SourceFileKind } from 'myst-spec-ext';
+import type { GenericParent } from 'myst-common';
+import classNames from 'classnames';
+import {
+  NotebookClearCell,
+  NotebookRunCell,
+  NotebookRunCellSpinnerOnly,
+  executableNodesFromBlock,
+} from '@myst-theme/jupyter';
+import { useGridSystemProvider, usePageKindProvider } from '@myst-theme/providers';
```
agoose77, line 12: Need to check if these deps are actually in package.json

packages/myst-to-react/src/block.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { MyST } from './MyST.js';
+import { SourceFileKind } from 'myst-spec-ext';
+import type { GenericParent } from 'myst-common';
+import classNames from 'classnames';
+import {
+  NotebookClearCell,
+  NotebookRunCell,
+  NotebookRunCellSpinnerOnly,
+  executableNodesFromBlock,
+} from '@myst-theme/jupyter';
```
rowanc1, line 11: This can't be a dep of myst-to-react! Can you pull these together at the theme level instead.

A lot of lighter-weight integrations don't need jupyter, and that comes with a lot of overhead and integration main in some deployments.

i.e. the myst-to-react has a basic block renderer. Jupyter, if included by a theme has a more specific block renderer which takes priority.

packages/providers/src/pageKind.tsx
```diff
@@ -0,0 +1,19 @@
+import React, { useContext } from 'react';
+import type { SourceFileKind } from 'myst-spec-ext';
+
+const PageKindContext = React.createContext<SourceFileKind | undefined>(undefined);
+
+export function PageKindProvider({
```
rowanc1, line 6: I would very much like to refactor to `ArticleProvider` from `ReferencesProvider` rather than introduce a new provider. I see why you went this way, but I think that it complicates downstream quite a bit to introduce many smaller providers (already theme/thebe/references/busyscope are a lot to handle).

Happy to take a crack later this week if that doesn't make sense.

packages/myst-to-react/src/block.tsx
stevejpurves: wherever this comes in leads to a patch-package in order just to build too :( 
Composing note renderers at theme level makes sense to me and is more scalable for core-myst functionality

agoose77: Yeah, I noticed the recursion when updating the deps!

I think this makes sense, thanks both!

agoose77: I've implemented a trivial block renderer (removing the Jupyter logic) in `myst-to-react`, and then moved the existing block renderer to `@myst-theme/jupyter`. I opted not to export the Jupyter renderer as part of the default exports; the `@myst-theme/site` package pulls these in for all themes, and based on this discussion I'm thinking that we might want to avoid that requirement. Perhaps in future we'll pull more renderer choices down to the themes.

packages/providers/src/pageKind.tsx
agoose77: There's a balance to be struck between having god-providers and having a component tree that looks like you could hang baubles upon it. I sometimes lean towards the latter in my preferences.

Given that we're mainly optimising for ourselves at the moment, I think it's a good nudge to fuse this with `ReferencesProvider`. I've updated the PR!

packages/jupyter/src/block.tsx
```diff
@@ -0,0 +1,83 @@
...
+      id={node.key}
+      node={node}
+      className={classNames(node.class, className)}
+    />
+  );
+};
+
+const NOTEBOOK_RENDERERS: NodeRenderers = {
+  block: {
+    'block[kind=notebook-code],block[kind=notebook-content]': NotebookBlockRenderer,
```
agoose77, line 79: The logic here is a little indirected, so let me explain:

1. MD files have `notebook-code` and otherwise blocks with no `kind`.
2. IPyNB files have `notebook-code` and `notebook-content` blocks.
3. An article may contain `notebook-code` blocks from the `code-cell` directive even though it is not considered a notebook (i.e., no frontmatter kernelspec).

This component therefore only renders for code-cells (both ipynb and .md) and content-cells (ipynb). The benefit of this is that we can use the block-kind to figure out if e.g. the cell is executable. @stevejpurves I am not sure if this is a breaking change, as historically we used a function.

This means that if a Markdown block has somehow set `hide-cell`, e.g. 
````markdown
+++ { "tags": ["hide-cell"]}
## Ready to dive right in?
Some prose...
````
it will render as a dropdown with `Block >` instead of `Notebook cell >`. I think that's OK as it's a niche case.

File: packages/site/src/components/ContentBlocks.tsx [hunk 5]
```diff
@@ -1,65 +1,11 @@
...
-  if (node.visibility === 'hide') {
-    return <Details title="Notebook Cell">{block}</Details>;
-  }
-  return block;
-}
+import { MyST } from 'myst-to-react';
+export { Block } from 'myst-to-react';
 
+/**
+ * @deprecated This component is not maintained, in favor of generic MyST component
```
agoose77, line 7:
```suggestion
 * @deprecated This component is not maintained, in favor of the generic `MyST` component
```

packages/providers/src/article.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const data = useContext(ArticleContext);
+  return data?.frontmatter;
+}
+
+export function usePageKind() {
+  const data = useContext(ArticleContext);
+  return data?.kind ?? SourceFileKind.Article;
+}
+
+export { ArticleProvider as ReferencesProvider };
```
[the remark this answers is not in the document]
rowanc1: Did you change the interface?

I would love to see a `@deprecated` on this.

themes/book/app/root.tsx
```diff
@@ -10,14 +10,19 @@ import {
...
 export { AppErrorBoundary as ErrorBoundary } from '@myst-theme/site';
 import { createSearch as createMiniSearch } from '@myst-theme/search-minisearch';
 import { Outlet, useLoaderData } from '@remix-run/react';
-import { SearchFactoryProvider } from '@myst-theme/providers';
+import { SearchFactoryProvider, mergeRenderers } from '@myst-theme/providers';
+import type { NodeRenderers } from '@myst-theme/providers';
 import type { ISearch, MystSearchIndex } from '@myst-theme/search';
 import { SEARCH_ATTRIBUTES_ORDERED } from '@myst-theme/search';
-import { useCallback, useMemo } from 'react';
+import { BLOCK_RENDERERS } from '@myst-theme/jupyter';
```
rowanc1, line 22: Can we call these `JUPTYER_RENDERERS`, I suspect we will need to change the `inlineOutputs` in the near future, so they won't all be blocks.

agoose77: Sure! It might be a bit weird that we also have the other renderers like output and embed, but we can figure that out later.

packages/providers/src/article.tsx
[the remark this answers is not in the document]
agoose77: The interface is extended (compatible). But, we do need to deprecate!

themes/book/app/root.tsx
agoose77: @rowanc1 I've exported the old `OUTPUT_RENDERERS` so that the `site` package continues to pull in only the output renderers, and not the new block renderer. We should look to deprecate that in future (probably just a breaking change).

agoose77:
@stevejpurves are you happy to merge this as-is? IIRC it got the green light from @rowanc1 in yesterday's meeting, but you're the one who should merge this in case there are unforeseen consequences.

rowanc1:
I will get this over the line today!

packages/jupyter/src/block.tsx
rowanc1: I put this in a comment.

rowanc1:
Have tested with thebe and various other paths. Everything still is working well.

https://github.com/user-attachments/assets/7c6eda9a-078c-43b1-889a-2c185e8a36af

PR #540, opened 2025-02-19
💇 Add support for user stylesheets

### LINKED ISSUE #321: Add support for loading user-defined CSS stylesheet
Users often want to define custom styling (CSS, SCSS, or SASS) for their site. This allows them to do lightweight theming without needing to create an entirely new theme. It also allows authors to make page-specific styling for certain UI components etc.

### Proposal

We should allow users to:

- Define custom CSS for their MyST sites.
- Allow users to write this as SCSS or SASS, including the use of variables.
- Ideally, allow for this either on a per-page basis (e.g. page metadata), or a project wide basis (e.g. `myst.yml`)

### Inspiration

1. Point to an (S)CSS file from `myst.yml`. [here's how sphinx does this](https://docs.readthedocs.io/en/stable/guides/adding-custom-css.html#how-to-add-custom-css-or-javascript-to-sphinx-documentation)
2. Add CSS directly to a page via HTML in the markdown (e.g. `<style>` tags)
3. Add CSS via frontmatter on the page or in `myst.yml` (e.g. a key like `site.style: <scss rules to add to the page or the website>`)

### Related

- https://github.com/jupyter-book/mystmd/issues/857 would allow users to add classes that could be styled with this CSS
- https://github.com/jupyter-book/mystmd/issues/1040 would allow people to add style directly to a page
- [Hugo transpiling SCSS docs](https://gohugo.io/hugo-pipes/transpile-sass-to-css/))
- [Quarto support for SASS variables](quarto.org/docs/output-formats/html-themes.html#sass-variables).

### Tasks and updates

- [ ] Automatically include e.g. `myst-theme.scss`

agoose77:
Fixes #321 by pulling in a user-defined stylesheet.

### DIFF
--- a/themes/book/app/routes/[custom-style.css].tsx
+++ b/themes/book/app/routes/[custom-style.css].tsx
@@ -1,0 +1,11 @@
+import type { LoaderFunction } from '@remix-run/node';
+import { getCustomStyleSheet } from '~/utils/loaders.server';
+
+export const loader: LoaderFunction = async (): Promise<Response> => {
+  const css = await getCustomStyleSheet();
+  if (!css) return new Response('No custom CSS stylesheet found', { status: 404 });
+  return new Response(
+    css.buffer,
+    css.contentType ? { headers: { 'Content-Type': css.contentType } } : undefined,
+  );
+};
--- a/themes/book/template.yml
+++ b/themes/book/template.yml
@@ -30,7 +30,7 @@ options:
     description: Disable the search
   - type: number
     id: outline_maxdepth
-    description: The maximum depth to show on the document outline, for example, `2` would show only two depths of headings (e.g. `<H1>` and `<H2>`).
+    description: The maximum depth to show on the document outline, for example, `2` would show only two depths of headings (e.g. `<H1>` and `<H2>`)
     min: 1
     max: 6
     integer: true
@@ -57,10 +57,13 @@ options:
     description: Plausible analytics key
   - type: boolean
     id: numbered_references
-    description: Show references as numbered, rather than in APA-style. Only applies to parenthetical citations.
+    description: Show references as numbered, rather than in APA-style. Only applies to parenthetical citations
   - type: boolean
     id: folders
-    description: Respect nested folder structure in URL paths.
+    description: Respect nested folder structure in URL paths
+  - type: file
+    id: style
+    description: Local path to a Tailwind CSS file
 build:
   install: npm install
   start: npm run start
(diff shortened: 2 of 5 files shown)


themes/book/template.yml
```diff
@@ -30,7 +30,7 @@ options:
     description: Disable the search
   - type: number
     id: outline_maxdepth
-    description: The maximum depth to show on the document outline, for example, `2` would show only two depths of headings (e.g. `<H1>` and `<H2>`).
+    description: The maximum depth to show on the document outline, for example, `2` would show only two depths of headings (e.g. `<H1>` and `<H2>`)
```
agoose77, line 33: Misc improvement to remove trailing full-stop.

themes/book/template.yml
```diff
@@ -57,10 +57,13 @@ options:
     description: Plausible analytics key
   - type: boolean
     id: numbered_references
-    description: Show references as numbered, rather than in APA-style. Only applies to parenthetical citations.
+    description: Show references as numbered, rather than in APA-style. Only applies to parenthetical citations
```
agoose77, line 60: Misc improvement to remove trailing full-stop.

themes/book/template.yml
```diff
@@ -57,10 +57,13 @@ options:
...
   - type: boolean
     id: folders
-    description: Respect nested folder structure in URL paths.
+    description: Respect nested folder structure in URL paths
+  - type: file
+    id: style
+    description: Local path to a Tailwind CSS file
```
agoose77, line 66: This may/may not end up being changed to `Local path to a CSS file`

themes/book/app/routes/[custom-style.css].tsx
```diff
@@ -0,0 +1,16 @@
+import type { LoaderFunction } from '@remix-run/node';
+import { getCustomStyleSheet } from '~/utils/loaders.server';
+
+export const loader: LoaderFunction = async (): Promise<Response> => {
+  const css = await getCustomStyleSheet();
+  if (!css) {
+    return new Response(`/* No custom stylesheet found */`, {
```
Comment (agoose77) [inline themes/book/app/routes/[custom-style.css].tsx:7]: @rowanc1 what do you think about this? 

The issue is that AFAICT we can't define the `links` function such that it only adds custom CSS if it's defined. So, rather than 404ing for any site that doesn't deploy custom CSS, the "work around" is to define an empty fall-back. Is this OK?

Comment (rowanc1) [reply themes/book/app/routes/[custom-style.css].tsx:7]: Yeah, I think this is good. An extra network request, but not a big deal.

stevejpurves:
ha started looking at this now :D only to realise it's merged @agoose77.

PR #541, opened 2025-02-27
ℹ️ Add note about JupyterHub requirements to launch bar

### LINKED ISSUE #511: Add help tooltips to launch button to communicate Hub/Binder feature requirements
The launch button uses nbgitpuller et al. to provision the environment that it launches into. It is not the case that all Hub / Binder deployments will have the necessary configuration. We should clarify that so that if the launch fails, users have the ability to determine why.

In the long run, 2i2c et al. can do more work in this space to standardise provider mechanisms.

agoose77:
Closes #511 as part of 2i2c's work on https://github.com/2i2c-org/infrastructure/issues/5045

### DIFF
--- a/.changeset/famous-wasps-bow.md
+++ b/.changeset/famous-wasps-bow.md
@@ -1,0 +1,5 @@
+---
+'@myst-theme/frontmatter': patch
+---
+
+Add note about JHub requirements
--- a/packages/frontmatter/src/LaunchButton.tsx
+++ b/packages/frontmatter/src/LaunchButton.tsx
@@ -465,6 +465,24 @@ function DetectLaunchContent(props: ModalLaunchProps) {
         </div>
       </Form.Field>
 
+      <div
+        className={classNames('p-4 rounded-lg text-blue-600 bg-blue-50 dark:bg-slate-900', {
+          hidden: detectedProviderType === 'binderhub',
+        })}
+      >
+        <p className="mb-1">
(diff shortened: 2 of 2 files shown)


agoose77:
@stevejpurves the UX logic for the launch button is becoming complex. There's an interplay between using `<form>` for accessibility and validation, using `useState` in order to handle input events, using `fetch` to asynchronously compute state, etc.

I think we need to use a reducer here, but the `fetch` and `<form>` angle make that a bit harder. Any thoughts?

stevejpurves:
@agoose77 I guess our static deployments and the fact that this is a potentially widely used component is the reason why you're not trying to use the remix Form, fetcher.Form patterns?

So that aside, using a "reducer" would maybe help you manage state here but react's `useReducer` hook is too limited and can't really be used for async fetch. above that I would look at a very isolated / surgical use of zustand to create a local store just for this component but in doing that I would consider pushing the launch button out to it's own mini library, certainly bringing a dep like zustand into `@myst-theme/frontmatter` seems wrong.

agoose77:
Thanks @stevejpurves! In the interest of getting this shipped for now, are you OK to approve this for merging? I'll circle back to this component down the road and pull it out.

agoose77:
This is a fairly benign change, and I don't want to add to the backlog for review. I'll bias for action and merge this for now. Let me know if you'd prefer a different approach next time!

Issue #527, opened 2025-01-27
Support Bokeh Plots


acocac:
I recently start experimenting [JB2-alpha](https://next.jupyterbook.org) for upgrading the [EDS book](https://github.com/alan-turing-institute/environmental-ds-book) resource. Some published notebooks in EDS book use the [Holoviz ecosystem](https://holoviz.org) which builds Bokeh/Matplotlib/Plotly plots.

It seems MyST has some issues to display bokeh-related plots. For instance the notebook, https://eds-book.github.io/b34facfa-cea8-48f5-89f6-f11ce00812a9/notebook#compute-ndvi-and-fdi-indices, using a `bokeh` backend, doesn't show up the plot in the output cell.  

Neither importing `output_notebook` from `bokeh.io import` nor a print statement of the interactive plot works. 

@stevejpurves mentioned a potential short term workaround similar to `plotly` which would be great to point/crossreference in this issue.

acocac:
If it helps to test new fixes, I've created a new repository using simple examples from the `hvplot` library, see https://github.com/eds-book/test-hvplot. The published website confirms MyST can handle [plotly](https://eds-book.github.io/test-hvplot/plotly) but it fails when using the [bokeh](https://eds-book.github.io/test-hvplot/plotly) backend.

acocac:
@stevejpurves - I wondered if there's any progress on supporting bokeh-related plots in MyST. I've tried in the test example some alternatives, preload jquery, export bokeh objects to html and embedded them within the notebook, etc. when I explore/render the repo in nbviewer, it displays the bokeh plots.

stevejpurves:
@acocac #545 should fix your immediate issues for porting a jupyter-book, with further work needed for all myst features in #545

stevejpurves:
#544 was merged, awaiting release

agoose77:
@acocac Jupyter Book 2 for now uses the same themes that MyST-MD uses, which are pulled in by `mystmd` / `jupyter-book` in the `_build/templates/site` directory. You won't need to update JB explicitly to get the latest themes, but sometimes updates are required in the core MyST engine (e.g. for user-CSS and/or landing pages).

You can play around with a local theme by serving your content with MyST, and [running the theme separately](https://github.com/jupyter-book/myst-theme?tab=readme-ov-file#working-with-themes).
[...]

PR #544, opened 2025-02-28
holoviews support: fix output path fetching

stevejpurves:
This fixes a bug in how outputs with multiple minified mime bundles (with path properties) were having their mime bundles properly fetched and rebuilt, resulting in some mimebundles being stripped.

This had a direct effect on holoviews (bokeh, plotly) which this PR fixes (related: #527)

This bug could have also been responsible for other unexpected behaviour where ever the multiple minified mimebundles are present.

### DIFF
--- a/packages/jupyter/src/hooks.ts
+++ b/packages/jupyter/src/hooks.ts
@@ -34,6 +34,7 @@ async function fetcher(url: string) {
         // pass
       }
     }
+    console.log('fetcher', url, content);
     return { content } as any;
   }
   throw new Error(`Content returned with status ${resp.status}.`);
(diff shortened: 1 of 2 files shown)


packages/jupyter/src/hooks.ts
```diff
@@ -34,6 +34,7 @@ async function fetcher(url: string) {
         // pass
       }
     }
+    console.log('fetcher', url, content);
```
rowanc1, line 37: Remove before merge:

```suggestion
```

agoose77:
@stevejpurves just seen the review request -- Will take a look tomorrow, but not worried if someone else reviews and merges first.

agoose77 (approved):
As I understand it, the fix here is that previously we were evaluating
```js
(...links) => links.forEach(fn(link[0]));
```
which ultimately only handled `links[0]`. Now we're evaluating
```js
(links) => links.forEach(fn(link));
```
LGTM!

packages/jupyter/src/hooks.ts
stevejpurves: darn too fast :(

stevejpurves: addressed in #552

PR #546, opened 2025-03-04
🧱 Only apply grid system to outer `<MyST>` component

agoose77:
One solution to jupyter-book/mystmd#1886 is to pull the top-level display logic _out_ of the `Block` component and into the caller.


agoose77:
@rowanc1 @stevejpurves mentions that you're already looking at this! I'm not working on this PRs any more, they were just a PoC for conversation.

rowanc1:
Basically took this approach, simplified the blocks a bit and are no longer passing the class down.

PR #547, opened 2025-03-04
🪜 Use `BlockDepth` provider to scope grid layout to top-level blocks

agoose77:
An alternative PR to #546


agoose77:
@rowanc1 @stevejpurves mentions that you're already looking at this! I'm not working on this PRs any more, they were just a PoC for conversation.

PR #548, opened 2025-03-04
🧱 Simplify Blocks (remove grid-system)

rowanc1:
image

Fixes jupyter-book/mystmd#1886

### DIFF
--- a/packages/myst-to-react/src/block.tsx
+++ b/packages/myst-to-react/src/block.tsx
@@ -2,50 +2,32 @@ import { Details } from './dropdown.js';
 import { MyST } from './MyST.js';
 import type { GenericParent } from 'myst-common';
 import classNames from 'classnames';
-import { useGridSystemProvider } from '@myst-theme/providers';
 import type { NodeRenderer } from '@myst-theme/providers';
 
-export function Block({
-  id,
-  node,
-  className,
-}: {
-  id: string;
-  node: GenericParent;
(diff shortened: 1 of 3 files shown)


packages/myst-to-react/src/block.tsx
```diff
@@ -2,50 +2,32 @@ import { Details } from './dropdown.js';
...
+  const cn = classNames(className, node.class, {
+    [node.data?.class]: typeof node.data?.class === 'string',
+  });
+  if (node.visibility === 'remove') return null;
+  // Only wrap this in a block if the block has a class name or an identifier
+  // Otherwise pass through the contents as is.
+  // This allows children (e.g. margin) access to the grid-system
+  const block =
+    cn || node.identifier ? (
+      <div id={node.identifier} className={cn}>
```
agoose77, line 17: @rowanc1 I don't think this change is friendly to future outputs work or #540:

Elements like `aside` need the CSS grid to position themselves correctly. This is only possible if they are top-level elements within a grid, or their parents use `subgrid`. Previously, we ensured that the outermost block created a subgrid.

This PR drops the subgrid feature by only wrapping the block children in a node if they have a class or an identifier. Both code-outputs that generate e.g. an `aside` and blocks that use #540 will set an identifier and/or class, but want to preserve the grid system.

I think instead we probably need to retain the grid e.g. #547.

PR #556, opened 2025-03-06
🌲 Add `root` renderer

agoose77:
By imposing the constraint that `root` cannot have a `class` or `label`, we can eliminate the need to render a `div` for this element.

### DIFF
--- a/packages/myst-to-react/src/MyST.tsx
+++ b/packages/myst-to-react/src/MyST.tsx
@@ -28,11 +28,12 @@ export function MyST({
   className?: string;
 }) {
   const renderers = useNodeRenderers();
-  if (!ast || ast.length === 0) return null;
+  if (!ast) return null;
   if (!Array.isArray(ast)) {
     const Component = selectRenderer(renderers, ast);
     return <Component key={ast.key} node={ast} className={className} />;
   }
+  if (ast.length === 0) return null;
   return (
     <>
       {ast?.map((node) => {
--- a/packages/myst-to-react/src/basic.tsx
+++ b/packages/myst-to-react/src/basic.tsx
@@ -54,6 +54,10 @@ type Glossary = {
   type: 'glossary';
 };
 
+type Root = {
+  type: 'root';
+};
+
 type BasicNodeRenderers = {
   text: NodeRenderer<spec.Text>;
   span: NodeRenderer<GenericNode>;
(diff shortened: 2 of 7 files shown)


packages/myst-to-react/src/basic.tsx
```diff
@@ -412,6 +417,15 @@ const BASIC_RENDERERS: BasicNodeRenderers = {
       </div>
     );
   },
+  root({ node, className }) {
+    if (node.children === undefined) {
+      return <></>;
+    }
+    const childComponents = node.children.map((child) => (
+      <MyST ast={child} className={className} />
```
rowanc1, line 425: Any map will need a key. I think we can just pass the `node.children` to MyST, and that does the same as this. Will make an update.

packages/myst-to-react/src/MyST.tsx
```diff
@@ -28,11 +28,12 @@ export function MyST({
   className?: string;
 }) {
   const renderers = useNodeRenderers();
-  if (!ast || ast.length === 0) return null;
+  if (!ast) return null;
   if (!Array.isArray(ast)) {
     const Component = selectRenderer(renderers, ast);
     return <Component key={ast.key} node={ast} className={className} />;
   }
+  if (ast.length === 0) return null;
```
rowanc1, line 36: Minor updates to this logic, technically a node with `{length: 0}` as a property would never render. This ensures first that we are dealing with an array.

themes/article/app/components/Article.tsx
```diff
@@ -93,7 +93,7 @@ export function Article({
           <ErrorTray pageSlug={article.slug} />
           <div id="skip-to-article" />
           <FrontmatterParts parts={parts} keywords={keywords} hideKeywords={hideKeywords} />
-          <MyST ast={tree.children as GenericParent[]} />
+          <MyST ast={tree} />
```
rowanc1, line 96: Confirmed that this now renderers without a wrapping div.

packages/myst-to-react/src/basic.tsx
agoose77: I've been reading a fair amount of React patterns lately, and still forgot `key`. Also, this is an obvious case for passing children directly ... not sure how I missed it.

rowanc1: 🚀

PR #560, opened 2025-03-11
🖥️ Add deploy workflow to CI

agoose77:
This PR adds a deployment workflow that uses a personal fine-grained GitHub token to push to `myst-templates`. It will expire in 365 days. I've already tested this with https://github.com/myst-templates/book-theme/commit/c4a0e5d80187e543ef3363ae725b2007aa9b3683

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -21,6 +21,9 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version: 16.x
+      - name: Install dependencies
+        # TODO: this only needs to install thebe install deps
+        run: npm install
       - name: Configure Git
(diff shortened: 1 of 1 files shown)


agoose77:
I'm merging this because it works!

There are other ways to this kind of machinery. We could use webhooks to trigger a build in `myst-templates`. This would be slightly more secure than using the current fine-grained token to push to the repo.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,35 @@
...
+
+      - name: Setup Node.js 16.x
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16.x
+      - name: Install Dependencies
+        # TODO: this only needs to install thebe install deps
+        run: npm install
+      - name: Build MyST Theme
+        run: npm run build
```
agoose77, line 28: This is annoying. We should be able to re-use the deployed artifacts. However, I don't *think* there's a trivial way to force `npm` to use the latest artifacts, and this approach avoids the possibility of a race condition.

PR #565, opened 2025-03-18
👣 Add site footer

### LINKED ISSUE #579: Grid does not write out class name
Input:

```
:::{grid} 2 2 4 4
:class: footer

...
:::
```

Does not write class `footer` to the resulting `<div>`.

stefanv:
Example content: https://github.com/stefanv/landing-pages/tree/footer
[Footer specification](https://raw.githubusercontent.com/stefanv/landing-pages/refs/heads/footer/footer.md)

Launch content server with:

```sh
myst start --template ../myst-theme/themes/book/template.yml
#                       ^-- fix path to your myst-theme location
```

This is a first attempt at adding a site footer. It may need some adjusting to the CSS to make it render better.

I've looked into adding social media icons, but I'm having trouble getting the [scienceicons](https://github.com/continuous-foundation/scienceicons/tree/main/myst#scienceiconsmyst) plugin to work.

image

Works towards https://github.com/jupyter-book/mystmd/issues/1920

Closes #579


agoose77:
This is very exciting @stefanv!

@fwkoch this reminds me — am I right in thinking that a common MyST pattern is to either define the parts explicitly as a list of files in the frontmatter, or inline in the document, where the document takes precedence? Can we apply the same pattern here @stevejpurves?

agoose77:
@stefanv these questions are _not_ blockers for you here -- I'm just touching base with the rest of the team more broadly!

themes/book/app/components/SiteFooter.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { filter } from 'unist-util-filter';
+
+
+export function SiteFooter({ tight, content }: { tight?: boolean }) {
+  const footerBlock = selectAll('block', content).find((node) => node.kind == 'footer');
+  const footerColumnsBlock = selectAll('block', content).find((node) => node.kind == 'footer-columns');
+  const footerLists = selectAll('list', footerColumnsBlock);
+
+  return (
+    <section className={classNames('col-screen article-grid', { 'mt-10': !tight, 'mt-0': tight })}>
```
kne42, line 16: nit: use conditional, add horizontal padding

```suggestion
    <section className={classNames('col-screen article-grid px-4', tight ? 'mt-0' : 'mt-10')}>
```

kne42:
Hi, I've been looking at various site footers to create a mood board:

Slide 16_9 - 1

I have some observations:
#### Often Has
- logo/link to main page
- columns of site links
- social media icons

#### Sometimes Has
- links formatted as site index (typically in documentation)
- legal links
- copyright

#### Can Have
- CTA/email signup

Overall I like the format suggested by this PR since it seems to encompass all of the basics that most sites use. CTAs are something a user who wants to get more in the weeds would use so they shouldn't be handled in the default template; also they would likely require some kind of additional backend anyways.

kne42:
Since we were talking about layout I did a little mockup of what layouts the different moodboard examples used:

Slide 16_9 - 1 (1)

Honestly, it's all over the place but here are some patterns I noticed:
- general site links are side-by-side with social media icons
- logo is the first element regardless of if the top layout is row or col (or grid)
- legal links tend to be separated from the general site links and are at the bottom of the page in another row

kne42:
mockup of scientific python site

Screenshot 2025-05-12 at 12 07 01 PM

agoose77:
@kne42 and @fwkoch and I were talking about how to design this in a generic way whilst also retaining the aesthetics that @stefanv has been building.

We suggested that the footer should be a two-column flex layout: right hand side is social icons and copyright, disappearing if neither are defined. Left hand side comes from a footer part. We *might* want to also define a `footer` block or expose `align-items` as a class that can be added to a MyST grid to preserve the vertical alignment.

Social icons will be a subsequent PR, so for now we'll only add the copyright.

choldgraf:
@agoose77 so mapping what you're describing onto @kne42 's mock-ups, it sounds like you'd accomplish the scientific python setup by having a 4-column grid for the left side, with (1) logo image and (2/3/4) being three-link columns. And the right side would be a bunch of image links. That seems reasonable to me!

kne42:
Something else to consider for websites is responsive design. You would typically want to collapse the grid into a column to preserve horizontal real estate.

Perhaps with this in mind, for default use cases, we may want something even more opinionated.

kne42:
> mockup of scientific python site
> 
> Screenshot 2025-05-12 at 12 07 01 PM

this is how I would structure it in code:
```html
<div class="lg:grid lg:grid-cols-3 flex flex-col">
  <!-- Logo -->
  <ScientificPythonLogo />
  <!-- Site Links -->
  <ul class="lg:grid lg:grid-cols-3 flex flex-col">
    <li>About</li>
    ...
    <li>Press kit</li>
  </ul>
  <!-- Social Media Icons -->
  <div class="lg:flex lg:flex-row grid grid-cols-3">
    <GitHubIcon />
    ...
    <DiscordIcon />
  </div>
  <!-- Copyright -->
  <p class="lg:col-start-3">copyright stuff blah blah blah</p>
</div>
```

kne42:
Given that responsive design would complicate things I'd like to suggest a new opinionated `footer` directive:

````
```{footer}
:logo: <valid markdown>
:[discord|mastodon|youtube|bluesky|instagram|etc]: <url>
:extra: <valid markdown>
:cols: <number, default 3> number of columns for your links
Directive Body: the links that will be formatted
```
````

e.g.
````
```{footer}
:logo: [![Scientific Python Logo](/images/logo.svg)](/)

:github: https://github.com/scientific-python/
:youtube: https://www.youtube.com/c/ScientificPython-org
:mastodon: https://fosstodon.org/@scientific_python
:bluesky: https://bsky.app/profile/scientific-python.org
:discuss: https://discuss.scientific-python.org/
:discord: https://discord.com/invite/vur45CbwMz

:extra: © 2025 Scientific Python team. All rights reserved.

- [About](/about)
- [Roadmap](/roadmap)
- [Code of Conduct](/code-of-conduct)
- [SPECs](/specs)
- [Summits](/summits)
- [Calendars](/calendars)
- [Press kit](/press-kit)
```
````

**Comment by stefanv** [comment]
@kne42 How would you split the columns out of the list? E.g., 3 in the first col, 2 in second, 5 in third.

To myst devs: how does the dedicated directive approach compare against the existing blocks?

Alternative that was rejected earlier: configuring footer data in config / myst.yml.

**Comment by kne42** [comment]
> With the directive, how would you split the columns out of the list? E.g., 3 in the first col, 2 in second, 5 in third.

You could probably group them with div directives if I understand how the syntax works correctly:
`````
````{footer}
...

```{div}
- 1
- 2
- 3
```

```{div}
- 4
- 5
```

```{div}
- 6
- 7
- 8
- 9
- 10
```
````
`````

**Comment by agoose77** [comment]
Following up from the in-person conversations:

@kne42's mockups have been really useful in synthesising some conversation about the core features that we need. In conversation, we've been continuing to explore the tension between wanting some control over layout without having verbose markup and/or requiring custom stylesheets out of the box.

In a conversation with @stefanv, I outlined the way in which directives and block annotations are not mutually exclusive: one could define a directive that generates some blocks. Often, the UX of using blocks is frustrating because they're top-level-only syntax elements (even if we've been relaxing the ability to nest them at the AST level).

@kne42 you're correct on the `div` usage, and I learned about `col-start` shorthand for tailwind! Nice :rocket: 

After touching base with @stefanv, we iterated on the above design to suggest that the MyST theme has an opinionated three-column layout as shown in https://github.com/jupyter-book/myst-theme/pull/565#issuecomment-2873703610. Each column can be populated by a named `part`, and the right-most column has a default that is populated by social-icon and copyright fields that are draw from frontmatter.

Where any column part is _not_ defined (and doesn't have a default), it isn't rendered (making the other columns "wider"). This means that the Scientific Python footer is trivial in markup:

````

+++ { "part": "footer-left" }
<h4>Scientific Python</h4>

```{image} https://scientific-python.org/images/logo.svg
:width: 50px
:align: left
```

This is a description of our project. And a [link to its homepage](https://scientific-python.org).

+++ { "part": "footer-center" }


:::{grid} 1 2 3 3
[About](https://scientific-python.org/about/)  
[Scientific Python blog](https://blog.scientific-python.org)  
[Tools](https://tools.scientific-python.org)

A second column!  
With multiple entries

And what about a third
:::
:::::

````

Where the `footer-right` block is not user-defined, so it falls back to the social icons and copyright. We apply a vertical center on each column.

**Comment by choldgraf** [comment]
I think that this direction sounds great. I feel like we have surfaced a lot of great ideas, and this feels like a really nice direction that is worth trying. +1 from me

**Comment by rowanc1** [comment]
My strong preference would be for a **single** "footer" part, which allows you to, for example, stack footers add a "banner" or branded footer element.

The part should be where the footer lives (the bottom, full width), then the directives should have the content inside of there. This allows us to have multiple stacked footers (e.g. different brands, or flexible ways of representing funding, see below).

With the proposed direction, we are constraining our selves to certain kinds of footers (ones that have left/center/right dividers, but no double brand footers, no funders section on every page, etc.).

This could be done in a single part, with maybe a `{footer}` and maybe a `{footer .simple}` directive in a single footer part. You could see a similar way of putting a copyright statement on every page `{div .center}`!, or your own plugins for footers in the future. If the part is "footer-center" we are much more constrained in the future.

My proposal is more similar to [this direction](https://github.com/jupyter-book/myst-theme/pull/565#issuecomment-2874322911) by @kne42. Have a footer directive, that includes the logo/links/social and deals with the layout in the directive. (could go with a list of lists for the links? Or a list-table? or just a div/grid that has a special class on it?)

This is an example of "two footers":
image

Here is an example of three stacked footers on every page:
image

**Comment by agoose77** [comment]
Thanks for your two-cents @rowanc1!

There are a couple/few orthogonal concerns here:
1. It's my feeling that social links should be defined in the frontmatter. They're ultimately a description of where a community can be reached. To me, it's essentially that a _project_ can have a social identity in the same way [that authors can](https://mystmd.org/guide/frontmatter#social-links). 
2. The MyST theme _can_ be opinionated. We already are opinionated about where action links are defined, and the presence of header icons. I'm somewhat of the view that social icons are part of this "theme-derived style". I'm keen for us to consciously figure out where the sensible midpoint is between giving people constraints vs total customisation. My preference here is to do out-of-the-box simply and well. I'm a bit reluctant to have too much content-defined layout vs having the theme make some choices. 
3. I would like us to recognise patterns and provide out-of-the-box ways to implement those, rather than encouraging people to think and use MyST as a feature-restricted web authoring canvas.

A technical constraint that is feeding into this is the mechanism for authoring footers. Some unordered thoughts:
- Landing pages use blocks, because they're intentionally top-level and non-nestable
- Block syntax is not nestable
- Blocks feel similar to divs
- Parts can be defined naturally _inside_ a document using blocks. This feels very natural, but prevents a part using blocks itself
- Perhaps we should have a block directive to break this (because it brings along the idea of a terminator `:::` or `` ``` ``.

**Comment by agoose77** [comment]
For Rowan's benefit (other stakeholders here are at a sync in-person dev summit):

@agoose77 thinks that some of the work here is _archetypal_, meaning that it represents a decision that will inform our approach / idioms / best practices in other places. Being aligned on that is very important.

But, finding alignment on this is hard work, and @agoose77 is particularly worried about stalling out _this_ PR whilst we work in that direction. As such, we're in the process of building an MVP next-step that takes into account most of the comments here _and crucially_ kicks the can down the road of whether MyST Themes are opinionated by _arbitrarily deciding that they are not, for now_.

**Comment by kne42** [comment]
As we're talking in-person this seems to be the pattern we're considering right now:
- Social media links are defined in the frontmatter
- The part can have anything in it
- We make a footer directive that can then be combined with other element in the part

With this, for the Scientific Python project, it might look something like this:
#### myst.yml
```yaml
project:
  github: https://github.com/scientific-python/
  youtube: https://www.youtube.com/c/ScientificPython-org
  mastodon: https://fosstodon.org/@scientific_python
  bluesky: https://bsky.app/profile/scientific-python.org
  discuss: https://discuss.scientific-python.org/
  discord: https://discord.com/invite/vur45CbwMz
```

#### footer part
```
::::{footer}
:logo: [![Scientific Python Logo](/images/logo.svg)](/)
:socials: github | youtube | mastodon | bluesky | discuss | discord

:::{div}
- [About](/about)
- [Roadmap](/roadmap)
- [Code of Conduct](/code-of-conduct)
:::
:::{div}
- [SPECs](/specs)
- [Summits](/summits)
- [Calendars](/calendars)
:::
:::{div}
- [Press kit](/press-kit)
:::
::::

:::{div .right}
© 2025 Scientific Python team. All rights reserved.
:::
```

rowanc1:
Thanks for the updates. @kne42 's latest proposal looks awesome from my perspective. Allows for multiple footers, other footer plugins, and more extensibility, while also being constrained and solid out of the box.

themes/book/template.yml
```diff
@@ -64,6 +64,9 @@ options:
   - type: file
     id: style
     description: Local path to a CSS file
+parts:
+  - id: footer
+    description: The site wide footer
```
agoose77, line 69:
```suggestion
    description: The site wide footer
    required: false
```
```suggestion
    description: The site wide footer
```

themes/book/app/components/ArticlePage.tsx
```diff
@@ -71,6 +72,7 @@ export const ArticlePage = React.memo(function ({
     ...siteDesign,
     ...pageDesign,
   };
+  const projectParts = useSiteManifest()?.parts;
```
agoose77, line 75: TODO: types
```suggestion
  const projectParts = useSiteManifest()?.parts;
  const pageParts = (article.frontmatter as any)?.site?.parts ?? {};
  const parts = {...projectParts, ...pageParts}
```

themes/book/app/components/ArticlePage.tsx
```diff
@@ -124,6 +126,7 @@ export const ArticlePage = React.memo(function ({
           {!hide_footer_links && !hide_all_footer_links && (
             <FooterLinksBlock links={article.footer} />
           )}
+          {projectParts?.footer && <SiteFooter content={projectParts.footer.mdast} />}
```
agoose77, line 129:
```suggestion
          {parts.footer && <SiteFooter content={parts.footer.mdast} />}
```

themes/book/app/components/SiteFooter.tsx
```diff
@@ -0,0 +1,39 @@
+import type { GenericNode, GenericParent } from 'myst-common';
```
agoose77, line 1: TODO: remove this file in this PR

stefanv:
After further discussion, we decided to break this up into two parts:

1. Provide a footer section via parts, coming in either via the project or page configuration.
2. A `footer` directive, that the user can use inside that part. However, they can also choose to style and format that part however they please.

Multiple footers can be provided by using blocks or divs inside the footer part, to which the user would attach a custom class.

stefanv:
Started along the proposed path.

Footer is now a generic block. The footer content that renders this:

image

Looks like:

```md
<h1></h1>

:::::{div}
:class: footer col-screen

::::{grid} 2 2 4 4

:::{div}
<h4>Scientific Python</h4>

```{image} https://scientific-python.org/images/logo.svg
:width: 50px
:align: left
```

This is a description of our project. And a [link to its homepage](https://scientific-python.org).
:::

:::{div}
- [About](https://scientific-python.org/about/)
- [Scientific Python blog](https://blog.scientific-python.org)
- [Tools](https://tools.scientific-python.org)
:::

:::{div}
- A second column!
- With multiple entries
:::

:::{div}
- And what about a third
:::

::::

:::::
```

I ran into two issues along the way: https://github.com/jupyter-book/mystmd/issues/2022 and https://github.com/jupyter-book/myst-theme/issues/579

Aesthetics-wise, I would still like to get the columns to top-align (not with the logo block, but only among the link lists.)

**Comment by choldgraf** [comment]
A few quick thoughts:

- The output looks really nice to me! I agree it'd be better aesthetics if the menu options upper-aligned with the text on the left. That said it doesn't feel like a dealbreaker.
- Why did you use `<h4>` style tags rather than `#### markdown` headers. Is that intentional?
- Why is there an empty `<h1></h1>` tag in there? (I'm wondering if that's an essential part of this footer or just copypasta)

Either way, none of that is worth blocking progress IMO. I am looking forward to be able to try this out on `main`!

**Comment by agoose77** [comment]
This PR is pretty close, I think.

After talking with Rowan RE #582, we've settled on continuing only to look at the `site` parts _for now_, meaning that we don't worry about per-page footers. I *think* that a footer should probably default to a page layout, but users should be able to define a full-width footer. I need to remind myself of the nature of nested grids to achieve this. I suspect that it will involve a similar solution to our original `Block` component that tested for `col-` classnames. 

Left to do in this PR:
- Finish the basic styling of the `Footer` component
- Document the new feature

I'm intending to look at (1) on Monday!

**Comment by agoose77** [comment]
I did some work on getting this to the finish line. In the last few commits, I pushed:

- Ability to pass in a `className` to this component. It's not used for now.
- Change default column for the footer layout grid
- Single-component footer, rather than nested `footer > div`

The selector for the default column has the same precedence as a class, so it's easy to override from a custom class & stylesheet.

**Comment by agoose77** [comment]
@stefanv can you confirm that this feels usable when combined with https://github.com/stefanv/landing-pages/pull/1?

File: packages/myst-to-react/src/grid.tsx [hunk 6]
```diff
@@ -127,7 +127,7 @@ function Grid({
 
 export const GridRenderer: NodeRenderer<GridSpec> = ({ node, className }) => {
   return (
-    <Grid columns={node.columns} className={className}>
```
agoose77, line 130: Fixes #579

themes/book/app/routes/$.tsx
```diff
@@ -150,24 +154,24 @@ export default function Page() {
...
-      <ProjectProvider>
-        <ComputeOptionsProvider
-          features={{ notebookCompute: true, figureCompute: true, launchBinder: false }}
-        >
-          <ThebeLoaderAndServer baseurl={baseurl}>
+    <ProjectProvider>
+      <ComputeOptionsProvider
+        features={{ notebookCompute: true, figureCompute: true, launchBinder: false }}
+      >
+        <ThebeLoaderAndServer baseurl={baseurl}>
```
agoose77, line 161: Lifted these up so that the footer behaves similarly to an article page w.r.t computation.

themes/book/app/routes/$.tsx
```diff
@@ -150,24 +154,24 @@ export default function Page() {
...
+          <ArticlePageAndNavigation
+            hide_toc={hide_toc}
+            hideSearch={hide_search}
+            projectSlug={data.page.project}
+          >
+            {/* <ProjectProvider project={project}> */}
```
agoose77, line 167:
```suggestion
    {/* <ProjectProvider project={project}> */}
    <ProjectProvider>
      <ComputeOptionsProvider
        features={{ notebookCompute: true, figureCompute: true, launchBinder: false }}
      >
        <ThebeLoaderAndServer baseurl={baseurl}>
          <ArticlePageAndNavigation
            hide_toc={hide_toc}
            hideSearch={hide_search}
            projectSlug={data.page.project}
          >
```

agoose77:
This PR is nearly done. Right now, executing code-cells in a footer probably (certainly) doesn't work. I'll look to tackle that tomorrow.

choldgraf (approved):
This looks good to me in terms of the UX. That said, I don't think that it closes https://github.com/jupyter-book/mystmd/issues/1920 because this doesn't document any of this functionality. Before we consider this done, I suggest we open a PR to mystmd that documents the "what, why, and how" of this feature.

In my opinion, executing code cells in the footer isn't necessary for this to get in. IMO if there's finite time, I think it'd be more useful to document this feature than to make execution work. We can always say "cell execution doesn't work in the footer" and add that in a future PR if people request it enough.

agoose77:
@choldgraf I'm quite keen to unblock this work - my intention is to treat the main issue as a user-story, and capture this PR and docs as sub issues.

choldgraf:
I should clarify - the reason I said this is because I saw that `Closes https://github.com/jupyter-book/mystmd/issues/1920` was in the PR body, which means when it is merged then the parent issue will auto-close :-)

stefanv:
Thanks for all the discussions and input, everyone, and @agoose77 for taking it over the finishing line. I'll take a stab at documenting it.

choldgraf:
I can help with review and edit cycles on the docs @stefanv , please don't hesitate to ping me in discord if you'd like a quick review

PR #570, opened 2025-03-25
🎓 Add support for `button` in `cite` nodes

agoose77:
This PR follows up #521 to support buttons with citations. In the long run, I think we might want to replace these selectors with something more React driven, but this is a first pass.

### DIFF
--- a/packages/myst-to-react/src/cite.tsx
+++ b/packages/myst-to-react/src/cite.tsx
@@ -67,11 +67,17 @@ export const Cite = ({
     return <InlineError value={label} message={'Citation Not Found'} className={className} />;
   }
   const url = doiString ? doi.buildUrl(doiString as string) : refUrl;
+  const isButtonLike = (className ?? '').split(' ').includes('button');
   return (
     <HoverPopover openDelay={300} card={<CiteChild html={html} />}>
       <cite className={className}>
         {url && (
-          <a href={url} target="_blank" rel="noreferrer" className="hover-link">
+          <a
+            href={url}
+            target="_blank"
+            rel="noreferrer"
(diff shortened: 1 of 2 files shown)


stevejpurves (commented):
Would it be better to instead decorate the cite node with a no-op class name and allow the user to target that in their user defined css?

When I first read the PR title I thought you might have been swapping the anchor tag out with a button and providing full button semantics rather than link semantics, hence the need to modify the node renderer but 

(1) It would be nice to minimise node renderer changes in general, especially for styling issues and this is a styling issue (2) the click action on citations is still a "navigation" action so an anchor is appropriate whilst a button is not, styling as a button here imo is poor web UX, so another reason to be left to the user rather than baked into the library.

I'd suggest we go back to the root requirement driving this change and instead use the no-op class + user styles to solve it.

agoose77:
@stevejpurves the context for this change is that we've deployed button-like link support in MyST (originally motivated by the work surrounding #474), but it was missed for citations that are built from links (e.g. DOIs).

Regarding 
> styling as a button here imo is poor web UX, so another reason to be left to the user rather than baked into the library.

I disagree. This is a well established pattern, and crucially we are not changing the a11y semantics. See https://tailwindcss.com/plus/ui-blocks/marketing/sections/heroes for examples of CTA button links.

agoose77:
@stevejpurves I'd like to get this merged. Do you have a second to read my last and yes/no it?

stevejpurves:
All I'd say is an inline citation is not the same as a CTA button but this is just a class, so fine.

stevejpurves (approved):

PR #571, opened 2025-04-03
☢️ DO NOT MERGE ☢️ Prepare for `outputs` AST change

agoose77:
This PR is a WIP to anticipate the new AST in jupyter-book/mystmd#1903.

### DIFF
--- a/packages/jupyter/src/embed.tsx
+++ b/packages/jupyter/src/embed.tsx
@@ -8,7 +8,7 @@ export function Embed({ node }: { node: GenericNode }) {
   return (
     <OutputDecoration
       key={node.key}
-      outputId={output.id}
+      outputId={output.id ?? output.key}
       title={node.source?.title}
       url={node.source?.url}
       remoteBaseUrl={node.source?.remoteBaseUrl}
--- a/packages/jupyter/src/output.tsx
+++ b/packages/jupyter/src/output.tsx
@@ -4,12 +4,12 @@ import type { MinifiedMimeOutput, MinifiedOutput } from 'nbtx';
 import classNames from 'classnames';
 import { SafeOutputs } from './safe.js';
 import { JupyterOutputs } from './jupyter.js';
-import { useMemo, createContext, useContext } from 'react';
+import { useMemo } from 'react';
 import { useCellExecution } from './execute/index.js';
-import type { IdOrKey } from './execute/types.js';
 import { usePlaceholder } from './decoration.js';
 import { Details, MyST } from 'myst-to-react';
 import { selectAll } from 'unist-util-select';
+import { useOutputsContext, OutputsContextProvider } from './providers.js';
 
 export const DIRECT_OUTPUT_TYPES = new Set(['stream', 'error']);
 
@@ -39,18 +39,8 @@ export function allOutputsAreSafe(
   }, true);
 }
 
-type OutputsContextType = {
-  allSafe: boolean;
-  outputsId: IdOrKey;
-};
-const OutputsContext = createContext<OutputsContextType | null>(null);
-
 export function Output({ node }: { node: GenericNode }) {
-  const context = useContext(OutputsContext);
-  if (!context) {
-    return <> </>;
-  }
-  const { outputsId, allSafe } = context;
+  const { outputsId, allSafe } = useOutputsContext();
   const { ready } = useCellExecution(outputsId);
   const outputs = [node.jupyter_data];
   return allSafe && !ready ? (
@@ -96,9 +86,9 @@ export function Outputs({ node }: { node: GenericNode }) {
         className,
       )}
     >
-      <OutputsContext.Provider value={{ outputsId, allSafe }}>
+      <OutputsContextProvider outputsId={outputsId} allSafe={allSafe}>
         <MyST ast={children} />
-      </OutputsContext.Provider>
+      </OutputsContextProvider>
     </div>
   );
 
--- a/packages/jupyter/src/output.tsx
+++ b/packages/jupyter/src/output.tsx
@@ -61,9 +61,7 @@ export function JupyterOutput({
   let component;
   if (allSafe && !ready) {
     if (placeholder && (!outputs || outputs.length === 0)) {
-      if (placeholder) {
-        return <MyST ast={placeholder} />;
-      }
+      return <MyST ast={placeholder} />;
     }
     component = <SafeOutputs keyStub={outputId} outputs={outputs} />;
   } else {
--- a/packages/jupyter/src/output.tsx
+++ b/packages/jupyter/src/output.tsx
@@ -8,6 +8,7 @@ import { useMemo } from 'react';
 import { useCellExecution } from './execute/index.js';
 import { usePlaceholder } from './decoration.js';
 import { Details, MyST } from 'myst-to-react';
+import { selectAll } from 'unist-util-select';
 
 export const DIRECT_OUTPUT_TYPES = new Set(['stream', 'error']);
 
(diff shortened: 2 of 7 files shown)


packages/jupyter/src/embed.tsx
```diff
@@ -8,7 +8,7 @@ export function Embed({ node }: { node: GenericNode }) {
   return (
     <OutputDecoration
       key={node.key}
-      outputId={output.id}
+      outputId={output.id ?? output.key}
```
agoose77, line 11: This feels like the right move, do you agree? 

I made this change because the `myst-execute` package, used for `myst build --execute`, does not presently set the `id`, and I wondered if it's not redundant with `key`?

packages/jupyter/src/output.tsx
```diff
@@ -37,43 +39,42 @@ export function allOutputsAreSafe(
...
-  className?: string;
-}) {
-  const { ready } = useCellExecution(outputId);
-  const outputs: MinifiedOutput[] = data;
+export function Output({ node }: { node: GenericNode }) {
+  const { outputsId, allSafe } = useOutputsContext();
+  const { ready } = useCellExecution(outputsId);
+  const outputs = [node.jupyter_data];
+  return allSafe && !ready ? (
+    <SafeOutputs keyStub={outputsId} outputs={outputs} />
```
agoose77, line 47: We only changed the interface above `SafeOutputs` and `JupyterOutputs` for now.

packages/jupyter/src/output.tsx
```diff
@@ -37,43 +39,42 @@ export function allOutputsAreSafe(
...
+  const { children, identifier, align } = node;
+  const outputsId = node.id ?? node.key;
+  const cellExecutionContext = useCellExecution(outputsId);
+  const { ready } = cellExecutionContext;
+
+  const outputs: MinifiedOutput[] = useMemo(
+    () => selectAll('output', node).map((child) => (child as any).jupyter_data),
+    [children],
+  );
   const allSafe = useMemo(
```
agoose77, line 64: This will likely change as this PR evolves. For now, I'm preserving as much original behaviour as possible in order to make the PR small and easy to reason about.

packages/jupyter/src/providers.tsx
```diff
@@ -107,3 +108,30 @@ export function ThebeLoaderAndServer({
...
+  children,
+}: {
+  children: React.ReactNode;
+  allSafe: boolean;
+  outputsId: IdOrKey;
+}) {
+  return (
+    <OutputsContext.Provider value={{ allSafe, outputsId }}>{children}</OutputsContext.Provider>
+  );
+}
```
agoose77, line 137: This just allows us to use the `<MyST />` visitor by lifting parent-child state into a context object.

packages/jupyter/src/output.tsx
```diff
@@ -37,43 +39,42 @@ export function allOutputsAreSafe(
   }, true);
 }
 
-export function JupyterOutput({
```
agoose77, line 40: Is it OK to just get rid of this?

agoose77:
@stevejpurves I've done a first pass at `outputs` support. This PR isn't finished, but your glance would be handy at some point.

packages/jupyter/src/providers.tsx
```diff
@@ -107,3 +108,30 @@ export function ThebeLoaderAndServer({
...
+  allSafe: boolean;
+  outputsId: IdOrKey;
+};
+const OutputsContext = React.createContext<OutputsContextType | null>(null);
+
+export function useOutputsContext() {
+  const context = useContext(OutputsContext);
+  if (context === null) {
+    throw new Error('useOutputsContext must be used within a OutputsContextProvider');
+  }
```
agoose77, line 122: Hmm, what should we do if a user embeds a particular numbered output? That wouldn't pull in the outer `outputs` node.

packages/jupyter/src/embed.tsx
stevejpurves: id's are critical but as long as they are unique and stable per-build then it should work - so falling back to key should be ok i think

packages/jupyter/src/providers.tsx
stevejpurves: that is beyond the current feature set correct? i.e. currently someone can only embed the whole set of outputs?

We can think forward to that though...

Previously the `outputId` referred to the whole `output` node, now we have 

`outputs.children[output, output, ...]` referred to with `outputsId` and `outputId`s respectively.

The `useOutputContent` could allow for using an outputsID od any of its child outputId's to reference the same parent outputs node?

packages/jupyter/src/providers.tsx
stevejpurves: yes... this felt a little "Why?!?" to begin with as the context is created in an Outputs renderer and only serves to provide exactly one `outputsId` relating to its parent.

but we're only really using this here to bridge across the interface of the `<MyST />` component, to keep the Output renderer independent.

So (just as discussion here) *if* all renderers received access to the full AST in addition to the node that they were called to render then we potentially would not need this `Context` as an `Output` could look up it's parent, correct? (although we might want it to avoid a lookup) @agoose77 do you know if there has been any discussion on that angle elsewhere?

stevejpurves: yes... this felt a little "Why?!?" to begin with as the context is created in an Outputs renderer and only serves to provide exactly one `outputsId` relating to its parent.

but we're only really using this here to bridge across the interface of the `<MyST />` component, to keep the Output renderer independent.

So (just as discussion here) *if* all renderers received access to the full AST in addition to the node that they were called to render then we potentially would not need this `Context` as an `Output` could look up it's parent, correct? (although we might want it to avoid a lookup) @agoose77 do you know if there has been any discussion on that angle elsewhere?

agoose77: @stevejpurves yes that's correct. We could drop the `outputID` by making that part of the AST.

packages/jupyter/src/embed.tsx
stevejpurves: ok so back tracking on this one.

`key` is not a unique identifier across the site so there is no way to match the outputs/output in the AST fragment within the embed with the same fragment in the source notebook, as the keys for reach fragment are different. The `id` in the output is / was stable and could be used to correlate that across pages.

This is not a thing that is immediately visible in the current theme as each page is independent, but for example should we begin to synchronise in-browser computation across pages (e.g. figures/embeds and their source notebooks) then we need the `id` - this was the intention in order to improve the browsing experience when people are using in-browser compute.

@agoose77 I think we should add the `id` into the `myst-execute` for that reason. Also we should augment our test cases with both pre-computed `ipynb` and `.md` examples

stevejpurves (changes requested):
Discussion points:

* [ ] introducing outputs.id and handling where missing
* [ ] plan for handling embeds/figures to existing xref content
* [x] changing allSafe at outputs level for isSafe at output level

stevejpurves:
Additional testing considerations:

* external xref in new format
* multiple outputs froma cell with
  * all safe
  * no safe
  * some safe

stevejpurves:
it is a slog to set up for testing but embedding external content for the new otuput format is confirmed as working.
image

## test setup

1. get test source content in `folderA`
  * myst start --execute --headless
  * content server running on `localhost:3100`
2. start myst-theme e.g. `npm run theme:book`
  * theme running on `localhost:3000`
3. get second set of test content in `folderB`
  * add `local: http://localhost:3000` to `myst.yml#project.references`
  * add external cross references and embeds to the local content in `folderA`
  * `myst start --execute --headless`
  * content server running on `localhost:3101`
4. in a second `myst-theme` folder, or without fully restarting the theme from step 2
  * add a `.env` file to `themes/book`
  * add `CONTENT_CDN_PORT=3101`
  * `npm run theme:book`
  * theme running on `localhost:3001`
  * test this website 👆

.changeset/fast-glasses-change.md
```diff
@@ -0,0 +1,8 @@
+---
+'@myst-theme/jupyter': major
+'@myst-theme/common': major
+'@myst-theme/article': major
+'@myst-theme/book': major
```
stevejpurves, line 5: @agoose77 @rowanc1 @choldgraf does this warrant a major release?

PR #572, opened 2025-04-04
⬆️ Bump Node.js version

agoose77:
In debugging a difference between the deployed and locally built themes, I noticed that we're only using Node.js 16. We purport to require >16 for MyST-MD, and we should really bump this to a newer version.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,10 +15,10 @@ jobs:
       - name: Checkout Repo
         uses: actions/checkout@v3
 
-      - name: Setup Node.js 16.x
+      - name: Setup Node.js 18.x
         uses: actions/setup-node@v3
         with:
-          node-version: 16.x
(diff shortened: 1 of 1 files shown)


stevejpurves:
@agoose77 v22 is the current LTS version, any reason not to just jump to v22? (I've been using it on our stack without issue.)

.github/workflows/release.yml
```diff
@@ -15,10 +15,10 @@ jobs:
       - name: Checkout Repo
         uses: actions/checkout@v3
 
-      - name: Setup Node.js 16.x
+      - name: Setup Node.js 18.x
```
agoose77, line 18:
```suggestion
      - name: Setup Node.js 22.x
```

agoose77:
Actually, let's stick to v18 so that our tests are in-sync.

PR #576, opened 2025-04-15
🖼️ Fix layout of `justified` landing blocks

### LINKED ISSUE #575: Landing page links steal from `justified` content width
The layout design for `justified` sections is such that links steal from the available content. The reason for this is due to the flexbox-based design that ensures links end up below content on narrow screens:

**Wide**:  
Image

**Narrow**:
Image

### Solution
- Use CSS grid to move layout according to media queries: https://play.tailwindcss.com/QyJzYdB1cu

agoose77:
Fixes #575 by switching to a CSS grid layout, rather than a set of flex boxes. This ensures that we can responsively move links _below_ the main body of the section, and prevent links from stealing _body_ width.

Screenshot 2025-04-15 at 15-00-08 Project Pythia

### DIFF
--- a/.changeset/quick-needles-judge.md
+++ b/.changeset/quick-needles-judge.md
@@ -1,0 +1,6 @@
+---
+'@myst-theme/landing-pages': patch
+'@myst-theme/styles': patch
+---
+
+Fix justified landing-page block layout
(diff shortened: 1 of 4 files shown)


agoose77:
Given that landing pages are something I'm pushing on, and this improves the existing UX, I'm merging. I don't think it should take energy from the rest of the team to review given its experimental nature.

choldgraf:
IMO this is a safe and reasonable decision - especially since this is fixing a bug. thanks Angus.

PR #585, opened 2025-05-21
🗽 Liberate the footer from the article-grid

stefanv:
It's mighty tight if you *have* to fit the footer inside the article-grid!

I think this is also more along the line of what we discussed: that the footer should have a large amount of flexibility in how it is laid out.

### DIFF
--- a/themes/book/app/components/Footer.tsx
+++ b/themes/book/app/components/Footer.tsx
@@ -7,7 +7,7 @@ export function Footer({ content, className }: { content: GenericParent; classNa
     // Outer footer, sets up the grid, leaves margin above
     <footer
       className={classNames(
-        'article footer article-grid bg-white dark:bg-slate-950 mt-10 shadow-2xl shadow py-10',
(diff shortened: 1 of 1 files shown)


agoose77:
@stefanv the intention here is to give the footer access to the grid defined by `article-grid`. If you want to make it wider, you can by adding `col-screen` to the footer. I'll make a PR to your landing-page repo!

stefanv:
Ah, OK! Is that a tailwind thing? A Myst thing? Remnant from bootstrap 😅

agoose77:
The grid system is defined by the theme, and has some documentation here: https://jupyter-book.github.io/myst-theme/?path=/docs/components-grid-system--docs

There's some CSS now in book-theme that sets the grid column for the footer to be `col-page` whereas it's `col-body` for the main article.

agoose77:
As such, are you happy for me to close this PR?

choldgraf:
I just want to applaud the use of PR title and emoji, that was top-notch @stefanv

stefanv:
Thanks @agoose77!

PR #586, opened 2025-05-21
🐬 Flip `main` and `article` elements

### LINKED ISSUE #584: Fix theme HTML5 structure
The `book-theme` (at least) generates an HTML structure of the form
```html
<article ...>
  <main ...>
```
This is prohibited by [the HTML spec](https://www.w3.org/TR/html-main-element/):

> Authors MUST NOT include more than one main element in a document.
>
> Authors MUST NOT include the main element as a child of an article, aside, footer, header or nav element.

We should correct this.

agoose77:
This PR should fix #584 for now, although we will want to revisit this topic more closely in future to consider all kinds of page content (especially if footers become per-page).

A subtle change here is that the `.article` styling now affects the inner element (`main article.article`) instead of the outer element (`article.article:has(> main)`).

I think this is an uninvasive change — the outer `main` has no children besides `article.article`.


themes/book/styles/grid-system.css
```diff
@@ -7,8 +7,7 @@
 */
   :where(footer).article-grid > *,
   :where(footer).article-left-grid > *,
-  :where(footer).article-center-grid > *,
```
agoose77, line 10: no-op style change

stevejpurves:
My commit above fixes article theme. We are left with the following DOM layouts in both themes. 

## Article

image


## Book

image

So, the requirement for one main is satisfied, and the requirement that article is within main is also satisfied. 

On merging this, I suggest we merge this once we are everyone is happy that the conversation and the related issue: https://github.com/jupyter-book/myst-theme/issues/584 on the semantics of where articles should be placed with respect to footer, header, content, etc... is concluded.

Note: I did try to move the article tab further up so that it included the article header, but that immediately breaks styling.

agoose77:
@stevejpurves how do you feel about just merging this as-is without getting full agreement on what the rest of the document should do? I don't think it's worth blocking on to have a deeper conversation -- I can always create a new issue if I/we find another part of the DOM is lacking.

stevejpurves:
@agoose77 I was happy with this based on the testing I carried out, and think it can be merged.

PR #587, opened 2025-05-21
🤏 Fix grid-system footer override syntax error

stefanv:
Remove a spurious comma that was left in the style file for book-theme.


stefanv:
I think some changes aren't correctly picked up by the build, because I did start seeing these errors after a fresh build. Thanks for merging!

Issue #391, opened 2024-05-16
Serve theme documentation from this repository and add PR previews


choldgraf:
Currently, we store the [documentation for our bundled myst themes in the mystmd repository](https://github.com/executablebooks/mystmd/blob/main/docs/website-templates.md#site-options). The documentation in _this_ repository is a demo site of components for references.

By having the documentation for the theme in a separate repository, we have a few downsides:

- Changes in functionality to this theme will require changes to documentation in a different space (mystmd.org).
- Changes in functionality to this theme aren't easily previewed in PRs, because no local documentation is built as a preview.

I propose that we do the following things:

- move all documentation that is related to the bundled myst themes into this repository (in `docs/`)
- move the components site to a sub-directory like `docs/components/`
- add a netlify PR preview for the documentation that uses the latest theme build in this repository
- *keep* templating/theme documentation in mystmd if it is about web templating in general, and not these themes specifically.
- *embed* the documentation for these themes into the mystmd.org docs so that it functionally looks the same that it does now (perhaps with "myst default themes" having a dedicated page)

agoose77:
I think you've identified some good points of improvement for our two-repo ecosystem. At the heart of this issue is the fact that we have three kinds of documentation:

- MyST documentation (CLI)
- MyST theme documentation
- MyST theme component documentation

The problems we need to solve are:

We're currently taking the approach that most users should just go to `mystmd.org` for documentation in this ecosystem -- a departure from the JB1 approach which has docs split around the ecosystem. That introduces two problems; fragmentation and repetition of information _and_ fragmentation of resources (i.e. multiple sites in your search engine).
However, this does mean that the coupling between components in terms of where the code lives is not so obvious. 

So, my summary of the problems we need to solve in this space is:

1. We should not bless our own web themes _too_ much; we want it to be obvious that developers can add their own _and users can look for them_.
3. We should have theme component documentation (Storybook)
2. We should make it possible to preview changes to our theme components
4. We should have documentation on our pre-designed themes
5. We should make it possible to preview changes to our pre-designed themes

It might not be super easy to solve all of these; there may be some trade-offs. But, I'll take a stab at something and see where we end up.

choldgraf:
[...]

PR #589, opened 2025-05-22
Myst theme standalone docs

stefanv:
This is showing how we can refactor the `myst-theme` docs into a standalone build, that we can then link to from the mystmd docs.

Don't pay attention to the content for now (all `.md` files, 452321ef7a5e56f0927a1fa279ec94c33369d2e0). I did not try and curate them, just grabbed some web related pages to how this would work.

This build includes the storybook React component docs, to make them more visible too.


stefanv:
Sure! I have no desire to implement that, personally, just wanted to see if we can move the docs to be co-located with the theme, and to expose the component docs.

choldgraf:
Since splitting out the repositories would be more complicated, perhaps we can make progress on this with less risk by starting to _centralize_ the documentation for `myst-theme` with the `myst-theme` repository. That way we aren't splitting docs and code across two different spaces.

I cleaned up the issue body of https://github.com/jupyter-book/myst-theme/issues/391 to track some next actions that could improve it. I think we need to agree on that issue in-principle, and then PRs like this could be steps towards implementation. I'd recommend we start off by creating a simple `myst-theme` docs site that lives on its own, and over time we can migrate. Does that work?

PR #595, opened 2025-05-27
🗺️ Upgrade to `react-router: 7`

agoose77:
We are currently using Remix 1.x, and the latest version is effectively Remix 3.x (react-router 7). This PR is a spike that reworks a previous upgrade attempt to use react-router. It needs a close eye, and only focuses on the book theme.

My intention here is to leave this open when I get some time to move it forward, and to explore the interesting new features like basepath handling, prerendering, etc.

Ignore the blame and commit count -- the commits need to be squashed into something more manageable.


themes/book/tsconfig.json
```diff
@@ -15,8 +18,9 @@
     "paths": {
       "~/*": ["./app/*"]
     },
+    "rootDirs": [".", "./.react-router/types"],
 
-    // Remix takes care of building everything in `remix build`.
+    // Vite takes care of building everything in `remix build`.
```
agoose77, line 23:
```suggestion
    // Vite takes care of building everything in `react-router build`.
```

agoose77:
The first thing that I think we might want to do is to have the theme proxy content-server requests. In other words, rather than the rendered web-page linking to https://my-cdn.com/static/foo.png, it links to `./static/foo.png`, which in turn resolves the CDN. The benefit of this is that people running theme applications do not need their CDNs to be externally visible (or proxied alongside the application to make it work).

Practically, this means instead of rewriting content links in the theme to `${CONTENT_CDN}/${path}` we rewrite to `./static/${path}`.

PR #599, opened 2025-06-03
Add flex-wrap to tabs

artoftheblue:
I suggest that this is necessary for really long tab sets so that all the tabs would fit by default. Right now, a scroll bar is created, which is less user-friendly, which seems to go against MyST values.

### DIFF
--- a/packages/myst-to-react/src/tabs.tsx
+++ b/packages/myst-to-react/src/tabs.tsx
@@ -52,7 +52,7 @@ export function TabSet({
   return (
     <TabSetStateProvider active={active}>
       <div className={classNames('my-5', className)}>
-        <div className="flex flex-row overflow-x-auto border-b border-b-gray-100">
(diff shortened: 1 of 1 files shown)


agoose77:
@artoftheblue thanks for this PR! Out of curiousity, would you be happy to share the content in question? I don't have a good feel for a scenario in which having many wide tabs makes sense. We do need a good story for `sm` displays, so this PR is likely an improvement over scrollbars in that context.

choldgraf (approved):
IMO this is a better UX than requiring a horizontal scrollbar. It feels easier to navigate and click through. I checked how Bootstrap does this and they do the same thing:

https://getbootstrap.com/docs/5.0/components/navs-tabs/

CleanShot 2025-06-06 at 12 16 00

choldgraf:
I'm gonna go ahead and merge this one because I think it's a clear improvement, and we are building a team culture that [biases towards action](https://github.com/jupyter-book/team-compass/issues/27) if it feels safe to merge PRs. ✨

PR #602, opened 2025-06-05
🏷️ Add missing `id` to notebook `block` renderer

agoose77:
Fixes https://github.com/jupyter-book/mystmd/issues/2085 by setting the block ID to `html_id`, `identifier`, or `key`, and ensures that the other IDs are unique. It also adds support for `html_id` in the non-nb block.

### DIFF
--- a/packages/jupyter/src/block.tsx
+++ b/packages/jupyter/src/block.tsx
@@ -27,8 +27,8 @@ export function NotebookBlock({ node, className }: { node: GenericParent; classN
               <NotebookRunCellSpinnerOnly id={`spinner-${node.key}`} />
             </div>
           </div>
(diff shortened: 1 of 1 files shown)


stevejpurves:
quantecon theme aside, @agoose77 have you already confirmed that these changes do not affect in browser execution in the core themes here? (i.e. notebook execution and execution of embed/figures that use notebook outputs)

stevejpurves:
@agoose77 this looks OK. It definitely isn't breaking the in-browser execution, which I thought it might. So I'd say this is probably fine. 

One thing that is making me hesitate is [this issue](https://github.com/jupyter-book/mystmd/issues/2103) that I ran into during testing. Which is probably unrelated, but of course I'm not 100% sure.

PR #604, opened 2025-06-06
Make Jupyter tables readable in dark theme

artoftheblue:
- Restructured the `jupyter.css` file to use Tailwind CSS
- Used the latter to apply different styles to the table depending on the theme, making it readable, fixing https://github.com/jupyter-book/mystmd/issues/1595
- Hovering over a row now highlights the entire row instead of just the header column of that row
- - Hovering over a table uses the same highlight color regardless of line parity within the [dark | light] theme

---

_Everything works on my end, but make sure to check everything carefully because this is my first PR within this repo; also, this is my first time using changesets so in case more details are necessary, do not hesitate to change something_

_(edit: thinking about it now, I would prefer the bullet points above to be appended to the changeset so that other people could more clearly see what has been fixed)_

### DIFF
--- a/.changeset/slimy-nights-fail.md
+++ b/.changeset/slimy-nights-fail.md
@@ -1,0 +1,5 @@
+---
+'@myst-theme/site': minor
+---
+
+Make Jupyter tables readable in dark theme
--- a/styles/jupyter.css
+++ b/styles/jupyter.css
@@ -15,7 +15,7 @@
 /* Here we use important to overwrite mt from typography.css */
 
 table.dataframe {
-  @apply border-none border-collapse border-spacing-0 text-black text-xs table-fixed m-0 !important;
+  @apply border-none border-collapse border-spacing-0 text-xs table-fixed m-0 !important;
 }
 
 .dataframe thead {
--- a/styles/jupyter.css
+++ b/styles/jupyter.css
@@ -15,7 +15,7 @@
 /* Here we use important to overwrite mt from typography.css */
 
 table.dataframe {
-  @apply border-none border-collapse border-spacing-0 text-xs table-fixed m-0 !important;
+  @apply border-none border-collapse text-black border-spacing-0 text-xs table-fixed !m-0;
 }
 
 .dataframe thead {
--- a/styles/jupyter.css
+++ b/styles/jupyter.css
@@ -1,41 +1,36 @@
 .font-system {
   font-family: Menlo, Consolas, 'DejaVu Sans Mono', monospace;
 }
+
 .jupyter-error {
-  background-color: rgb(255, 221, 221);
+  @apply bg-red-100;
 }
+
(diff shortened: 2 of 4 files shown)


styles/jupyter.css
```diff
@@ -1,47 +1,43 @@
...
+
 table.dataframe {
-  border: none;
-  border-collapse: collapse;
-  border-spacing: 0;
-  color: black;
-  font-size: 1em;
-  table-layout: fixed;
-  margin: 0 !important;
+  @apply border-none border-collapse border-spacing-0 text-xs table-fixed m-0 !important;
```
artoftheblue, line 18: Check whether I've used the !important tag correctly, because I think it might be overriding something else (thus why black text was in both themes before a recent commit)

artoftheblue: I messed up a bit with last 4 commits, but the diff is as intended

choldgraf (approved):
I tested this locally and it improved the style for dataframes quite a lot! Thanks so much @artoftheblue for the contribution!

One request: could you fix the linter error? I think you need to run something like the following command in the jupyter folder:

```
prettier --check "src/**/*.{ts,tsx,md}" --write
```

here's a GIF:

CleanShot 2025-06-06 at 11 34 20

In doing that, I had to spin up my own theme server, and I used this opportunity to improve our documentation on that a little bit. I've pushed that to this PR (hope that's OK @artoftheblue ).

In my opinion this is an improvement over what's there and I'd be happy to merge as-is. I'll wait until early next week to see if somebody more knowledgeable about tailwind etc wants to object first.

artoftheblue:
@choldgraf done! 

_p. s. still weird commit issues though_

choldgraf:
Don't worry about the commit weirdness - the diff is the important thing and we [squash-merge by default anyway](https://compass.jupyterbook.org/practices#when-merging-prs-use-squash-commits).

artoftheblue:
Once again messed up with the commit structure; diff didn't change

styles/jupyter.css
rowanc1: I think this should be `!m-0` ([docs](https://v3.tailwindcss.com/docs/configuration#important-modifier)).

```suggestion
  @apply border-none border-collapse border-spacing-0 text-xs table-fixed !m-0;
```

.changeset/slimy-nights-fail.md
```diff
@@ -0,0 +1,5 @@
+---
+'@myst-theme/site': minor
```
rowanc1, line 2:
```suggestion
'@myst-theme/site': patch
```

artoftheblue:
@rowanc1 looks identical to when I fixed the "dark text on dark background" my own way, but translates the original CSS better

Everything looks great! You or @choldgraf can merge the PR

PR #605, opened 2025-06-07
Add custom CSS markers to myst-to-react classes

### LINKED ISSUE #566: Make all React components targetable through CSS classes
Presently, we only annotate a single DOM node with a class name from the MyST AST. This requires users that want more comprehensive styling to style-by-position rather than by class. We can probably improve this by adding explicit names to each React component.

### LINKED ISSUE #598: Extract inline CSS into CSS classes
Currently it is not possible to easily customize the site style. There could be a better approach, but this is the structure of how I envision it.

1. For instance, if a class for a site block is defined in the myst theme, like `.grid-area-subtitle`, it's trivial to customize it using a `.css` style file.
2. Whereas the directives (e.g., tabs, admonitions) I'd like to customize do not have CSS classes that would be limited just to them.

## Proposal

Split sets of Tailwind CSS responsible for defining style for different site parts into external `.css` files. This way styles corresponding to each directive would not be hard-coded.

An example of how I imagine this based on [this Tailwind CSS example](https://github.com/jupyter-book/myst-theme/blob/3fbd8c7fe81674dbe6bc5eb5e71979f5806cd802/packages/icons/src/myst.tsx#L40).

```css
.made-with-myst-text {
  @apply flex mx-auto text-gray-700 w-fit hover:text-blue-700 dark:text-gray-200 dark:hover:text-blue-400;
}
```

However, the inclusion of classes in blocks heavily depends on the properties of, say, [admonitions](https://github.com/jupyter-book/myst-theme/blob/3fbd8c7fe81674dbe6bc5eb5e71979f5806cd802/packages/myst-to-react/src/admonitions.tsx#L195):

```tsx
          className={classNames('m-0 font-medium py-1 flex min-w-0', {
            'text-lg': !simple,
            'text-md': simple,
            'bg-gray-100 dark:bg-stone-700': simple,
            'text-blue-600 bg-blue-50 dark:bg-slate-900': !simple && (!color || color === 'blue'),
```

Therefore, the original example would not be sufficient; thus, the CSS structure I imagine would be as follows:

```tsx
         className={classNames('admonition-base', {
            'admonition-simple-text-style': !simple,
            'admonition-base-text-style': simple,
            'admonition-simple-text-color': simple,
            'admonition-note-text-color': !simple && (!color || color === 'blue'),
```

```css
.admonition-base {
  @apply m-0 font-medium py-1 flex min-w-0;
}

.admonition-simple-text-style {
  @apply text-lg;
}

.admonition-base-text-style {
  @apply text-md;
}

.admonition-simple-text-color {
  @apply bg-gray-100 dark:bg-stone-700;    
}

.admonition-note-text-color {
  @apply text-blue-600 bg-blue-50 dark:bg-slate-900;
}
```

Would this be an adequate way to integrate more user-customizable CSS? If so, I could work on converting it to this format.

### LINKED ISSUE #600: Improve usability of custom CSS stylesheets
Presently it is somewhat inconvenient to style specific elements. With MyST `class` annotations, we can target those elements on the page, but in many components the choice of which DOM node obtains this class is somewhat arbitrary. Furthermore, navigating relative to these nodes as a root is nontrivial, and oftentimes requires arcane selectors.

A secondary issue relates to general styling (e.g. make all admonitions red). Right now we don't expose "is-a" information at the CSS level.

artoftheblue:
- fixes #566 
- fixes #598 
- fixes #600 (⚠️ make sure that there aren't any other MyST classes that should be targeted)

## To-do list

- [x] Add custom CSS markers to all myst-to-react classes. Files:
  - [x] admonitions
  - [x] aside
  - [x] card
  - [x] code
  - [x] dropdown
  - [x] exercise
  - [x] grid
  - [x] proof
  - [x] tabs

## Overview

Tested out how custom declarations of CSS classes interact with existing classes and each other, and it works as follows:

1. Top-level classes overwrite empty properties of classes of same and lower levels
2. Since the user's custom style sheet gets imported last, all their declarations take precedence over myst-theme's classes
3. Within same-level classes, the one that is defined last in the custom style sheet takes precedence

Therefore, a `@layer` structure from Tailwind CSS does not seem necessary since currently child classes correctly inherit properties of parent classes and it is possible to set up a style sheet so that the user could overwrite a custom subset of classes

## CSS Structure

Explicit names of React components I have introduced in this PR:

```
#### Admonitions

-admonition
  -admonition-header
    -admonition-header-icon
    -admonition-header-text
  -admonition-body
  
#### Asides

-aside
  -aside-${kind}
    -aside-title
    -aside-body

  -aside-default
    -aside-default-title
    -aside-default-body

#### Cards

-card
  -card-header
  -card-body 
    -card-title
  -card-footer

#### Dropdowns

-dropdown
  -dropdown-header
    -dropdown-header-title
    -dropdown-header-icon
  -dropdown-body

#### Code

-code
  -code-filename-row
    -code-filename-icon
    -code-filename-title
  -code-body
  -code-copy-icon

#### Exercise

-exercise
  -exercise-header
    -exercise-title
  -exercise-body

#### Grids

-grid

#### Proofs

-proof
  -proof-header
    -proof-title
  -proof-body
  
#### Tabs

-tab-set
  -tab-set-row
    -tab-item-header
    -tab-item-header-active
    -tab-item-header-inactive
      -tab-item-body
        -tab-item-text
```

### DIFF
--- a/packages/jupyter/src/controls/NotebookCellControls.tsx
+++ b/packages/jupyter/src/controls/NotebookCellControls.tsx
@@ -5,7 +5,7 @@ export function NotebookRunCell({ id }: { id: string }) {
   const { ready, cellIsExecuting, notebookIsBusy, execute } = useCellExecution(id);
   if (!ready) return null;
   return (
-    <Run ready={ready} executing={cellIsExecuting} disabled={notebookIsBusy} onClick={execute} />
+    <Run ready={ready} executing={cellIsExecuting} disabled={notebookIsBusy} onClick={execute} title="Run cell" />
   );
 }
 
--- a/packages/myst-to-react/src/admonitions.tsx
+++ b/packages/myst-to-react/src/admonitions.tsx
@@ -175,16 +175,16 @@ export function Admonition({
       dropdown={dropdown}
       open={open}
       className={classNames(
-        '-admonition-wrapper my-5 shadow-md dark:shadow-2xl dark:shadow-neutral-900',
+        '-admonition my-5 shadow-md dark:shadow-2xl dark:shadow-neutral-900',
         'bg-gray-50/10 dark:bg-stone-800',
         'overflow-hidden',
         {
-          '-admonition-wrapper-default rounded border-l-4': !simple,
-          '-admonition-wrapper-simple border-l-2': simple,
-          '-admonition-wrapper-blue border-blue-500': !color || color === 'blue',
-          '-admonition-wrapper-green border-green-600': color === 'green',
-          '-admonition-wrapper-yellow border-amber-600': color === 'yellow',
-          '-admonition-wrapper-red border-red-600': color === 'red',
+          '-admonition-default rounded border-l-4': !simple,
+          '-admonition-simple border-l-2': simple,
+          '-admonition-blue border-blue-500': !color || color === 'blue',
+          '-admonition-green border-green-600': color === 'green',
+          '-admonition-yellow border-amber-600': color === 'yellow',
+          '-admonition-red border-red-600': color === 'red',
         },
         className,
       )}
@@ -236,7 +236,7 @@ export function Admonition({
           )}
         </HeaderElement>
       )}
-      <div className={classNames('-admonition-paragraph-text px-4', { 'py-1': !simple, 'details-body': dropdown })}>
+      <div className={classNames('-admonition-body px-4', { 'py-1': !simple, 'details-body': dropdown })}>
         {children}
       </div>
     </WrapperElement>
--- a/packages/myst-to-react/src/admonitions.tsx
+++ b/packages/myst-to-react/src/admonitions.tsx
@@ -193,14 +193,14 @@ export function Admonition({
         <HeaderElement
           dropdown={dropdown}
           className={classNames('myst-admonition-header m-0 font-medium py-1 flex min-w-0', {
-            'myst-admonition-header-default text-lg': !simple,
-            'myst-admonition-header-simple text-md': simple,
+            'text-lg': !simple,
+            'text-md': simple,
             'bg-gray-100 dark:bg-stone-700': simple,
             'text-blue-600 bg-blue-50 dark:bg-slate-900': !simple && (!color || color === 'blue'),
             'text-green-600 bg-green-50 dark:bg-slate-900': !simple && color === 'green',
             'text-amber-600 bg-amber-50 dark:bg-slate-900': !simple && color === 'yellow',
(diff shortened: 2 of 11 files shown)


choldgraf:
It sounds like this is still work-in-progress, but just noting that in my opinion this would be really helpful. It is really nice to have semantically-meaningful CSS tags to write your own custom rules.

artoftheblue:
@choldgraf definitely a work in progress! I have created this pull request early to make what I'm working public and get others' opinions

artoftheblue:
As of now, all MyST-specific classes I was originally interested in have CSS markers

artoftheblue:
@choldgraf if I were to target a specific color class and replace it with some other color, I might replace all instances of that color on the page, which personally isn't ideal.

Overall, I'd prefer to have each type of admonition targetable in case the user wants to separately customize each admonition type. For instance, they could define custom colors for the background or text style specifically for, say, all blue-colored admonitions.

e. g. redefining text size wouldn't work targeting the in-build Tailwind CSS if the text class is used anywhere else (which it is) as it would override it in too many unwanted places.

Ideally, adding customizations to each kind of admonition would be perfect (alternatively, this feature could be implemented by a plugin to simply create custom admonitions)

choldgraf:
@agoose77 I'd suggest you recommend a namespace / prefix that seems reasonable, then we can merge this in and unblock @artoftheblue , and in the documentation we can note the prefix we've used and warn it may change. In the future if we change it then it should be easy for any developer to update their rules, and we'll be clear that we don't promise stability here for now.

choldgraf:
+1 on both the myst- prefix and also on using chained selectors (`.myst-admonition.bg-blue`) rather than color-specific semantic selectors (`myst-admonition-blue`). We can document that expected behavior and recommended practices to clarify

artoftheblue:
Thanks for all the feedback! I should've integrated everything discussed above in the new commit.

Another thing to clarify: is there a way to target dropdown vs non-dropdown classes, as well as simple vs default callouts? Do you recommend this to remain semantic or is there a better way?

artoftheblue:
@choldgraf could I request a review of these changes?

choldgraf (approved):
I think this is safe to merge - @agoose77 since you've made the latest review I'll hold off on merging and let you do so tomorrow in case there's something you want changed. If you don't do anything then I'll merge this in the next couple of days when I have a moment to get back to it.

agoose77:
This looks good to me! I'll merge if once the tests pass.

choldgraf:
wooo thanks Angus!

PR #611, opened 2025-06-14
👽 Render external links from TOC

stefanv:


packages/site/src/components/Navigation/InlineTableOfContents.tsx
```diff
@@ -19,7 +19,7 @@ export const InlineTableOfContents = ({
   if (!headings) return null;
   return (
     <nav aria-label="Table of Contents" className={className} ref={sidebarRef}>
-      <Toc headings={headings} />
+      <Toc headings={headings} open_urls_in_new_tab={config?.options?.open_toc_urls_in_new_tab} />
```
fwkoch, line 22: I'll take a little more time to test these two toc url PRs in the next day or two - but I had one early comment now:

It feels like the default behavior should be external URLs open in new tabs. This is is the preference described in comments here: https://github.com/jupyter-book/mystmd/issues/2098. It's also the behavior of all non-`internal` links elsewhere in `myst-theme`, e.g. https://github.com/jupyter-book/myst-theme/blob/main/packages/myst-to-react/src/links/index.tsx#L126 - I think the only argument for opening in the same tab is that was how URLs in the toc worked in JB1 (I don't think we have to follow that if there are other good reasons not to).

If we agree that new tab should be the default, then we should change this option to `open_urls_in_same_tab` (i.e. `new_tab -> same_tab`). That way, when the option is unset by the user, the `undefined` value is false-y and automatically gives the correct behavior ("do not open in `same_tab`").

As a further benefit (in my opinion), this wording for the option reduces the ambiguity of toc URL vs. toc internal link discussed in https://github.com/jupyter-book/mystmd/issues/1445. If unset, the default is "urls - new tab, internal links - same tab" - but if `open_urls_in_same_tab` is `true`, all toc items are same tab... Does that make sense?

packages/site/src/components/Navigation/PrimarySidebar.tsx
```diff
@@ -189,7 +189,10 @@ export const PrimarySidebar = ({
               aria-label="Table of Contents"
               className="flex-grow overflow-y-hidden transition-opacity ml-3 xl:ml-0 mr-3 max-w-[350px]"
             >
-              <Toc headings={headings} />
+              <Toc
+                headings={headings}
+                open_urls_in_new_tab={config?.options?.open_toc_urls_in_new_tab}
```
fwkoch, line 194: One other thought - would we ever want to have different values for this option for different toc entries...?

If so, we couldn't do this as a single, site-wide option; it would have to be an optional field on each toc entry:

```
toc:
  - file: index.md
  - title: Example
    url: https://example.com
    same_tab: true
```
or something...

I suspect this toc-entry-level granularity is unnecessary at this point, but maybe a limitation to keep in mind.

packages/site/src/components/Navigation/InlineTableOfContents.tsx
stefanv: I changed the variable to `open_urls_in_same_tab`.

packages/site/src/components/Navigation/PrimarySidebar.tsx
stefanv: This has now been implemented: `open_in_same_tab`

packages/common/src/types.ts
```diff
@@ -21,6 +21,7 @@ export type Heading = {
   level: number | 'index';
   group?: string;
   enumerator?: string;
+  url?: string;
 };
```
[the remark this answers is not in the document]
choldgraf: it might be because the default assumption is that it's a Table of Contents, and so each item is a heading (e.g. an `H1`). With this PR we are breaking that assumption which might be why it now feels weird to call it a heading.

packages/site/src/components/Navigation/TableOfContentsItems.tsx
```diff
@@ -72,9 +78,32 @@ function LinkItem({
...
+        )}
+        to={heading.url}
+        onClick={() => {
+          onClick?.();
+          setOpen(false);
+        }}
+        target={target}
+      >
+        {`${heading.enumerator ? `${heading.enumerator} ` : ''}${heading.title}`}
+        <ArrowTopRightOnSquareIcon className="inline h-4 w-4 align-baseline ml-[0.2rem]" />
```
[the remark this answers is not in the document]
stefanv: I had trouble with vertical icon alignment, so I'm going to merge your suggestion; thanks!

PR #613, opened 2025-06-16
👣 Allow custom primary sidebar footer

artoftheblue:
Originally was supposed to implement functionality discussed in 

- fixes https://github.com/jupyter-book/mystmd/issues/2080

However, the PR was expanded to allow the user to define a custom footer in the primary sidebar to replace the default `Made with MyST ` footer.

The config is similar to how the site-wide footer is currently added:

```yaml
site:
  parts:
    footer: footer.md                              # side-wide
    sidebar_footer: sidebar_footer.md              # new addition
```

For example, creating the file named `sidebar_footer.md` with the following contents

```
:::{warning} Warning!
This site is currently WIP.
:::
```

renders as 

image

### DIFF
--- a/themes/book/app/components/SidebarFooter.tsx
+++ b/themes/book/app/components/SidebarFooter.tsx
@@ -8,7 +8,7 @@ export function SidebarFooter({ content }: { content?: GenericParent }) {
   }
 
   return (
-    <div className="myst-primary-sidebar-footer text-sm text-slate-900 dark:text-white [&>*]:m-0 [&_a]:underline [&_a]:font-semibold">
+    <div className="article footer myst-primary-sidebar-footer">
       <MyST ast={content} />
     </div>
   );


choldgraf:
I definitely agree we need to let users disable this if they wish. I have one suggestion but it would require a bit of extra work so it's not a huge deal if you dont have the bandwidth for it now:

Rather than making this a feature flag, could we make it an option to modify the _content_ in the sidebar footer?

For example rather than `hide_myst_branfing` call this `primary_sidebar_footer`. It could accept any myst markdown and parse it before inserting it. It would default to the branding language+image listed now, and if it were `none` or `false` etc, it would simply be removed.

choldgraf:
Just noting that a similar thing was added here for the site-wide footer:

- https://github.com/jupyter-book/myst-theme/pull/565

My assumption would be that this can follow a similar pattern. Perhaps @stefanv could provide guidance about how that was implemented and how this is similar or different?

stefanv:
> My assumption would be that this can follow a similar pattern.

Yes, I think this PR already uses part in the same way, so :+1: 

It could be useful to wrap in a special element or a div, so it can be styled in CSS. See https://github.com/jupyter-book/myst-theme/blob/main/themes/book/app/components/Footer.tsx#L8.

stefanv:
Would be good to also add some documentation for this feature.

agoose77:
@choldgraf see also!

themes/book/app/components/SidebarFooter.tsx
```diff
@@ -0,0 +1,15 @@
...
+import { MyST } from 'myst-to-react';
+import { MadeWithMyst } from '@myst-theme/icons';
+
+export function SidebarFooter({ content }: { content?: GenericParent }) {
+  if (!content) {
+    return <MadeWithMyst />;
+  }
+
+  return (
+    <div className="text-sm text-slate-900 dark:text-white [&>*]:m-0 [&_a]:underline [&_a]:font-semibold">
```
stefanv: Would be good to add a `myst-` class here to enable styling.
Link this to https://github.com/jupyter-book/myst-theme/pull/667 so that we can follow up there.

agoose77:
I'm not totally sure yet on whether we should add special styling to sidebars, etc. I don't think we do for footer, on the basis it can have wierd consequences for complex markup trees.

stefanv:
@agoose77 So, what would you like to see changed here before merge?

choldgraf:
My feeling is that this is a simple change that unblocks users who don't want the made with myst logo forced on them, or who otherwise want a custom sidebar footer. I'm not sure what you mean by "special styling", but I am fine adding any language to these docs to clarify that anything "special" is not supported, even going so far as to say only common mark is supported. 

It is easy to add more functionality in the future if need be, and I think we can bias towards incremental progress by merging this so long as there isn't a major strategic risk here.

My criteria for major strategic risks would be:

- we are taking an action we cannot undo or fix and that may do extreme harm to our users or maintainers
- OR, we could technically fix something but it will be a huge lift and we may not have the capacity to fix it

My suggestion here (and in every other PR) is that if concerns don't rise to this level, we should just merge things and move forward. For an under-resourced project like ours, slowing down too much is a bigger risk than moving too quickly.

themes/book/app/components/SidebarFooter.tsx
```diff
@@ -0,0 +1,15 @@
...
+import { MyST } from 'myst-to-react';
+import { MadeWithMyst } from '@myst-theme/icons';
+
+export function SidebarFooter({ content }: { content?: GenericParent }) {
+  if (!content) {
+    return <MadeWithMyst />;
+  }
+
+  return (
+    <div className="myst-primary-sidebar-footer text-sm text-slate-900 dark:text-white [&>*]:m-0 [&_a]:underline [&_a]:font-semibold">
```
choldgraf:
```suggestion
    <div className="article footer myst-primary-sidebar-footer">
```

themes/book/app/routes/$.tsx
```diff
@@ -102,7 +103,7 @@ function ArticlePageAndNavigationInternal({
       <PrimaryNavigation
         sidebarRef={toc}
         hide_toc={hide_toc}
-        footer={<MadeWithMyst />}
+        footer={<SidebarFooter content={projectParts?.primary_sidebar_footer?.mdast} />}
```
stefanv: As a follow-up, we may want to rename `footer` here to something more descriptive, since this does not refer to the main page footer.

Maybe `sidebarFooter`.

stefanv (approved):

themes/book/app/routes/$.tsx
choldgraf: IMO, ideally this would be primarySidebarFooter and it could live alongside secondarySidebarFooter one day. Or, call it something more generic like `primarySidebarEnd` to recognize that other things than footers might be placed there one day.

I think my suggestion would be we just try to re-use the JB1 structure naming as much as possible for the book theme, since it has been used and battle-tested for a long time now:

https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/layout.html

Issue #601, opened 2025-06-04
Nav buttons disappear for small & medium-sized windows on landing pages


artoftheblue:
**Browser:** Opera GX 119, **OS:** Windows 11 (although everything works the same in my other browsers and on my phone)

## Showcase

Showcase for small and medium window sizes, respectively:

![Image](https://github.com/user-attachments/assets/815cc020-bc40-4b86-a937-1a3f0ba17d52)

![Image](https://github.com/user-attachments/assets/b5ff15a9-a413-4d1f-95f6-58e74d63c943)

## Additional Details

If the TOC is not hidden, there is a menu button that shows the TOC and the nav menu buttons below it. However, hiding the TOC and resizing the window to be smaller removes the buttons from the navigation bar in the header and does not create any UI that would allow the user to access them.

## Possible Solutions

1. There is a "see more" button that houses all the action buttons in the header. I suggest placing all the missing nav buttons there, given that the action buttons are a part of the header.
2. Do not hide the "show menu" button on the left for small and medium window sizes, and only remove the TOC section from it, keeping the nav buttons intact.

P. S. to reproduce: examples are from my landing site at https://artemis.place

artoftheblue:
Effectively is a duplicate of #514

Issue #514, opened 2025-01-07
Hiding the primary sidebar also removes the header navigation on mobile


choldgraf:
When the sidebar (AKA the Table of Contents) is hidden with `hide_toc`, then the header navigation items are no longer accessible on mobile. For example, see [next.jupyterbook.org](https://next.jupyterbook.org).

Here's a video showing this happening.

https://github.com/user-attachments/assets/7776a707-a967-4395-9a33-507971907309

I think this might be related to:

- #458

Issue #462, opened 2024-09-05
when there's a toc but no nav, there should not be a separator


stevejpurves:
the separator should be hidden
<img width="605" alt="Site title 🔊 2024-09-05 16-01-16" src="https://github.com/user-attachments/assets/e7a01ce5-70ca-4421-837c-a571aac88eae">

agoose77:
That's an oversight. I thought we did do that, but must have missed it!

Issue #458, opened 2024-08-27
Ensure the `hide_toc` theme option behaves as expected


stevejpurves:
Since #444 this may not longer be the case as we not also have nav components in the sidebar.
For any changes consider not drilling `hide_toc` down into components and how they might map to `mobileOnly` & `hideToc` props.

Desired behaviour `hide_toc` is that the table of contents should never be shown.

### todo

* let's map out the user stories for end states

agoose77:
Adding to this (Steve and I were in a Zoom meeting), right now it feels like we have a lot of confusing state options internally. We identified that there are really three behaviours that we want to capture:

1. Users have a ToC, and maybe have a nav. They want the primary sidebar to be visible at all times (either as a drawer, or fixed on the page). They want the burger menu to appear only when the sidebar behaves as a drawer.
2. Users don't have a ToC, and have a nav. They want to `hide_toc` so that the primary sidebar is not visible _unless_ the primary nav needs to be folded into the sidebar. They want the burger menu to appear when the sidebar is visible and acting as a drawer.
3. Users don't have a ToC or a nav. They never want to see the primary sidebar or burger menu.

To me, that means `hide_toc` is the only configurable theme option. The rest is just reactive semantics.

Steve reminded me that we need to think about how to achieve this with SSR and hydration.

PR #614, opened 2025-06-16
🫣 Fix primary sidebar & header not showing navigation buttons correctly when ToC is hidden

artoftheblue:
Fixes these issues:

- fixes https://github.com/jupyter-book/myst-theme/issues/462
- fixes https://github.com/jupyter-book/myst-theme/issues/514
- fixes https://github.com/jupyter-book/myst-theme/issues/601
- Makes text styling consistent (text-medium) in primary nav to ensure parity with top nav 

Given what was discussed in #458 by @agoose77 and @stevejpurves:

> 1. Users have a ToC, and maybe have a nav. They want the primary sidebar to be visible at all times (either as a drawer, or fixed on the page). They want the burger menu to appear only when the sidebar behaves as a drawer.
> 2. Users don't have a ToC, and have a nav. They want to hide_toc so that the primary sidebar is not visible unless the primary nav needs to be folded into the sidebar. They want the burger menu to appear when the sidebar is visible and acting as a drawer.
> 3. Users don't have a ToC or a nav. They never want to see the primary sidebar or burger menu.

all these states are accounted for in this PR (please, check!); thus, it should

- fixes https://github.com/jupyter-book/myst-theme/issues/458

### DIFF
--- a/packages/site/src/components/Navigation/TopNav.tsx
+++ b/packages/site/src/components/Navigation/TopNav.tsx
@@ -112,7 +112,20 @@ export function TopNav({ hideToc, hideSearch }: { hideToc?: boolean; hideSearch?
     <div className="bg-white/80 backdrop-blur dark:bg-stone-900/80 shadow dark:shadow-stone-700 p-3 md:px-8 sticky w-screen top-0 z-30 h-[60px]">
       <nav className="flex items-center justify-between flex-nowrap max-w-[1440px] mx-auto">
         <div className="flex flex-row xl:min-w-[19.5rem] mr-2 sm:mr-7 justify-start items-center shrink-0">
-          {!hideToc && (
+          {(nav && hideToc) && (
+            <div className="block lg:hidden">
+              <button
+                className="flex items-center border-stone-400 text-stone-800 hover:text-stone-900 dark:text-stone-200 hover:dark:text-stone-100"
+                onClick={() => {
+                  setOpen(!open);
+                }}
+              >
+                <MenuIcon width="2rem" height="2rem" className="m-1" />
+                <span className="sr-only">Open Menu</span>
+              </button>
+            </div>
+          )}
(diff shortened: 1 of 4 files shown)


packages/site/src/components/Navigation/TopNav.tsx
```diff
@@ -112,7 +112,20 @@ export function TopNav({ hideToc, hideSearch }: { hideToc?: boolean; hideSearch?
...
+                onClick={() => {
+                  setOpen(!open);
+                }}
+              >
+                <MenuIcon width="2rem" height="2rem" className="m-1" />
+                <span className="sr-only">Open Menu</span>
+              </button>
+            </div>
+          )}
+          {(!nav || !hideToc) && (
```
artoftheblue, line 128: There should be a better way to do this, but I haven't been able to make, say, a ternary operator work

My idea was to use something like _(pseudo-code)_ `{"block", (nav && hideToc) ? "lg:hidden" : "xl:hidden"}`, but I keep getting syntax errors trying to use this approach

agoose77: Something like 
```jsx
<div className={classNames("block", (nav && hideToc) ? "lg:hidden" : "xl:hidden")}>
```

choldgraf (approved):
I'm going to merge this - here's my rationale:

1. This fixes a pretty big UX bug (no mobile nav if you hide the TOC, which is often the case on landing pages). See the footnote of this fixing the Jupyter landing page.[^1]
2. I'm pretty sure we've already got a round of review from @stevejpurves and @agoose77 and @artoftheblue has implemented their suggestion.

[^1]: For example, here's before and after (I **think** the vertical spacing is because of the extra horizontal taken up by the button).

    Main: CleanShot 2025-06-20 at 12 28 10@2x

    This PR: CleanShot 2025-06-20 at 12 28 46@2x

PR #619, opened 2025-06-24
Add bokeh example

### LINKED ISSUE #527: Support Bokeh Plots

choldgraf:
This demonstrates that bokeh plots work with this theme.

- closes #527

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,4 +47,4 @@ deploy-book:
 
 build-docs:
 	make build-book
-	cd docs && myst build -d --execute --html
+	cd docs && myst build -d --execute --html --strict
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -73,6 +73,27 @@ chart = alt.Chart(df.reset_index()).mark_circle().encode(
 chart
 ```
 
+### Bokeh
+
+```{code-cell}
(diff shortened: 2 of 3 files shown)


choldgraf:
I've added a cross-ref to that other example repository at the top, so they at least know about each other now. We can unify or reuse content later on when we have the cycles to figure that out. Merging now!

PR #620, opened 2025-06-25
🔗 Allow specifying custom URL for navigation bar logo

### LINKED ISSUE #615: Make nav bar logo link customizable
This is a new feature request:

Can we get an option to customize the link associated with the logo at top left of the nav bar in the book-theme?

Currently this is [hard-coded as a HomeLink to /](https://github.com/jupyter-book/myst-theme/blob/main/packages/site/src/components/Navigation/HomeLink.tsx).

It would be great to be able to customize this, perhaps with something like

```
site:
  template: book-theme
  options:
    logo: mylogofile
    logo_link: <URL>
```

The use case I'm thinking of is the nav bar for the [Project Pythia sites](https://projectpythia.org). There are several different MyST sites that share a common nav bar to present the look and feel of a single site. The desired behavior is that the Pythia logo link always points to the project home page, rather than `/` of whichever site we are currently viewing.

stefanv:
Closes #615

### DIFF
--- a/packages/site/src/components/Navigation/HomeLink.tsx
+++ b/packages/site/src/components/Navigation/HomeLink.tsx
@@ -6,11 +6,13 @@ export function HomeLink({
   logoDark,
   logoText,
   name,
+  url
 }: {
   logo?: string;
   logoDark?: string;
   logoText?: string;
   name?: string;
+  url?: string;
 }) {
   const Link = useLinkProvider();
   const baseurl = useBaseurl();
@@ -18,7 +20,7 @@ export function HomeLink({
   return (
     <Link
       className="flex items-center ml-3 dark:text-white w-fit md:ml-5 xl:ml-7"
-      to={withBaseurl('/', baseurl)}
+      to={url ? url : withBaseurl('/', baseurl)}
       prefetch="intent"
     >
       {logo && (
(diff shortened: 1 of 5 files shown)


choldgraf:
Could you add brief docs with intended behavior? Looks pretty good to me though !

stefanv:
I *think* the templates get documentation auto-generated for them, but will check.

stefanv:
@choldgraf Would you like documentation in addition to: https://mystmd.org/guide/website-templates#template-site-myst-book-theme

choldgraf:
Ah nice you are correct - ok this one looks pretty simple to me so I'm happy to merge it, thanks for the improvement!

packages/site/src/components/Navigation/HomeLink.tsx
```diff
@@ -6,19 +6,21 @@ export function HomeLink({
...
+  url?: string;
 }) {
   const Link = useLinkProvider();
   const baseurl = useBaseurl();
   const nothingSet = !logo && !logoText;
   return (
     <Link
       className="flex items-center ml-3 dark:text-white w-fit md:ml-5 xl:ml-7"
-      to={withBaseurl('/', baseurl)}
+      to={url ? url : withBaseurl('/', baseurl)}
```
agoose77, line 23: @stefanv I'm on mobile, so not super able to dig in to the code. Is this how we also handle the nav links? It seems like we might want to apply base url handling to the logo url too.

I know at one point we might not have included the base url with nav links, which needs (needed?) to be fixed!

PR #622, opened 2025-06-28
🐶 Add husky configuration

stefanv:
This adds the (optional) ability to run a pre-commit hook to lint changes.

I find it quite helpful, in that it prevents me from having to go back and run eslint or prettier on my commits. We've had it for a while on the mystmd repo without complaints.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,8 @@
     "publish:ci": "npm run clean && npm run build && changeset publish && git push --follow-tags",
     "bump:myst:packages": "npx npm-upgrade-monorepo -w 'packages/*' 'myst-* @myst-theme/* thebe*'",
     "bump:myst:themes": "npx npm-upgrade-monorepo -w 'themes/*' 'myst-* @myst-theme/* thebe*'",
-    "bump:myst": "npm run bump:myst:packages && npm run bump:myst:themes"
+    "bump:myst": "npm run bump:myst:packages && npm run bump:myst:themes",
+    "install-pre-commit": "husky"
   },
   "devDependencies": {
     "@changesets/cli": "^2.26.1",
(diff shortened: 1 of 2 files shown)


agoose77 (approved):
LGTM! it's opt in, so we can easily merge and iterate. One question - our built-in lint tasks have implementions (definitions) per-subpackage, I.e you can define a different lint rule in each package. Is there a way that we should be defining this to match?

stefanv:
This seems to only be the case for the `themes/` subdirectory.

So, if you want, we can do prefix detection and run different commands for files coming from those directories, via a Python script e.g.

package.json
```diff
@@ -55,5 +58,8 @@
     "node": ">=14.0.0"
   },
   "packageManager": "npm@8.10.0",
-  "dependencies": {}
+  "lint-staged": {
+    "*.ts": "eslint --config ./.eslintrc.cjs --fix",
+    "*.{js,jsx,ts,tsx,md,html,css,json}": "prettier --write"
```
stevejpurves, line 63: How does this align with the separate turborepo powered linting that we have in place in scripts already? can the lint staged call the respective npm scripts above?

Also will opening the pattern here beyond `ts, tsx, md` especially to include `js, css` mean that we are running prettier on the contents of `dist/` and other build output folders?

agoose77: This is the same question I ask above, although I'm not sure if my meaning came through as clearly as it does here.

stefanv: The lint-staged package lets us only lint files that were touched during the commit.

I am not sure how to use the existing scripts, since afaik they don't take file arguments.

stefanv: @stevejpurves I tried to make this work, but because the existing formatter grabs *all* files already, I get:

```
> prettier --write "**/*.{ts,tsx,md}" --write /home/stefan/p/myst/myst-theme/packages/site/src/pages/Error404.tsx
```

(where Error404.tsx is the file I touched).

If we systematically go through and add `format` and `format:files`, then I can use the `format:files` target to make it work, and `format` can use `format:files` internally.

stefanv: Re: files included, this is a pre-commit hook, so will only run on files touched in a commit.

stefanv:
We have this on the myst repo, FWIW: https://github.com/jupyter-book/mystmd/pull/1919

PR #634, opened 2025-07-22
📦 Move to React 19 and update `peerDependencies`

### LINKED ISSUE #827: Update the myst-theme to React 19
We currently run on an older version of React - we've discussed several times the idea that we should upgrade to version 19, and a few times others have said "we'd be able to do X if we upgrade to React 19". In the name of making sure our priorities are rooted in solving problems either for our users or for ourselves, I'm creating this issue to track this overall effort and the things it might unlock. (or, it'll just live as a little stub that gets closed once we merge https://github.com/jupyter-book/myst-theme/pull/634 who knows 😅)

### Things we've be able to do better if we had React 19

- @stefanv said we'd be able to use React-Router to more naturally handle URL redirects without hacky HTML generation

### References

- We have a PR that takes a stab at this here: https://github.com/jupyter-book/myst-theme/pull/634

rowanc1:
This disables storybook for now by (1) removing it from the npm workspace and (2) moving the github workflow out of the right place.

---

- closes #827


rowanc1:
Now that we have the myst-theme docs, I think storybook might be a bunch less useful, and perhaps we just simplify and remove it rather than keeping it alive?

I can confirm that the preview both works and is on React 19.1.0.

image

stefanv (approved):

rowanc1:
image

Tested some with jupyter outputs.

I cannot get the package-lock to do anything, even with overrides.

However, `bun` works on the first shot.

We could switch with this as well?

choldgraf:
Is it a problem that JupyterLab doesn't yet support React19? It sounds like:

- No, the curvenote team are already using jupyterlab components and they're running React 19, and haven't run into major issues.
- So it's safe to try React19

Is it fine to remove the storybook framework?

- Yes we think so. It's ok!

agoose77:
> Is it a problem that JupyterLab doesn't yet support React19? It sounds like:
> 
>     * No, the curvenote team are already using jupyterlab components and they're running React 19, and haven't run into major issues.
> 
>     * So it's safe to try React19
> 
> 
> Is it fine to remove the storybook framework?
> 
>     * Yes we think so. It's ok!

It's a problem insofar as it's a potential landmine if we encounter a breakage — we later can't easily revert this if `main` has moved on (e.g. we move to react router).

But, I think we *can* scope the React-18 revert to just the Jupyter components, which would mean that we could use e.g. a portal. Having an escape hatch is all we need.

packages/jupyter/src/components.tsx
```diff
@@ -8,12 +8,12 @@ export const MaybeLongContent = ({
...
   content?: string;
   path?: string;
-  render: (content: string) => JSX.Element;
+  render: (content: string) => React.JSX.Element;
 }) => {
   const { error, data } = useLongContent(content, path);
   if (error) {
     return (
-      <div className="myst-jp-long-error text-red-500">Error loading content: {error.message}</div>
+      <div className="text-red-500 myst-jp-long-error">Error loading content: {error.message}</div>
```
stefanv: Don't we typically use myst classes first?

packages/myst-demo/src/index.tsx
```diff
@@ -473,7 +473,7 @@ export function MySTRenderer({
           {previewType === 'DOCX' && (
             <div>
               <button
-                className="p-3 border rounded"
+                className="p-3 rounded border"
```
stefanv: Another random shuffle; looks fine, just confirm. A few more further down.

stefanv (approved):

package.json
```diff
@@ -69,5 +69,11 @@
   "patchedDependencies": {
     "@jupyter-widgets/controls@5.0.12": "patches/@jupyter-widgets%2Fcontrols@5.0.12.patch",
     "@remix-run/dev@1.17.1": "patches/@remix-run%2Fdev@1.17.1.patch"
+  },
+  "overrides": {
+    "react": "^19.2.4",
+    "react-dom": "^19.2.4",
+    "@types/react": "^19.2.14",
+    "@types/react-dom": "^19.2.3"
   }
```
rowanc1, line 75: This is needed for `thebe-core` override and some jupyter content.

rowanc1:
I am going to merge the 19 branch in, and we can do some more on main on the other side of these pretty large changes. To move back to react 18, if we run into anything, is a 4 line package.json change and a resolve of deps.

package.json
```diff
@@ -48,8 +48,8 @@
     "lint-staged": "^16.1.2",
     "npm-run-all": "^4.1.5",
     "prettier": "latest",
-    "react": "^18.2.0",
-    "react-dom": "^18.2.0",
+    "react": "^19.2.4",
+    "react-dom": "^19.2.4",
```
rowanc1, line 52: This is where to change if we want to go back to react 18.

PR #641, opened 2025-08-15
Add aria-label for DocumentOutline.tsx dropdown button

kevinlin1:
The dropdown button next to "Contents" in the secondary sidebar is missing an `aria-label`, which is flagged as an accessibility issue in Lighthouse. Following the `aria-label="Open Folder"` scheme for folder dropdowns in the primary sidebar

https://github.com/jupyter-book/myst-theme/blob/4ee2d3fa2315a73f000f1bab8cbebdc43a70e695/packages/site/src/components/Navigation/PrimarySidebar.tsx#L54

I added the `aria-label="Open Contents"` for the contents secondary sidebar dropdown.

### DIFF
--- a/.changeset/great-ads-wonder.md
+++ b/.changeset/great-ads-wonder.md
@@ -1,0 +1,5 @@
+---
+'@myst-theme/site': patch
+---
+
+Add `aria-label` for `DocumentOutline.tsx` dropdown button


stevejpurves (changes requested):
Look great @kevinlin1 - could you please add a changeset? https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md#i-am-in-a-multi-package-repository-a-mono-repo

PR #654, opened 2025-10-04
Fix id link for 'Link to Keywords' `HashLink`

### LINKED ISSUE #653: Link to Keywords is a broken link
The on-hover "Link to Keywords" element is a broken link since the anchor `#keywords` does not exist. Compare the broken implementation of `Keywords.tsx`:

https://github.com/jupyter-book/myst-theme/blob/939ad377990c32443ac0aab2675afa620bf71933/packages/site/src/components/Keywords.tsx#L15-L28

To the working implementation of `Abstract.tsx`:

https://github.com/jupyter-book/myst-theme/blob/939ad377990c32443ac0aab2675afa620bf71933/packages/site/src/components/Abstract.tsx#L17-L25

Line 18 of `Abstract.tsx` defines the `id` for the `h2`, which provides the destination for the `HashLink`. This is missing from `Keywords.tsx`

While we could easily add an `id` to the `Keywords.tsx` span, how important is the "Keywords" link? Unlike the abstract, the keywords aren't given a specific heading level so I'm wary of just adding the `id`.

I suggest removing the `HashLink` in `Keywords.tsx`. If this sounds good, I can submit a pull request.

kevinlin1:
Adds a `myst-keywords` ID to the anchor link that's generated in keywords in the abstract part.

---
- fixes #653


choldgraf:
rather than removing this link, I think it'd be simpler to just add the `id` of the thing that the link is supposed to point to. So I've merged main into this branch and added `id="myst-keywords"` to that link. I'll take a look at the docs preview and merge if it looks ok!

PR #656, opened 2025-10-05
Fix accessibility bugs in search form dialog

kevinlin1:
When the search form dialog is open (in the book theme), WAVE reports two accessibility errors:

1. The `input` element's `aria-labelledby` refers to a non-existent element `search-label`.
2. The `label` element is empty (no text, only an SVG magnifying glass icon).

This patch fixes both issues with accessibility in the search dialog by:

1. Correcting the `label` element's `id` to be `search-label`, rather than `search-list`. This seems to have been a typo since there is another element `search-list` that is used to present the list of search results.
2. The `label` element now has a hidden `span` with the text "Search query".

### DIFF
--- a/.changeset/five-toys-float.md
+++ b/.changeset/five-toys-float.md
@@ -1,0 +1,5 @@
+---
+'@myst-theme/search': minor
+---
+
+Fix accessibility bugs in search form dialog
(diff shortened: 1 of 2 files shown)


agoose77:
Thanks for this PR! I'm merging this in the spirit of "looks like a good improvement".

PR #658, opened 2025-10-08
🧜‍♀️ Upgrade mermaid to ^11

agoose77:
In re-building `jupyterlab-myst`, I ran into problems bundling `mermaid` with webpack because it pulls in `cytoscape` which didn't bundle properly. I could dig into why that is, but it's a lot easier to just update everything (which fixed it!)

@rowanc1 why is the mermaid import async here -- is this for explicit code splitting? Otherwise, can I hoist the import to the module level?

### DIFF
--- a/packages/diagrams/src/index.tsx
+++ b/packages/diagrams/src/index.tsx
@@ -1,14 +1,16 @@
 import type { NodeRenderer } from '@myst-theme/providers';
 import { useEffect, useId, useState } from 'react';
 import classNames from 'classnames';
+import type { Mermaid } from 'mermaid';
 
-async function parse(id: string, text: string): Promise<string> {
-  const { default: mermaid } = await import('mermaid');
-  return await new Promise<string>((resolve) => {
-    (mermaid as any).render(id, text, (code: any) => {
-      resolve(code);
-    });
-  });
+let _mermaid: Mermaid | undefined = undefined;
+async function loadMermaid(): Promise<Mermaid> {
+  if (_mermaid === undefined) {
+    const module = await import('mermaid');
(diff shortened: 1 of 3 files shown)


packages/diagrams/src/index.tsx
```diff
@@ -1,14 +1,16 @@
...
+import type { Mermaid } from 'mermaid';
 
-async function parse(id: string, text: string): Promise<string> {
-  const { default: mermaid } = await import('mermaid');
-  return await new Promise<string>((resolve) => {
-    (mermaid as any).render(id, text, (code: any) => {
-      resolve(code);
-    });
-  });
+let _mermaid: Mermaid | undefined = undefined;
```
agoose77:
```suggestion
// Follow previous flow of importing mermaid dynamically.
// This is just a refactor, it might not be necessary.
let _mermaid: Mermaid | undefined = undefined;
```

stefanv (approved):

packages/diagrams/src/index.tsx
stefanv:
```suggestion
let _mermaid: Mermaid | undefined = undefined;

```

packages/diagrams/src/index.tsx
```diff
@@ -3,8 +3,6 @@ import { useEffect, useId, useState } from 'react';
 import classNames from 'classnames';
 import type { Mermaid } from 'mermaid';
 
-// Follow previous flow of importing mermaid dynamically.
```
agoose77: Ah, I meant to leave that in! I'm sure someone will still find this thread if needs be.

PR #659, opened 2025-10-10
Add additional publish step for themes through GH releases

parmentelat:
this is a follow-up on https://github.com/jupyter-book/mystmd/issues/2317

in this first step, we add a GH workflow that publishes both book and article themes as GH releases attached to the current repo

this clearly is only a first step, but I would argue it should be merged right away

in particular this means a regular fork can use the same mechanism to publish its custom version by just setting a tag

see the issue for a status of the overall migration path


parmentelat:
As it is now, pushing a tag on github will trigger a release named after the tag
Let me know if that requires more work

stevejpurves:
Hi @parmentelat,

Thanks for the PR and it's great to have the longer thread linked for context! 

I had some questions/comments:

* Is the intent for this action to only publish the `book` and `article` themes or to also bump and publish all the dependent packages at the same time via `changesets`? (in which case we'd need to bring run `version` and `publish` too.)
* Were you aware of the existing `build-theme` step in the Makefile, or was there a reason you did not opt to call that - using the existing packaging mechansim would mean one source of truth for building the zip
* Should this not also run `make deploy-theme` so that the theme bundles on `myst-templates` are updated at the same time? and gh release version and the published templates are in sync at least wile that is the primary method for the myst CLI to download them?

parmentelat:
one final comment on this one

I've been using this locally for quite a while and it is convenient, but I always only set one tag at a time

so I'm thinking, the `tags: ['*']` policy might cause trouble with your process that seems to set numerous tags like @myst-theme/anywidget@1.3.0 and so on for each release, and I'm not sure of the result in this case, as we certainly don't want to clobber the releases space !

if that's a problem, maybe we can go for a more restrictive naming scheme, like
- trigger only on tags of the form `publish-foobar`
- which would result in the `foobar` artefacts becoming available

let me know if that sounds like a useful addition

choldgraf:
Why don't we just make this trigger with a release, rather than with any tags? Then when you wanted to make a new asset bundle you'd just make a release in your fork (assuming that is possible?). I think I that's the workflow we'd prefer to use here anyway.

So basically, when a release is made in GitHub's UI, from the CLI, etc, then our workflow would run and attach the theme assets to that release.

parmentelat:
makes a lot of sense, sounds simpler this way indeed

README.md
```diff
@@ -55,3 +55,25 @@ This updates the git repository, and sometimes is a large diff and can cause git
...
+### How to publish
+
+This normally happens automatically: when a changesets release publishes to NPM, the `release.yml` workflow creates the GitHub release and then calls this workflow to attach the theme zips to it.
+
+You can also trigger it manually by pushing a tag, which works on forks too — handy for publishing a custom build of the themes:
+
+```bash
+git tag my-test-tag
+git push origin my-test-tag
+```
```
choldgraf, line 87:
```suggestion
```

choldgraf (approved):
This looks good to me! One comment to update the docs a bit more, but +1 once that's done!

parmentelat:
as far as as I can tell the failing check has to do with the absence of GITHUB_TOKEN which is expected on this fork

rowanc1:
Talked in the community cafe — we are merging! Thank you so much @parmentelat for all the work on this first piece.

We will follow up on the next release to see how this works!

PR #673, opened 2025-11-08
Re-fix bug for hidden play button

choldgraf:
- Resolves https://github.com/jupyter-book/mystmd/issues/2392

I think the issue is that it was fixed here:

https://github.com/jupyter-book/myst-theme/pull/606/files#diff-d93664df46cc39b49aabfb97a5f33a50774bb57a7dbab6e563329cbf213aa1b4L30-R31

but then re-introduced here:

https://github.com/jupyter-book/myst-theme/commit/eb484d97#diff-d93664df46cc39b49aabfb97a5f33a50774bb57a7dbab6e563329cbf213aa1b4R30-R33

This just re-introduces the change

### DIFF
--- a/.changeset/two-eels-crash.md
+++ b/.changeset/two-eels-crash.md
@@ -1,0 +1,5 @@
+---
+'@myst-theme/jupyter': patch
+---
+
+Fix execution button positioning


choldgraf:
@agoose77 ok I reverted to both of the lines that were changed in that commit

PR #674, opened 2025-11-09
feat: add `hide_authors` option

agoose77:
This small PR refactors the author removal logic to act as a view transform. This is a bit ugly -- I think we want to systematise our site option handling to do something more elegant than prop drilling, but I think this is fine for now.


agoose77:
> don't have to do that now but head towards that - is that what you were thinking?

I think yes, over time.

PR #677, opened 2025-11-12
Rename XIcon to TwitterIcon from scienceicons

choldgraf:
This is breaking a bunch of in-progress PRs, so I thought I'd just fix it as an independent PR. THe package "scienceicons" seems to have renamed XIcon to TwitterIcon - I'm not sure why else this would randomly change. This just makes that rename here


choldgraf:
I just realized @agoose77 reverted this in the banner PR. @agoose77 could you explain why you did that? This was the only way that I could get `make-build` working locally. Is there another way?

PR #678, opened 2025-11-12
Update developer guide

choldgraf:
This tries to make our local development and contributing guide a bit more discoverable and navigable. It tries to put more actionable information towards the top of files, and simplifies it a bit so there is less text. It puts docs in a `docs/developer/` folder and links to that folder from the README instead of having one giant readme.

### DIFF
--- a/.changeset/lucky-files-fold.md
+++ b/.changeset/lucky-files-fold.md
@@ -1,0 +1,5 @@
+---
+'@myst-theme/book': patch
+---
+
+Add visual warning when no CSS is present
--- a/.changeset/yellow-kiwis-sin.md
+++ b/.changeset/yellow-kiwis-sin.md
@@ -1,0 +1,5 @@
+---
+'myst-to-react': patch
+---
+
+Consume title in iframe renderer
--- a/docs/reference/blocks.md
+++ b/docs/reference/blocks.md
@@ -103,3 +103,11 @@ graph LR
     B -->|Yes| C[Action]
     B -->|No| D[End]
 ```
+
+## IFrame
+
+```{iframe} https://mystmd.org
+:title: This is a test title
+
+Here's an iframe with a title
+```
--- a/packages/myst-to-react/src/iframe.tsx
+++ b/packages/myst-to-react/src/iframe.tsx
@@ -41,6 +41,7 @@ export const IFrame: NodeRenderer = ({ node, className }) => {
           width="100%"
           height="100%"
           src={node.src}
+          title={node.title}
           allowFullScreen
           allow="autoplay"
           style={{
--- a/themes/book/app/root.tsx
+++ b/themes/book/app/root.tsx
@@ -83,6 +83,54 @@ function createSearch(index: MystSearchIndex): ISearch {
   return createMiniSearch(index.records, options);
 }
 
+/*
+ * Component that shows a "no CSS loaded" warning when a page 
+ * loads without the built-in MyST stylesheet. This can happen on static builds
+ * when the BASE_URL doesn't match the deployment base URL.
+ */
+function NoCSSWarning() {
+  const CLIENT_THEME_SOURCE = `
+    (() => {
+            // Test for has-styling variable set by the MyST stylesheet
+            const node = document.currentScript.parentNode;
+            const hasCSS = window.getComputedStyle(node).getPropertyValue("--has-styling");
+            if (hasCSS === ""){
+                    node.showModal();
+            }
+
+    })()
+`;
+  return (
+    <>
+      <dialog
+        id="myst-no-css"
+        // Use inline styles to ensure styling without stylesheets
+        style={{
+          position: 'fixed',
+          left: '0px',
+          top: '0px',
+          width: '100vw',
+          height: '100vh',
+          fontSize: '4rem',
+          padding: '1rem',
+          color: 'black',
+          background: 'white',
+        }}
+        // Opening the modal sets an open attribute, so we need to disable the warning
+        suppressHydrationWarning
+      >
+        <strong>Site not loading correctly?</strong>
+        <p>
+          This may be due to an incorrect <code>BASE_URL</code> configuration. See{' '}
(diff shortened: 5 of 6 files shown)


docs/developer/local.md
```diff
@@ -0,0 +1,80 @@
...
+npm run dev
+```
+
+## Start a theme server
+
+Finally, start the theme server application, which will take the AST from the content server in Terminal 1 and render it for you to preview:
+
+```shell
+# Terminal 3
+$ npm run theme:book
```
stefanv: I learned yesterday while working on the theme that `theme:book` also does monitoring. Doing the `run dev` at the same time causes a lot of craziness on my machine, as both try to do the same thing.

Certain changes, like adding a new component, will not be picked up this way. For that, you need to do `npm run build`.

docs/developer/local.md
```diff
@@ -0,0 +1,80 @@
...
+CONTENT_CDN=https://remote.example.com npm run theme:article
+```
+
+## Preview components and UI with storybook
+
+We have a lightweight [storybook](https://storybook.js.org/) configuration, which is another way to preview the design of these components.
+
+This is the same tool that powers [the MyST Theme components documentation](https://jupyter-book.github.io/myst-theme/?path=/docs/components-introduction--docs).
+
+To use Storybook, run these commands:
```
stefanv:
```suggestion
To use Storybook:
```

docs/developer/static-previews.md
```diff
@@ -0,0 +1,33 @@
...
+**Install the dependencies to build the documentation**
+
+This includes both the latest version of MyST, as well as the Python dependencies to run documentation examples. Run the following command to do so:
+
+```shell
+$ pip install -r docs/requirements.txt
+$ npm install
+```
+
+Then, build the latest version of the theme and preview the local documentation with it
```
stefanv: Here we are one something new, how to use the built bundle. Is this included in `build-docs`? Then maybe a heading is required for the next part too.

docs/developer/tips.md
```diff
@@ -0,0 +1,12 @@
+# Developer tips and best practices
+
+## Using Providers for MyST content
+
+When using components that render MyST content (using `<MyST ast={content} />`), they must be wrapped wrap them in various providers e.g. `TabStateProvider` to ensure proper React context is available.
```
stefanv: I think this has little to do with `MyST` components. It might be something like:

`Providers` create [React contexts](https://react.dev/reference/react/createContext) for communicating app states across components. For example,
the `<ArticleProvider>...</ArticleProvider>` shares an `ArticleContext` with all its children, telling them the current article `kind`, `frontmatter`, and `references`.

docs/developer/tips.md
```diff
@@ -0,0 +1,12 @@
...
+These providers facilitate an alternative to prop drilling, the name used to describe passing props down through several React components. They are also used to isolate state between different subtrees, e.g. `TabStateProvider` defines a synchronised tab context.
```
stefanv:
```suggestion
These providers facilitate an alternative to prop drilling, the name used to describe passing props down several layers of React components.
```

>  They are also used to isolate state between different subtrees, e.g. `TabStateProvider` defines a synchronised tab context.

Maybe that should stay, but I'm not sure user has ability to understand without further explanation.

README.md
```diff
@@ -12,132 +12,19 @@ Packages for creating MyST websites themes using React and Remix.
 All dependencies for `myst-theme` are included in this repository (a monorepo!).
 The core themes are also included in this repository to aid in development.
 
-## What's inside?
+## Local development and live server preview
```
stefanv: Maybe just link to the docs, instead of replicating the TOC here, otherwise we have to keep in sync.

docs/developer/local.md
choldgraf: So what are the three incantations we should recommend here? Can you make an edit to the text to make this make sense?

stefanv: I can make a follow-up PR to clarify this.

docs/developer/tips.md
stefanv: This one still, then merge?

README.md
bsipocz: Could GH handle some include instead instead of just linking? 

I envision it could be useful for those who read the readme on the repo. But otherwise I agree don't copy paste any content as it will just get outdated really quickly

docs/developer/local.md
stefanv: We do need all these monitors for now, so will dismiss this comment.
We can address the overlap causing churn later.

docs/developer/local.md
```diff
@@ -0,0 +1,80 @@
...
+$ npm run theme:book
+```
+
+## Preview your changes
+
+Open the port that is printed in the terminal for your theme server (usually, `https://localhost:3000`). The theme server will start serving the AST from the content as a website at that port.
+
+## Use a custom port
+
+By default, the theme server will use the same port as the content server for changes to the AST. If you'd like to use a custom port, you can do so like this:
```
choldgraf:
```suggestion
By default, the theme server will watch the default port for the content server (`:3100`) for changes to the AST. If you'd like to use a custom port, you can do so like this:
```

docs/developer/local.md
```diff
@@ -0,0 +1,80 @@
...
+myst start --headless --server-port 3111
+CONTENT_CDN_PORT=3111 npm run theme:book
+```
+
+To connect to a remote content server, set the `CONTENT_CDN` environment variable:
+
+```bash
+CONTENT_CDN=https://remote.example.com npm run theme:book
+CONTENT_CDN=https://remote.example.com npm run theme:article
+```
```
choldgraf:
````suggestion
```bash
CONTENT_CDN=https://remote.example.com CONTENT_CDN_PORT=3111 npm run theme:book
CONTENT_CDN=https://remote.example.com CONTENT_CDN_PORT=3111 npm run theme:article
```
````

choldgraf:
I updated this per the comments and then got nerd-sniped by the great mystmd release armageddon of November 18 2025 😅

I'll merge this now!

PR #682, opened 2025-11-12
Prevent banner / sidebar overlap by making primary sidebar sticky

### LINKED ISSUE #679: Banner overlaps with the top of the primary sidebar
When a banner is present, it pushes down the navigation bar so that it overlaps with the top of the primary sidebar, rather than pushing the primary sidebar down so it's visible:

Image

choldgraf:
Seeing if this fixes the banner image sidebar overlap issue

- fixes #679

### DIFF
--- a/themes/book/app/components/Banner.tsx
+++ b/themes/book/app/components/Banner.tsx
@@ -45,7 +45,7 @@ export function Banner({
       className={classNames(
         'myst-banner w-full bg-blue-50 dark:bg-blue-900/30 border-b border-blue-200 dark:border-blue-800',
         'px-4 py-3 sm:px-6 lg:px-8',
-        'sticky top-0 z-40',
+        'relative z-40',
         className,
       )}
     >


choldgraf:
OK _just_ making it sticky fixes the banner problem but results in _really_ weird behavior with the page title etc


https://github.com/user-attachments/assets/6c9059fc-97d4-4062-966d-484afce96f13

stefanv:
So, a few ways to handle this. We could try work around it with CSS (make things relative instead of absolute), or we could add a ThemeProvider in the Banner that lets the theme know whether the banner is showing or not, or ...

choldgraf:
What do you think would be the simplest and most maintainable solution? And, that would work just fine for static builds?

stefanv:
I think this also impacts `useSidebarHeight`, which is now no longer correctly calculated.

stefanv:
See #683

PR #683, opened 2025-11-13
🫳 Adjust PrimarySideBar downwards when banner is visible

### LINKED ISSUE #679: Banner overlaps with the top of the primary sidebar

stefanv:
Expose the banner state for other components to use as well.

Closes #679

TODO:

- [x] Fix flickering PrimarySidebar

### DIFF
--- a/themes/book/app/routes/$.tsx
+++ b/themes/book/app/routes/$.tsx
@@ -25,7 +25,7 @@ import {
   useSiteManifest,
   useThemeTop,
   ProjectProvider,
-  BannerStateProvider
+  BannerStateProvider,
 } from '@myst-theme/providers';
 import { ComputeOptionsProvider, ThebeLoaderAndServer } from '@myst-theme/jupyter';
 import { MadeWithMyst } from '@myst-theme/icons';


stefanv:
@agoose77 In case you get a chance to look this over tomorrow: an alternative would be to add the banner state into the theme context. Right now, the top calculation feels a bit hacky; would be better to have the theme aware of the banner and just provide the correct top.

But, as mentioned above and in our discussion on Discord, ideally we can just use CSS and do away with all this; it really doesn't feel necessary to calculate the positioning / size of the primary sidebar at all.

themes/book/app/components/Banner.tsx
```diff
@@ -13,24 +14,37 @@ export function Banner({ content, className }: { content: GenericParent; classNa
   const contentString = JSON.stringify(content);
   const bannerId = hashString(contentString);
 
-  // Start hidden, only show after checking localStorage on client
-  // This avoids flickering on initial load
-  const [isVisible, setIsVisible] = useState(false);
+  // // Start hidden, only show after checking localStorage on client
```
agoose77: What's the double comment here for?

themes/book/app/routes/$.tsx
```diff
@@ -94,18 +96,21 @@ function ArticlePageAndNavigationInternal({
...
         {projectParts?.banner && <Banner content={projectParts.banner.mdast} />}
       </TabStateProvider>
       <TopNav hideToc={hide_toc} hideSearch={hideSearch} />
-      <PrimaryNavigation
-        sidebarRef={toc}
-        hide_toc={hide_toc}
-        footer={<SidebarFooter content={projectParts?.primary_sidebar_footer?.mdast} />}
-        projectSlug={projectSlug}
-      />
+      {bannerState.visible != undefined && (
```
agoose77: See earlier point about `bannerState` being optional, although less relevant here because we're a derived theme.

I'm also not sure about this guard due to SSR — I'm pretty sure this needs to be removed.

themes/book/app/components/Banner.tsx
stefanv: That should go, thanks.

packages/site/src/components/Navigation/PrimarySidebar.tsx
```diff
@@ -103,15 +104,18 @@ export function useSidebarHeight<T extends HTMLElement = HTMLElement>(top = 0, i
   const toc = useRef<HTMLDivElement>(null);
   const transitionState = useNavigation().state;
   const wide = useIsWide();
+  const { bannerState } = useBannerState();
```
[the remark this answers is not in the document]
stefanv: `BannerState.height` is always defined and starts off as 0, so should be OK?

themes/book/app/routes/$.tsx
stefanv: No, the flickering was too bad, so I left it in.

This is in the main route, so I suspect it's OK?

stefanv:
Hrm, I still see flashing, so going to remove that guard.

PR #686, opened 2025-11-14
FIX: Whitespace for checkboxes

choldgraf:
Checkboxes seem broken in the myst theme. I believe that this fixes it. I also added a page with a bunch of syntax we could use as a reference to look at.

- closed https://github.com/jupyter-book/mystmd/issues/2290

Broken:

CleanShot 2025-11-13 at 20 33 53@2x

With this PR:

CleanShot 2025-11-13 at 20 42 21@2x


stefanv:
Looks like a bunch of unrelated commits snuck in.

PR #699, opened 2025-11-20
implement an embed endpoint

akhmerov:
This is more a feature proposal than an actual PR, I just thought that it's easier to discuss if there's a prototype (**AI disclosure**: implemented by codex).

I propose to add an `/embed` route to the book, and potentially the article theme. This would allow to make an iframe-friendly view of a single component, indexed by URL and ID (or a complete page).

To make this complete, I'd also add a small header linking to the source, providing page info, etc.

Here's an embed view of a random section from the docs:

image

### DIFF
--- a/packages/site/src/pages/Embed.tsx
+++ b/packages/site/src/pages/Embed.tsx
@@ -1,0 +1,297 @@
+import React, { useEffect, useMemo, useState } from 'react';
+import { MyST } from 'myst-to-react';
+import { selectMdastNodes, type GenericNode, type GenericParent, type References } from 'myst-common';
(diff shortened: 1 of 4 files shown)


akhmerov:
I'm sorry for not being clear. I have created this PR only as a proof of concept and didn't intend it to be merged or reviewed. I wanted to state this in the PR description, but I see now that I should have been clearer. I also agree that linking a fork would be more appropriate.

I did get some feedback on discord, and I'll follow up via creating an issue.

choldgraf:
No worries - thanks for giving this a shot and driving some discussion!

Issue #94, opened 2023-05-26
not enough space after an image


stevejpurves:
### Describe the bug

**context**
When I do add an image to a myst file either using the commonmark image link syntax or an image directive, the text following the image is too close for comfort.

**expectation**
I expected there to be a reasonable amount of whitespace after the image

**bug**
But instead thre is no padding

![image](https://github.com/executablebooks/myst-theme/assets/1473646/4ff67208-c905-444a-abf9-4b4007ebd416)

**problem**
This is a problem for people doing publishing blog articles because that don't want to use the figure directive.

### Reproduce the bug

insert and image using `![]()` or the `image` directive immedaitely followed by text

PR #710, opened 2025-11-24
Fix inline images

### LINKED ISSUE #709: Images that are given in-line still have line-breaks
If an author adds an in-line image reference, I'd expect that the image will be displayed in-line as well (perhaps with the height set to match the `REM` of text). Instead, it creates a line break before/after the image:

Image

choldgraf:
- Fixes #709 
- Adds an extra build step to the dev docs
- Fixes a bug where centered images (the default) would remove their bottom margin, fixes https://github.com/jupyter-book/myst-theme/issues/94

New inline images (we didn't have an example before so I added one):

CleanShot 2025-11-24 at 13 57 17@2x

### DIFF
--- a/.changeset/yellow-taxis-peel.md
+++ b/.changeset/yellow-taxis-peel.md
@@ -1,6 +1,6 @@
 ---
-"@myst-theme/styles": patch
-"@myst-theme/docs": patch
+'myst-to-react': patch
+'@myst-theme/styles': patch
 ---
 
 Fix inline images and centered figure captions


stefanv:
The PR captions says "Centers figure captions under the figure", but I don't see that in the screenshot. Presumably, captions are meant to remain left-aligned?

choldgraf:
I'm working on the screenshot! needed netlify to finish building

docs/developer/local.md
```diff
@@ -6,6 +6,17 @@ To interact with the themes in development mode (e.g. with live-reload of compon
...
+## First build a local version of the book theme
+
+Run this to build a local version of the book theme:
+
+```shell
+$ make build-book
+```
+
+This is a requirement of the way our static site is generated.
+It expects to find a built theme locally, and the `myst start --headless` command below will error if it isn't done at least once.
```
stefanv: If you build the theme, then it should not be necessary to run a separate content/theme server.

styles/figures.css
```diff
@@ -5,9 +5,13 @@ figure.fig-quote figcaption > p:before {
   content: '—';
   padding-right: 0.5em;
 }
-figure.fig-quote.pull-quote > blockquote { 
+figure.fig-quote.pull-quote > blockquote {
   font-size: 1.4rem;
 }
+/* Center captions for standard figures to match centered images */
+figure:not(.fig-quote) figcaption {
```
stefanv: Just to make sure I understand: is this CSS rule sufficient to capture all cases?

styles/typography.css
```diff
@@ -35,9 +35,19 @@
...
+  /* Keep inline images inside text flow from forcing line breaks */
+  .article p img,
+  .article table img,
+  .article dd img,
+  .prose p img,
+  .prose table img,
+  .prose dd img {
     margin: 0;
     display: inline-block;
+    vertical-align: text-bottom;
```
stefanv: Do we want `text-bottom` or `baseline`?

https://www.w3schools.com/cssref/tryit.php?filename=trycss_vertical-align

styles/typography.css
```diff
@@ -35,9 +35,19 @@
...
+    max-height: 1.2em;
+    width: auto;
```
stefanv: Required?

styles/typography.css
choldgraf: hmmm maybe I should just remove this and not be opinionated about it until we see what behavior people expect?

styles/figures.css
choldgraf: OK I figured out a more targeted rule to just capture the `figure` directive case, because I realized that we're using `<figure>` in a few other ways as well (like pull quotes)

package.json
```diff
@@ -12,7 +12,7 @@
     "postinstall": "patch-package --patch-dir patches",
     "compile": "turbo run compile",
     "build": "turbo run build",
-    "dev": "turbo run dev --parallel --filter='./packages/*'",
+    "dev": "turbo run dev --parallel --filter='./packages/*' --filter=@myst-theme/styles",
```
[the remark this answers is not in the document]
choldgraf: It turns out this wasn't necessary after all! I, an idiot, was running my theme server in the wrong folder because I'm using worktrees

choldgraf:
OK I figured out some improvements:

- We now have a more targeted rule for the figure centering stuff.
- We left-justify the text, but center it with `width: fit-content` so it should now center under the image, BUT also be left-justified.

docs/myst.yml
```diff
@@ -21,7 +21,7 @@ project:
     - file: reference/index.md
       children:
       - pattern: 'reference/!(index)*.md'
-    - file: developer/index.md
+    - title: Developer documentation
```
stefanv: Unintended change?

choldgraf: There is no `developer/index.md` so I saw an error! Do you want me to make another PR for this one too?

styles/figures.css
```diff
@@ -1,11 +1,17 @@
...
 }
 figure.fig-quote figcaption > p:before {
   content: '—';
   padding-right: 0.5em;
 }
-figure.fig-quote.pull-quote > blockquote { 
+figure.fig-quote.pull-quote > blockquote {
   font-size: 1.4rem;
 }
 figure.fig-code > div {
```
choldgraf:
```suggestion
/* fig-code are captions on code blocks */
figure.fig-code > div {
```

docs/myst.yml
stefanv: https://github.com/jupyter-book/myst-theme/pull/712

akhmerov:
I believe this PR fixes https://github.com/jupyter-book/myst-theme/issues/94 as well

stefanv (approved):

PR #714, opened 2025-11-25
Add SCSS compilation for custom user stylesheet

choldgraf:
This adds an SCSS compiler when the user provides a custom stylesheet that ends in `.scss` or `.sass`. It's meant as a proof-of-concept, but it was also so easy to implement that maybe we can just ship it?

The biggest change is the dependency on sass-dart. How do folks think about that?

- closes part of jupyter-book/mystmd#2495 (but doesn't implement SCSS for our internal CSS)

I'm not sure what our versioning rules are, but I like `effver` and so am assuming this is `patch` because it doesn't break anything and only adds a new behavior if a user provided `.scss` in their stylesheet

Example: https://deploy-preview-714--myst-theme.netlify.app/ui/#custom-styling

### DIFF
--- a/.changeset/forty-suits-study.md
+++ b/.changeset/forty-suits-study.md
@@ -1,0 +1,8 @@
+---
+"@myst-theme/site": patch
+"@myst-theme/article": patch
+"@myst-theme/book": patch
+"@myst-theme/docs": patch
+---
+
+Add SCSS compilation for custom user stylesheet


agoose77:
Thanks for the PR Chris!  

Can I learn more about what problem this is trying to solve? The choice of CSS preprocessor seems theme specific, and for MyST theme I'd prefer PostCSS from an extensibility and modularity perspective. 

At the same time, there are features in various preprocessors like imports that we can't faithfully support when the preprocessor is performed at theme time, rather than as a build plugin in the MyST Document Engine. So, my preference is towards adding preprocessor support via build plugins (like https://github.com/jupyter-book/book-theme-tailwind-quickstart, but with a build hook that doesn't yet exist)

stefanv:
CLI-side makes sense to me.

Could y'all explain what you have in mind with plugin / build hook support?

agoose77:
Right now, `mystmd` "hands off" the stylesheet to MyST theme through a `file` option that the theme exposes in its `template.yml`: https://github.com/jupyter-book/myst-theme/blob/bd194887a9c0fc66323d519b7f5fea6d66707d21/themes/book/template.yml#L70-L72

`mystmd` doesn't do anything special here, it just knows how to fulfill `file` options by copying them into the right spot in the site build. When extending what MyST can do, it's nice to think about how can build a general pattern rather than special case CSS. For example, plugins might want to (in future) build LaTeX style files.

### Ideas

I suspect @rowanc1 and I have different ideas for how this might work. 

The "stupidest" way to plumb in plugins is to add some kind of event that plugins can listen for, and at that time they generate their own `style.css`. It's _this_ file that is defined in `myst.yml#site.options`. 

A smarter way might be to build out a way to have "dynamic frontmatter" like Python has for pyproject.toml (https://peps.python.org/pep-0808/, https://peps.python.org/pep-0621/#dynamic). This would be less "oh, we ran a plugin and now there's a file in the right spot", and more "we ran a plugin and it modified the project metadata".

I'm sure there are more approaches. I'm also :+1: in favour of allowing themes to declare plugins that they expect the user to have installed. This would allow us to bring functionality with a theme whilst respecting the theme-engine separation.

stefanv:
> When extending what MyST can do, it's nice to think about how can build a general pattern rather than special case CSS. For example, plugins might want to (in future) build LaTeX style files.

Or JS files, like for custom components? I like the idea of being able to produce arbitrary static files and hook them into the right places.

PR #719, opened 2025-11-25
Use `npm ci` instead of `npm install` for themes

stefanv:
xref: https://github.com/jupyter-book/mystmd/pull/2487

closes https://github.com/jupyter-book/mystmd/issues/2493

### DIFF
--- a/.changeset/plenty-paws-retire.md
+++ b/.changeset/plenty-paws-retire.md
@@ -1,0 +1,6 @@
+---
+"@myst-theme/article": patch
+"@myst-theme/book": patch
+---
+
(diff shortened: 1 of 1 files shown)


rowanc1:
Probably also without the scripts (if that does indeed actually work...!)

stefanv:
No, this was part of the test, just to see if it builds on CI in a clean env!

stevejpurves:
@stefanv can this come in independently of the linked `mystmd` PR and is it "ready"?

agoose77:
I believe it is. I'm merging!

PR #732, opened 2025-12-09
:mag: Add MyST search types to `providers` package deps

agoose77:
We export a type that uses these types, so it's a transient dep.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -43971,7 +43971,8 @@
       "version": "1.0.1",
       "license": "MIT",
       "dependencies": {
-        "@myst-theme/common": "^1.0.1"
+        "@myst-theme/common": "^1.0.1",
+        "@myst-theme/search": "^1.0.0"
       },
       "devDependencies": {},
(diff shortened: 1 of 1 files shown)


packages/providers/package.json
```diff
@@ -22,7 +22,8 @@
     "build": "npm-run-all -l clean -p build:esm"
   },
   "dependencies": {
-    "@myst-theme/common": "^1.0.0"
+    "@myst-theme/common": "^1.0.0",
+    "@myst-theme/search": "^1.0.0"
```
rowanc1: I think just `devDependencies`, as we are only doing types?

https://github.com/jupyter-book/myst-theme/blob/main/packages/providers/src/search.tsx#L2

I think in future moving search/xref/config types to spec or common would be ideal, as that brings the contract, but none of the code.

agoose77: These types are exposed in the signatures exported by this package, so they're production.

PR #733, opened 2025-12-09
🚛 migrate content for cross references

stevejpurves:
The purpose of this PR is to close off an edge / corner case introduced by the recent MyST AST upgrade around outputs. Specifically, when we have hover popovers activated on external cross-references, that causes a dynamic on-the-fly lookup of the external AST, which, of course, now may be a different version of the AST compared to the current version supported natively by the theme code at any point in time. 

It introduces a specific migration step for cross-references as well as establishing a place in a @myst-theme/common where we can announce the current MyST version that the theme will natively support. 

UPDATE: Due to the nature of local vs remote cross-references this also means that the themes will fully support AST migration via their loaders -- which is all good

Testing Video - V3 Theme Consuming a V2 AST and V2 external xref

https://github.com/user-attachments/assets/269e9117-c396-40d4-857b-b7a714be3319

UPDATE: Article theme also tested

### DIFF
--- a/themes/book/app/utils/loaders.server.ts
+++ b/themes/book/app/utils/loaders.server.ts
@@ -87,14 +87,10 @@ export async function getPage(
     throw redirect(projectName ? `/${projectName}/${newSlug}` : `/${newSlug}`);
   }
   let slug = opts.loadIndexPage || opts.slug == null ? project.index : opts.slug;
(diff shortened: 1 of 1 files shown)


stevejpurves:
merging after extended testing and convo with @agoose77

PR #735, opened 2025-12-10
Add a github release maker workflow

choldgraf:
This basically copies the pattern that we use in mystmd here:

https://github.com/jupyter-book/mystmd/blob/main/.github/workflows/publish-github-release.yml

It does the following:

- if changesets says we've published
- it grabs the tag of the current release using `package.json`
- it grabs the tag of the latest *github* release using the `gh` cli
- it generates a changelog using `github-activity` using `last-tag`
- it generates a github release using that changelog with `new-tag`

### DIFF
--- a/.github/workflows/publish-github-release.yml
+++ b/.github/workflows/publish-github-release.yml
@@ -1,0 +1,57 @@
+name: Publish GitHub Release
+
+on:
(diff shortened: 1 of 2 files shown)


stevejpurves (approved):
this looks good to me

PR #736, opened 2025-12-12
👮 non-fatal fallbacks for `output`

stevejpurves:
This adds fallback protection for malformed or unexpectedly shaped `output` nodes and avoids a fatal full page error when `outputs > output` heirarchy is not present (achieved by removing throw from `useOutputContext`).

image

Partly superceeds #726 for the case of output nodes.


stefanv (commented):
Thanks, Steve! 

It would be useful to know which case each if clause handles (with a one-liner comment). I can't check for completeness of all cases, but what's here looks good.

PR #739, opened 2025-12-12
Fix passing git tag to release action

choldgraf:
I accidentally deleted the most important part in my copypaste efforts. This PR makes the release action actually grab the myst-to-react version from package.json and turn that into a tag name for the github publish release action to use.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -72,7 +72,7 @@ jobs:
             echo "Unable to read version from packages/myst-to-react/package.json" >&2
             exit 1
           fi
-          TAG="myst-to-react@$VERSION"
+          TAG="myst-theme@$VERSION"
           echo "tag_name=$TAG" >> "$GITHUB_OUTPUT"
 
   # Trigger a github release workflow only if we've published


.github/workflows/release.yml
```diff
@@ -64,6 +64,17 @@ jobs:
...
 
+      - name: Determine tag from myst-to-react package.json
+        id: get_tag
+        run: |
+          VERSION=$(jq -r '.version' packages/myst-to-react/package.json)
+          if [ -z "$VERSION" ]; then
+            echo "Unable to read version from packages/myst-to-react/package.json" >&2
+            exit 1
+          fi
+          TAG="myst-to-react@$VERSION"
```
choldgraf: we are calling these tags / releases `myst-to-react` because we're getting the version from that package. if `myst-theme` is better we could do that? but this seemed simpler

.github/workflows/release.yml
```diff
@@ -64,6 +64,17 @@ jobs:
       - name: Checkout Repo
         uses: actions/checkout@v4
 
+      - name: Determine tag from myst-to-react package.json
+        id: get_tag
+        run: |
+          VERSION=$(jq -r '.version' packages/myst-to-react/package.json)
```
choldgraf: just checking my logic - this will be run *after* the changeset PR has been merged, which means that the version it finds will be the *new* version (which is what we want)

.github/workflows/release.yml
```diff
@@ -64,6 +64,17 @@ jobs:
...
+          if [ -z "$VERSION" ]; then
+            echo "Unable to read version from packages/myst-to-react/package.json" >&2
+            exit 1
+          fi
+          TAG="myst-theme@$VERSION"
```
choldgraf: after conversation in Discord switched this to myst-theme

so it reads the version from myst-to-react but then it calls the release myst-theme

PR #740, opened 2025-12-15
Make linked images display as block if they're the only content on a line

choldgraf:
- Fixes https://github.com/jupyter-book/jupyter-book/issues/2505

Here's a screenshot from `main`:

CleanShot 2025-12-15 at 14 38 22

Screenshot of this PR:

CleanShot 2025-12-16 at 10 43 43@2x


choldgraf:
I think that the latest commit will fix this, it just runs a quick check in the `paragraph` function to decide whether it's wrapping a single linked image, and if so, it does not wrap in `<p>` tags. I _think_ that's safe and I added a few examples to test out whether it is.

Issue #746, opened 2025-11-18
GitHub hover preview icon is not correct for merged PRs or issues closed as uncompleted


bsipocz:
The hover card previews for GitHub issues and pull requests are not correct in some cases.

### If the PR is closed without merging, the icon is a purple check instead of an X or red thing

In MyST:

<img width="1492" height="806" alt="Image" src="https://github.com/user-attachments/assets/255192ab-0de1-4e4c-9a88-aa9e55714d24" />

The issue:

<img width="1966" height="482" alt="Image" src="https://github.com/user-attachments/assets/9e482a0b-d756-4760-9141-cf3d5eca5cf5" />

### If the PR is merged, the icon is a purple check instead of a merge icon

In MyST:

<img width="1084" height="620" alt="Image" src="https://github.com/user-attachments/assets/f09f8e70-3ff3-409e-aea5-dbafbd4ca1e2" />

The issue:

<img width="1898" height="896" alt="Image" src="https://github.com/user-attachments/assets/64ba5112-b374-43ec-bead-dd2c4e65a8af" />

### If the issue is closed as `wontfix`, the icon is a purple check instead of a grey thing

In MyST:

<img width="1142" height="744" alt="Image" src="https://github.com/user-attachments/assets/43b9b692-d306-43f4-863c-272e028a0af9" />

The issue:

<img width="1922" height="868" alt="Image" src="https://github.com/user-attachments/assets/ddd55db6-3995-4951-b452-993b2384505d" />

PR #747, opened 2025-12-22
Add more preview icons for github issues and PRs

choldgraf:
This adds a few more icons for the hover previews with github issues and PRs. Because we have a bunch of icons now, it moves some of the logic for choosing the right icon above the part where we actually render it. The other changes are just minor naming updates so that it's easier to understand what the variables are doing. I added a little docs page so we can preview the outputs.

I had to make some arbitrary decisions about the icons to use, since we're using heroicons and some of them don't quite have the same icons that github uses, so I'm happy to switch to other icon choices if those are better.

Demo page:

https://deploy-preview-747--myst-theme.netlify.app/reference/github-previews/

Here's a video of what they look like:

https://github.com/user-attachments/assets/7f606060-8d54-40ad-b8b0-bd06c49f8d98

- closes jupyter-book/myst-theme#746.

### DIFF
--- a/.changeset/weak-cups-pay.md
+++ b/.changeset/weak-cups-pay.md
@@ -1,6 +1,5 @@
 ---
 "myst-to-react": minor
-"@myst-theme/docs": patch
 ---
 
 Add more preview icons for github issues and PRs


agoose77:
This looks really useful! I think we should create a test page for this in the docs, and create some test issues / PRs somewhere (probably a new repo, for ease) that we can use to visually spot check it. I can get to that next week / at the weekend.

choldgraf:
OK I've done these things:

- Added a test repository here: https://github.com/jupyter-book/test-repo
- Created several issues and PRs of the various states we want
- Updated our table to use those instead of the grab-bag we had before

Check out the table preview in the docs

PR #749, opened 2026-01-04
feat: add route to handle static assets

bryanwweber:
A new route `/static/*` is added to retrieve assets configured as static in myst.yml.

See jupyter-book/mystmd#1921

TODO:

- [ ] Just return the raw response from the content server, no need to check if the path is configured since there's now the `/static` prefix to differentiate. This might mean that the content server needs to have a handler to handle the mapping from path -> hashed filename rather than handling it here.
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/common/src/utils.ts
+++ b/packages/common/src/utils.ts
@@ -131,6 +131,9 @@ export function updateSiteManifestStaticLinksInplace(
     if (!action.static) return;
     action.url = updateUrl(action.url);
   });
+  data.static?.forEach((asset) => {
+    asset.url = updateUrl(asset.url);
+  });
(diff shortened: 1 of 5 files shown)


packages/common/src/utils.ts
```diff
@@ -131,6 +131,9 @@ export function updateSiteManifestStaticLinksInplace(
     if (!action.static) return;
     action.url = updateUrl(action.url);
   });
+  data.static?.forEach((asset) => {
+    asset.url = updateUrl(asset.url);
+  });
```
bryanwweber: question: unless Zed is mistaken, this function isn't called anywhere. Is this change needed?

Issue #362, opened 2024-04-16
Support for icon links


choldgraf:
It would be useful if MyST sites could link out to other websites via the theme's user interface. This would allow communities to define other external links that are relevant across the entire site.

It is common for links like this to be represented with small, standardized icons, and to provide users the ability to both choose the icon and the external link.

This often goes in either the navigation bar at the top, or the primary sidebar.

### Examples in the wild

The mystmd.org website implements its own custom icon links widget in the top-right (these are hidden on mobile)

![image](https://github.com/executablebooks/myst-theme/assets/1839645/0d9fd429-1613-4d95-866d-ce2e21054b71)

The Sphinx Book Theme implements icon links in the left sidebar (these are visible on mobile in the left drawer)

![image](https://github.com/executablebooks/myst-theme/assets/1839645/73714657-06dd-4f9f-acac-f6cd50120369)

The PyData Sphinx Theme implements icon links in the navigation bar (these are visible on mobile in the left drawer)

![CleanShot 2024-04-16 at 11 32 29@2x](https://github.com/executablebooks/myst-theme/assets/1839645/56fd8f51-7ad9-466d-9600-15ac4113f983)

PR #752, opened 2026-01-07
Add scienceicon renderers

brianhawthorne:
Enable CSF's `{scienceicon}` role (https://github.com/continuous-foundation/scienceicons?tab=readme-ov-file#myst-markdown) to render to React.

This is primarily in service of the **tools.scientific-python.org** site conversion to myst (https://github.com/scientific-python/tools.scientific-python.org/pull/81).

It is also a move in the direction of solving this standing `myst-theme` issue: https://github.com/jupyter-book/myst-theme/issues/362

---

- closes https://github.com/jupyter-book/myst-theme/issues/803

### DIFF
--- a/packages/myst-to-react/src/scienceicon.tsx
+++ b/packages/myst-to-react/src/scienceicon.tsx
@@ -1,0 +1,23 @@
+/**
+ * A small wrapper component around scienceicons so that we can add a .scienceicon class to style.
+ */
+import type { NodeRenderer } from '@myst-theme/providers';
+import { ScienceIconRenderer } from '@scienceicons/myst/react';
+import classNames from 'classnames';
+
+// Wrap upstream renderer output with a class so our theme CSS can target it
+// If scienceicons starts shipping with a class then we can delete this!
+export const ScienceIconWithClass: NodeRenderer = ({ node, className }) => {
+  return (
+    <span className={classNames('scienceicon', className)}>
+      <ScienceIconRenderer node={node} />
+    </span>
+  );
+};
+
+const SCIENCEICON_CLASS_RENDERERS = {
+  // Override the "scienceicon" node renderer key with our class-attached wrapper.
+  scienceicon: ScienceIconWithClass,
+};
+
+export default SCIENCEICON_CLASS_RENDERERS;
--- a/docs/_site/primary_sidebar_footer.md
+++ b/docs/_site/primary_sidebar_footer.md
@@ -1,1 +1,1 @@
-**Custom Footer** | [GitHub](https://github.com/jupyter-book/myst-theme)
+[{scienceicon}`github`](https://github.com/jupyter-book/myst-theme) [{scienceicon}`website`](https://mystmd.org) [{scienceicon}`mastodon`](https://bsky.app/profile/mystmd.org)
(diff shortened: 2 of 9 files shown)


choldgraf:
I think it'd be helpful if we let users easily define scienceicons in the book/article themes 👍 . Based on this implementation, it seems like we don't actually add any user-facing functionality, right? They'd still need to follow these instructions on their own from the science icons docs:

```
Copy the file scienceicons.mjs from dist/ intoyour local mystmd project and add it to the list of plugins in myst.yml:

project:
  plugins:
    - scienceicons.mjs
```

I also suspect that if they did this, and then wrapped the icons in a link (which seems to me the most common use-case), that they'd inherit the blue link syntax we use for prose. That feels suboptimal as well - we'd probably want "theme styling" for scienceicons when they're used.

Could we enable more "out of the box" functionality here by either:

1. bringing this plugin along with the theme, and defining some theme-specific CSS to make sure they look nice (standalone, as links, etc), OR
2. documenting how a user is expected to display their own scienceicons by loading the plugin themself (and confirming this works in our docs)

@rowanc1 what's your vision for how we should bring in functionality for scienceicons? Should it be all baked into this theme and out of the box? (e.g. so that anybody that uses these themes gets `{scienceicon}` roles for free?) Should it instead be user-enabled via plugin config, and this theme simply knows how to render the icons if they do this?

choldgraf:
@agoose77 what do you think is the best path forward for this PR? We could add it in as a step forward for now? My interest in this is that we'd be a step closer to https://github.com/jupyter-book/myst-theme/issues/362 without adding any new config surface area (instead, users would need to add a plugin).

To me the ideal use-case would be that a plugin can define both the roles/directives/node types that get put in the AST, *and* rules for rendering those nodes in various contexts. That's a step closer to our goal of mystmd/themes having a less opinionated base setup, and relying more on extension points for extra functionality. But it feels like we aren't there yet 

Maybe the MIMEbundle conversation is related to this? If a new node had a MIMEbundle-like thing inside, then a renderer could step through that bundle to decide how to render it even if it hadn't been explicitly told how to do it. If you combine that with the ability to ship things like PNGs with plugins then you could probably get pretty far.

packages/myst-to-react/src/scienceicon.tsx
```diff
@@ -0,0 +1,23 @@
...
+ */
+import type { NodeRenderer } from '@myst-theme/providers';
+import { ScienceIconRenderer } from '@scienceicons/myst/react';
+import classNames from 'classnames';
+
+// Wrap upstream renderer output with a class so our theme CSS can target it
+// If scienceicons starts shipping with a class then we can delete this!
+export const ScienceIconWithClass: NodeRenderer = ({ node, className }) => {
+  return (
+    <span className={classNames('scienceicon', className)}>
```
brianhawthorne: Might be nicer if the underlying element came already equipped with `.scienceicon` CSS class instead of adding another wrapper `<span>` around it.

Not sure if we have purview to modify that library... but that would be here &rarr; https://github.com/continuous-foundation/scienceicons/blob/59ae001228b3da486499668eb411be55c1ecd690/myst/src/react/scienceicons.tsx#L25

Issue #734, opened 2025-12-10
🔗 Internal footer links open in a new tab


brianhawthorne:
Footer links are rendered with `SimpleLink` (from `myst-to-react` package), which expects the given link node to have a boolean `internal` property, but the footer link nodes have no `internal` property and so render as external by default, causing the anchor tag to have `target="_blank"` attribute.

rowanc1:
Do you mean these [FooterLinks](https://github.com/jupyter-book/myst-theme/blob/main/packages/site/src/components/FooterLinksBlock.tsx#L6):

<img width="744" height="120" alt="Image" src="https://github.com/user-attachments/assets/399f4ab2-3a8d-44ae-88e7-121beb748c95" />

Or links in a footer block?

PR #753, opened 2026-01-09
Fix internal site footer links and standardize logic for internal vs. external URLs

brianhawthorne:
Ensure that internal links in the site footer do not open in a new tab.

Closes https://github.com/jupyter-book/myst-theme/issues/734

### DIFF
--- a/docs/developer/testing.md
+++ b/docs/developer/testing.md
@@ -31,3 +31,21 @@ Tests are located in one of two places, depending on the package:
 2. **Separate test directory** (used by `@myst-theme/jupyter`):
    - Create a `test/` directory in the package root
    - Example: `packages/jupyter/test/thebeOptions.spec.ts`
+
+## Add a test
+
+The root `npm run test` uses `turbo` to run tests.
+It will run tests in one of the repository packages if you define a `test` script in its `package.json`.
+
+For example, see the `myst-to-react` configuration:
+
+```{code-block} json
+:filename: package.json
+{
+  "name": "myst-to-react",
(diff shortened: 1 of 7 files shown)


choldgraf:
I think there are a few other places where we check for external links (eg the nav bar) mind if I build on this PR and push a commit that tries to standardize all of these on one helper function? Can also add a little test while we're at it

brianhawthorne:
> I think there are a few other places where we check for external links (eg the nav bar) mind if I build on this PR and push a commit that tries to standardize all of these on one helper function? Can also add a little test while we're at it

Please feel free!  Note, I cribbed this `isInternalUrl` function straight from `myst-cli`: https://github.com/jupyter-book/mystmd/blob/main/packages/myst-cli/src/build/site/manifest.ts#L222-224

One small concern I was hoping for feedback on here is that SimpleLink appears to expect the given mdast node to have a `.internal` property.  I wasn't able to figure out where (if anywhere) upstream that would be set.  So possibly some other code is responsible for setting that property, but broken, in which case the new logic here would sort of paper over another bug.. ¯\_(ツ)_/¯

choldgraf:
OK I just pushed a little commit here where you can see a few other spaces where we are now re-using the same function:

https://github.com/jupyter-book/myst-theme/pull/753/changes/0f4a9b4be3fd065c39d22109035d54bb0cd39912

I also added a test, and also added docs about how to add tests 😅

lemme know if that seems reasonable!

re: the `.internal` thing, I am equally confused :-)

PR #757, opened 2026-01-16
Standardize link styles

### LINKED ISSUE #59: Provide a visual cue when something has a "preview" for hover/click

### LINKED ISSUE #689: Standardize and improve the external and internal reference link colors and UI
Currently there are several different kinds of behavior users experience depending on the link that is shared. This can be confusing to users, because they feel quite different in style even though they are similar in function ("reference a thing"). There are two goals to this issue:

- Is there general agreement that this should be improved?
- What does the ideal UI/UX look like?

For example, it might make sense to differentiate "cross references" vs. "links", but not so drastically.

### Examples of link syntax

[Link to sandbox demo](https://mystmd.org/sandbox?tab=demo&myst=LQAtAC0ACgBhAGIAYgByAGUAdgBpAGEAdABpAG8AbgBzADoACgAgACAAdABlAHMAdAA6ACAAVABlAHMAdABpAG4AZwAhAAoALQAtAC0ACgAoAHQAZQBzAHQAKQA9AAoAIwAjACMAIABIAGUAYQBkAGUAcgAKAAoASABpAAoACgA6ADoAOgB7AGcAbABvAHMAcwBhAHIAeQB9AAoARABlAGYAaQBuAGkAdABpAG8AbgAKADoAIABEAGUAZgBpAG4AaQB0AGkAbwBuACAAbABpAHMAdAAKADoAOgA6AAoACgBIAGkAIABbAF0AKAAjAHQAZQBzAHQAKQAuAAoACgBFAHgAdABlAHIAbgBhAGwAOgAgAFsAZwBvAG8AZwBsAGUALgBjAG8AbQBdACgAaAB0AHQAcABzADoALwAvAGcAbwBvAGcAbABlAC4AYwBvAG0AKQAuAAoACgBEAGUAZgBpAG4AaQB0AGkAbwBuADoAIAB7AHQAZQByAG0AfQBgAEQAZQBmAGkAbgBpAHQAaQBvAG4AYAAuAAoACgBBAGIAYgByAGUAdgBpAGEAdABpAG8AbgA6ACAAdABlAHMAdAAKAAoACgAKAA%3D%3D)

Image

# Proposal

- All links: have an underline and are blue
- Links w/ hover preview: That underline is dotted.
- Non-links w/ hover preview: Also have a dotted underline of same style, but are normal text color (e.g. terms).
- External links: There's a small "external" link button to the right.

### Rationale

Here's my rationale:

- We have two axes: internal vs. external, and 'something happens on hover' or not.
- We want a consistent visual market for each axis.
- We don't want to use just color to distinguish this because that's not a11y friendly.
- Underlines for links and dashed underlines for "hover on this" are already common-place. We already use this for MyST "terms" for example.
- External link badges are also common-place. If authors didn't like it they could easily remove it with some custom CSS.

### Example

Something like this:

Image

choldgraf:
This standardizes our link styles so that they behave consistently. See the issue below for some discussion and the proposal this implements:

- https://github.com/jupyter-book/myst-theme/issues/689

Major changes:

- All links are underlined and blue
- External links have a little "external" button to the right
- Hoverable links have a _dashed_ underline
- Non-link hoverables (terms) have a _dashed_ underline and black

Also cleaned up the local install instructions along the way

👉 [Preview here](https://deploy-preview-757--myst-theme.netlify.app/reference/typography/#links)

Example of how this looks in netlify:

CleanShot 2026-01-18 at 13 31 15

---

- closes https://github.com/jupyter-book/mystmd/issues/59
- closes https://github.com/jupyter-book/myst-theme/issues/689

### DIFF
--- a/styles/hover.css
+++ b/styles/hover.css
@@ -41,5 +41,5 @@
 
 /* Non-link hover elements (e.g., abbreviations): gray dotted underline */
 .hover-text {
-  @apply underline decoration-dotted decoration-[0.15em] underline-offset-[0.15em] decoration-gray-500 dark:decoration-gray-400 cursor-help;
+  @apply underline decoration-dotted decoration-[0.1em] underline-offset-[0.15em] decoration-gray-500 dark:decoration-gray-400 cursor-help;
 }
--- a/styles/hover.css
+++ b/styles/hover.css
@@ -36,6 +36,10 @@
   @apply text-sm bg-white border rounded shadow-xl dark:bg-slate-800 border-gray-50;
 }
 
-.hover-link {
-  @apply font-normal text-blue-700 underline-offset-2 decoration-slate-200 dark:decoration-slate-600 dark:text-blue-100 hover:text-blue-500;
+/* Hoverable links: dashed underline */
+.hover-link { @apply decoration-dashed; }
+
+/* Non-link hover elements (e.g., abbreviations): gray dashed underline */
+.hover-text {
+  @apply underline decoration-dashed decoration-[1.5px] underline-offset-2 decoration-gray-400 dark:decoration-gray-500 cursor-help;
 }
--- a/styles/hover.css
+++ b/styles/hover.css
@@ -36,10 +36,10 @@
   @apply text-sm bg-white border rounded shadow-xl dark:bg-slate-800 border-gray-50;
 }
 
-/* Hoverable links: dashed underline */
-.hover-link { @apply decoration-dashed; }
+/* Hoverable links: dotted underline */
+.hover-link { @apply decoration-dotted; }
 
-/* Non-link hover elements (e.g., abbreviations): gray dashed underline */
+/* Non-link hover elements (e.g., abbreviations): gray dotted underline */
 .hover-text {
-  @apply underline decoration-dashed decoration-[1.5px] underline-offset-2 decoration-gray-400 dark:decoration-gray-500 cursor-help;
+  @apply underline decoration-dotted decoration-[0.15em] underline-offset-[0.15em] decoration-gray-500 dark:decoration-gray-400 cursor-help;
 }
--- a/styles/hover.css
+++ b/styles/hover.css
@@ -1,45 +1,49 @@
-.hover-card-content {
-  animation-duration: 0.6s;
-  animation-timing-function: cubic-bezier(0.16, 1, 0.3, 1);
-  z-index: 10;
-}
-.hover-card-content[data-side='top'] {
-  animation-name: slideUp;
-}
-.hover-card-content[data-side='bottom'] {
-  animation-name: slideDown;
-}
-
-@keyframes slideUp {
-  from {
-    opacity: 0;
-    transform: translateY(10px);
+@layer components {
+  .hover-card-content {
+    animation-duration: 0.6s;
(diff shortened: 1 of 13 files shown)


choldgraf:
I made one more quick improvement here:

1. dotted instead of dashed - this seemed like a more common pattern
2. Relative units (em) instead of fixed pixels - also increased the thickness a little bit so it's more noticeable              
3. Solid colors so they are more visible in general

styles/hover.css
```diff
@@ -36,6 +36,10 @@
...
 }
 
-.hover-link {
-  @apply font-normal text-blue-700 underline-offset-2 decoration-slate-200 dark:decoration-slate-600 dark:text-blue-100 hover:text-blue-500;
+/* Hoverable links: dotted underline */
+.hover-link { @apply decoration-dotted; }
+
+/* Non-link hover elements (e.g., abbreviations): gray dotted underline */
+.hover-text {
+  @apply underline decoration-dotted decoration-[0.15em] underline-offset-[0.15em] decoration-gray-500 dark:decoration-gray-400 cursor-help;
```
choldgraf:
```suggestion
  @apply underline decoration-dotted decoration-[0.1em] underline-offset-[0.15em] decoration-gray-500 dark:decoration-gray-400 cursor-help;
```

choldgraf (commented):
the line felt a bit too wide so this shrinks it a bit

agoose77:
I really like the intention of this PR, and having the visual reference is really useful.

I added a commit to move styles (some pre-existing) to the components layer. This is preferred for "is-a" things vs "has-a"/"like-a" things. As such, we have distinct classes `link` and `hover-link` etc.

I don't think we can merge yet. There's an a11y regression in buttons: 
image

I think the solution here is for button styles to take precedence over links. In other words, the design language in this PR should not apply to buttons in my view. This will be a bit fiddly because presently button styling is solely via CSS, whereas I think we'll need to add some React logic to add _either_ `link` _or_ `button` styles. I'm out of time this morning to get to that unfortunately.

choldgraf:
buttons are happy now

CleanShot 2026-01-20 at 09 38 53@2x

bsipocz (approved):
The preview looks to be an improvement to me, so 👍 for merging this and iterate later if necessary.

~(I would note however that the netlify preview link doesn't seem to have any working hovers, is that expected and just the side effect of the preview itself? -- if it's the expected behaviour then go ahead with the merge, but if it should work then do hold of merging)~ it works now, so it was something on my end; or transient for the previous CI run.

choldgraf:
Plan from Colab cafe:

- Merge this in
- Open an issue describing the need to have less inheritance by components
- Link that issue in the source code

PR #759, opened 2026-01-18
Make table cells match table headers

choldgraf:
The table headers were slightly off from the table cells, and I think (?) this standardizes them.

First reported here:

https://discord.com/channels/1083088970059096114/1462535518997713226/1462535518997713226

Main (note the offset):

CleanShot 2026-01-18 at 13 18 37@2x

This PR:

CleanShot 2026-01-18 at 13 18 21@2x

### DIFF
--- a/.changeset/mean-hats-smell.md
+++ b/.changeset/mean-hats-smell.md
@@ -1,0 +1,5 @@
+---
+"@myst-theme/styles": patch
+---
+
+Make table cells match table headers


choldgraf:
oops my bad - I just removed the tableofcontents changes so it's only the table header change remaining. sorry about that! I branch off of the wrong base 😅

bsipocz (approved):
Clearly looks to be an improvement, thanks!

choldgraf:
ok I think the latest commit standardizes this for right-aligned tables as well, let's see how the docs look

choldgraf:
oh I think this is already fixed - I just realized that @bsipocz linked the myst guide and not this PR, but this PR fixes both cases

CleanShot 2026-01-21 at 06 39 53@2x

PR #767, opened 2026-01-22
Improve link creation for static files

choldgraf:
This adds logic to check whether a link is a static file, and if so we just generate a link directly rather than using InternalLink, because the link for static files already includes the `BASE_URL` in it. This was causing us to incorrectly generate this twice!

### Rationale

My thinking is that the issue is that this function adds the baseurl to the static assets folder:

https://github.com/jupyter-book/mystmd/blob/80b028f4b1908048af97189c9ba2b7c30d279215/packages/myst-cli/src/build/html/index.ts#L95-L98

and then our `InternalLink` component here is also running `with base url`:

https://github.com/jupyter-book/myst-theme/blob/31815080ce054780c715f3a19af5ca33dee6a2ef/packages/myst-to-react/src/links/index.tsx#L48-L55

so the base url was being appended twice, and leading to links like @rylo described here:

```
stat238.berkeley.edu/spring-2026/spring-2026/build/syllabus-f303471004b9e1c88fe6fb615d2d475a.pdf
```

---
- fixes https://github.com/jupyter-book/mystmd/issues/2670

### DIFF
--- a/.changeset/red-seahorses-do.md
+++ b/.changeset/red-seahorses-do.md
@@ -1,0 +1,5 @@
+---
+"@myst-theme/site": patch
+---
+
+Increase contrast of search bar on dark mode
--- a/packages/site/src/components/Navigation/Search.tsx
+++ b/packages/site/src/components/Navigation/Search.tsx
@@ -174,7 +174,7 @@ function SearchShortcut() {
   return (
     <div
       aria-hidden
-      className="myst-search-shortcut items-center hidden mx-1 font-mono text-sm text-gray-600 sm:flex gap-x-1"
+      className="myst-search-shortcut items-center hidden mx-1 font-mono text-sm text-gray-600 dark:text-gray-300 sm:flex gap-x-1"
     >
       <kbd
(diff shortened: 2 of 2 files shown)


packages/myst-to-react/src/links/index.tsx
```diff
@@ -127,14 +137,24 @@ export const RORLinkRenderer: NodeRenderer<TransformedLink> = ({ node, className
 };
 
 export const SimpleLink: NodeRenderer<TransformedLink> = ({ node, className }) => {
-  const internal = node.internal ?? !isExternalUrl(node.url);
-  if (internal) {
+  const linkType = classifyLinkNode(node);
+  if (linkType === 'static') {
+    // Static links (e.g., downloads) already include the baseurl so we render them directly
```
rowanc1: In this case, all links by the time they get here already have the baseUrl I think?

Before did this just fall through to the `<a>` below. Are you just wanting to not show the `external` icon and open it in the same tab?

PR #768, opened 2026-01-22
Increase contrast of search bar on dark mode

choldgraf:
I noticed our search bar has bad contrast, maybe because of the link defaults work we did?

On main:

CleanShot 2026-01-22 at 13 37 05@2x


choldgraf:
added one more rule to fix another item we missed...I'll merge this now!

PR #770, opened 2026-01-22
🔗 Static links for downloads

rowanc1:
Do not add baseUrl.

Fixes https://github.com/jupyter-book/mystmd/issues/2670

An alternative to #767


packages/myst-to-react/src/links/index.tsx
choldgraf: We should add a conditional for this one, in this case, so that the externallinkicon doesn't start showing up for download links etc

PR #778, opened 2026-01-23
Fix IntersectionObserver logic for highlighted sections

### LINKED ISSUE #777: Clicking a page contents header highlights the wrong header if a scroll-down was triggered
If the page has to scroll down in order to reach the page header that you've clicked, then the scroll over-compensates and highlights a different section than the one you clicked.

### Reproduce this

- Go to https://myst-theme.netlify.app/reference/structure/
- Click one of the lower headers
- Note how the scroll in the contents area goes way too far

Here's a quick video example:

Image

### Cause?

I think there's probably a bug in the IntersectionObserver code that handles this, if we do it here the same way the sphinx-book-theme did it. I think this because it only happens on scrolling events not when you click a heading that doesn't require a scroll.

choldgraf:
For scrolling / header highlighting, we are now keeping a list of the "intersection elements" each iteration and adding/removing from that list. This is because in the previous mode, we _overwrote_ our list of intersecting elements on each event, which means that we were throwing out old elements that were still on screen but hadn't triggered the event.

### Main

**Note how the right TOC overshoots the actual header we click by a lot**

Image

And slow scrolls jump all over the place:

CleanShot 2026-01-23 at 17 24 40

### This PR

**Now the proper header is there**
CleanShot 2026-01-23 at 09 35 38

---
- fixes #777 
- fixes https://github.com/jupyter-book/mystmd/issues/2202
- fixes https://github.com/jupyter-book/mystmd/issues/1877

### DIFF
--- a/docs/permalinks.md
+++ b/docs/permalinks.md
@@ -1,3 +1,7 @@
+---
+description: Test for hover cards
+thumbnail: _static/myst-logo-light.svg
+---
 # Permalinks and URLs
 
 Configuration options for controlling site navigation and URL structure.
--- a/docs/reference/blocks.md
+++ b/docs/reference/blocks.md
@@ -118,3 +118,75 @@ graph LR
 
 Here's an iframe with a title
 ```
+
+
+## Topics
+
+```{topic} Special Topic
+This is a topic block. It's used to highlight special content or create a visually distinct section.
+```
+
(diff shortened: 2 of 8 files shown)


bsipocz:
Maybe it also resolves https://github.com/jupyter-book/mystmd/issues/1877?

PR #781, opened 2026-01-27
🫓 change from npm to bun

stefanv:
This proof-of-concept was generated by LLM so we can take a look at what is required.

It surfaced quite a few instances of packages being missing from `package.json` while being imported. `npm` does some flattening of the namespace for a monorepo, looks like?

I fixed several TypeScript failures, but I suspect many of those were actually due to packages not found and types then becoming generic. I can undo those changes one by one to see how far we get, although perhaps OK to make the types explicit too.


stefanv:
@rowanc1 This would benefit from a careful lookover, but the mechanics are in place.

package.json
```diff
@@ -1,65 +0,0 @@
-{
```
rowanc1: This shouldn't be deleted.

rowanc1 (commented):
Some comments as I go through.

package.json
rowanc1: Sorry for my terse notes. :) Mostly to myself as I was working through it this morning!!

stefanv:
Dependencies are not correctly being bundled, atm.

stefanv:
@rowanc1 I couldn't get differential server dependencies working; how bad is it to just build same way as prod as in https://github.com/jupyter-book/myst-theme/pull/781/commits/fb2532660a55a6c10dccbd8aa00548b398091164?

choldgraf:
We discussed this in the collab cafe today and aim to merge this w/ a light review _after_ the next release, so we have a bit of time to test it out on `main` and work out the kinks with some time before the next release.

rowanc1:
Ok - I think this is good to go:

- install works
- build works
- themes work in dev mode
- hot reload works
- built themes work
- built themes are installed with npm, and still work
- release files match mystmd
- visual smoke test done on dev docs
- jupyter components seem to work

I think we should aim to get this in as soon as we can, then follow up with a react upgrade then let it sit on dev for a few weeks.

stefanv:
Thanks for the detailed review, @rowanc1. Any thoughts on https://github.com/jupyter-book/myst-theme/pull/781#issuecomment-4063930017 ?

themes/article/package.json
```diff
@@ -8,27 +8,34 @@
     "build:css": "tailwindcss -m -i ./styles/app.css -o app/styles/app.css",
     "dev:copy": "cp remix.config.dev.js remix.config.js",
     "prod:copy": "cp remix.config.prod.js remix.config.js",
-    "clean": "rimraf public/build build api",
-    "build:thebe": "node ../../bin/copy-thebe-assets.cjs ./public",
-    "prod:build": "npm run prod:copy && npm run build:thebe && npm run build:css && remix build",
+    "clean": "rm -rf public/build build api",
+    "build:thebe": "copy-thebe-assets ./public",
```
agoose77, line 12: Here we delete the copy function

rowanc1: It is shipped in `thebe-core`.

agoose77: Makes sense — it is apparently a binary provided by thebe-core.

choldgraf, line 12:
```suggestion
    // This is a binary provided by the thebe-core package
    "build:thebe": "copy-thebe-assets ./public",
```

rowanc1: I don't think we can have comments in json?

rowanc1:
In testing @stefanv — the remix bundler with same process as dev/prod seems to just be working well.

I suggest we try it for a bit and see. :)

agoose77:
Shall we merge?

themes/article/package.json
rowanc1: @agoose77 merged and everything went red. :) Going to mentally remove the `?` from that ...

Issue #482, opened 2024-09-30
Scrollbars appear inevitably


lokgar:
Hi, at the moment when I build a site with `book-theme` I inevitably get unnecessary scrollbars, even with empty page, in any browser window size:

![image](https://github.com/user-attachments/assets/253b317b-e7bd-485c-822e-08adc96ab4bf)
![image](https://github.com/user-attachments/assets/272cc00a-1d3c-4873-b2ef-c12220b1e828)

They are quite annoying and do not appear on the mystmd.org/guide.

PR #785, opened 2026-01-29
Use parent width instead of screen size for page components to fix overflow

### LINKED ISSUE #780: The horizontal scroll bar is always displayed
Even in large screens the bar is always displayed

Image

choldgraf:
I think that this might fix the scrollbar bug. Apparently it's only on windows! I was able to confirm this with a BrowserStack session.

The problem seems to be the following:

- For the "no CSS" dialog, we set `width: 100vw`
- On windows, scrollbars are always displayed, on mac they are only displayed on scroll
- This means that on windows, the scrollbar's existence "pushes" most of the content to the left slightly.
- But things with `width: 100vw` still have the width of the whole window.
- So those things slightly overflow everything else, and create extra space to the right that is the same width as the scrollbar.

I am pretty sure these are the elements that may be causing it. Please speak up if somebody thinks that setting one of these to `w-full` instead of `w-screen` would be a problem, but I don't _think_ it should change the behavior because this seems to be reserved for components w/ a parent or positioning that spans the whole screen.

### Before this PR (with scrollbar)

CleanShot 2026-01-29 at 12 27 03@2x

### With this PR (no scrollbar)

CleanShot 2026-01-29 at 12 26 36@2x

---

- fixes #780 
- fixes https://github.com/jupyter-book/myst-theme/issues/482


rowanc1:
Yeah, I think that is solid. Go for the merge. :)

PR #786, opened 2026-01-30
More aggressively hide external link icon

choldgraf:
This flips the logic of our icon-links (for external links and downloads): instead of adding new rules to hide it, we hide it by default and have a single rule that says "only under these conditions should you _show_ it". Let's see if this makes for more consistent behavior.

### Visual testing

- ✅ Images no longer have external link icons: https://deploy-preview-786--myst-theme.netlify.app/reference/images/
- ✅ Links only have them if they should: https://deploy-preview-786--myst-theme.netlify.app/reference/typography/#links-and-terms
- ✅ Buttons do not have them: https://deploy-preview-786--myst-theme.netlify.app/reference/typography/#buttons

IMO this is a better fix!

### DIFF
--- a/styles/button.css
+++ b/styles/button.css
@@ -10,9 +10,5 @@
     &:hover {
       @apply bg-blue-700;
     }
-    /* Hide external link icon on buttons */
-    & .link-icon {
-      @apply hidden;
-    }
   }
 }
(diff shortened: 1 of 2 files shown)


styles/button.css
```diff
@@ -10,9 +10,5 @@
     &:hover {
       @apply bg-blue-700;
     }
-    /* Hide external link icon on buttons */
```
choldgraf: removing this because now we're selectively _showing_ it, instead of selectively _hiding_ it, which means that hopefully we don't need these one-off rules

rowanc1 (commented):
Looks solid, a more conservative approach. Test docs look good.

Issue #789, opened 2026-02-03
Outline (right-hand TOC) flashes when multiple `col-page-right` blocks are close to each other


jedwin3210:
Hey, I noticed this bug while developing this textbook for a course at UC Berkeley:

Webpage: [https://data89.org/notes/discrete-models/](https://data89.org/notes/discrete-models/)

GitHub: [https://github.com/data-89-berkeley/notes/blob/main/content/Chapter_02/02_Discrete_Models.ipynb](https://github.com/data-89-berkeley/notes/blob/main/content/Chapter_02/02_Discrete_Models.ipynb)

(linked to chapter 2.2 which has the issue - scroll down to a code cell the right-hand side toc will start flashing; this wasn't originally the case with a few cells spread out)

When there are multiple elements using the col-page-right class on the same page (in our case, two or more code cells with #| class: col-page-close together), the right-hand outline (secondary sidebar / in-page TOC) flashes repeatedly instead of staying hidden while content occupies that space.

I think I pinpointed the issue:

`useMarginOccluder()` in `packages/site/src/components/DocumentOutline.tsx` sets outline visibility from an intersection observer on each margin element with several col-page-right blocks near the viewport edge, the intersecting set flips every frame and the outline flashes.

example of GIF problem:

![Image](https://github.com/user-attachments/assets/d818343f-b897-4bbc-b559-91fb9e9097e5)

PR #790, opened 2026-02-04
Only create new intersectionobserver arrays if we've changed

choldgraf:
I think I've figured out the bug behind #789 . For each intersection observer callback, we were creating a _new_ array of elements on screen, even if the elements on the screen hadn't actually changed. Creating a new array seemed to be re-triggering the same callback (since there were new elements to observe). This PR only returns a _new_ array if the list has actually changed.

I found this hard to test, so I recorded two gifs before and after this on a page that demonstrated the same content.

### Before this PR

Note these are docs I just auto-generated to test out this bug, I'm not adding them in the PR!

(it was a little hard to capture because of framerate issues, but I assure you it was flashing!)

CleanShot 2026-02-03 at 16 20 53

### After this PR

I tried to move around the screen to see if we ever got flashing and we did not

CleanShot 2026-02-03 at 16 22 46

---

- closes This section has more margin content to continue testing as you scroll.

### DIFF
--- a/packages/site/src/components/DocumentOutline.tsx
+++ b/packages/site/src/components/DocumentOutline.tsx
@@ -26,6 +26,7 @@ const onClient = typeof document !== 'undefined';
  * loops with IntersectionObserver and MutationObserver.
  */
 function arrayIfChanged<T>(prev: T[], next: T[]): T[] {
+  if (prev === next) return prev;
   if (prev.length !== next.length) return next;
   for (let i = 0; i < prev.length; i++) {
     if (prev[i] !== next[i]) return next;


packages/site/src/components/DocumentOutline.tsx
```diff
@@ -19,6 +19,20 @@ const SELECTOR = [1, 2, 3, 4].map((n) => `main h${n}`).join(', ');
...
 const onClient = typeof document !== 'undefined';
 
+/**
+ * Returns `next` if it differs from `prev`, otherwise returns `prev` unchanged.
+ * This prevents unnecessary React state updates (and re-renders) when the
+ * array contents haven't actually changed - important for breaking feedback
+ * loops with IntersectionObserver and MutationObserver.
+ */
+function arrayIfChanged<T>(prev: T[], next: T[]): T[] {
+  if (prev.length !== next.length) return next;
```
stefanv:
```suggestion
  if (prev === next) return prev;
  if (prev.length !== next.length) return next;
```

stefanv (commented):
LGTM! Left a potential syntactic improvement, but not necessary to merge.

PR #793, opened 2026-02-04
Fix widebar width on mobile

### LINKED ISSUE #791: [mobile] Sidebar pop-out grows and shrinks width based on open TOC subtrees
On mobile, the width of the sidebar will grow if you open a subtree that has a long title. Closing the sub-tree will shrink the sidebar when the closing is done. Here's a GIF of what I mean:

Image

### Expected behavior

- The width of the sidebar stays the same.
- Sub-tree items with long titles wrap to the next line when they hit the edge of the sidebar.
- Toggling sidebar items open/closed doesn't impact its width at all.

I'm not sure, but we should also double check that the sidebar width doesn't change based on the longest top-level TOC item either.

choldgraf:
This updates our sidebar width to be fixed width instead of minimum width. I _think_ this only affects mobile. Its goal is to prevent the sidebar growing and shrinking as you open dropdown trees in the TOC.

#### Before this PR

CleanShot 2026-02-04 at 13 19 37

### After this PR

CleanShot 2026-02-04 at 13 20 16

---

- closes #791

### DIFF
--- a/packages/site/src/components/Navigation/PrimarySidebar.tsx
+++ b/packages/site/src/components/Navigation/PrimarySidebar.tsx
@@ -169,7 +169,7 @@ export const PrimarySidebar = ({
         'myst-primary-sidebar',
         'fixed',
         `xl:${grid}`, // for example, xl:article-grid
-        'grid-gap xl:w-screen xl:pointer-events-none overflow-auto max-xl:w-[350px]',
+        'grid-gap xl:w-screen xl:pointer-events-none overflow-auto max-xl:w-[75vw] max-xl:max-w-[350px]',
         { 'lg:hidden': nav && hide_toc },
         { hidden: !open, 'z-30': open, 'z-10': !open },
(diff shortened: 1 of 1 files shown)


choldgraf:
cool - so we have a few options:

- Reduce the width in general (it seems most web dev recommends ~330px max, we are at 350 now)
- Give it a max width (e.g. no more than 70% of screen)
- Make it grow and shrink based on content

I think I'd prefer to _avoid_ the last option. My understanding is that it's generally seen as bad practice to have UI elements "jump around" especially in unexpected ways. It's not a hill I will die on but I'd prefer to try the other options and see if they feel ok

choldgraf:
@bsipocz check out the latest commit - that:

- Sets the sidebar to 75% of the screen width
- Caps it at 350px

bsipocz:
It's much better, and in fact looking good now. Thanks!

PR #794, opened 2026-02-04
Make sidebar modal take up full height

### LINKED ISSUE #792: [mobile] There's a gap between the pop-out sidebar and the navbar on mobile
Note the gap between the sidebar and the top of the screen where you can see a greyed out navbar above.

Image

### Expected behavior

I think either:

1. The sidebar takes up the whole vertical screen real estate on mobile pop-out.
1. The sidebar perfectly aligns with the bottom of the navbar

I think (1) is the standard but we should double check

### LINKED ISSUE #797: Scrolling the page on mobile results in the sidebar being offset
If you scroll down the page and then open the sidebar, it is offset from the bottom instead of always going from bottom to top. Here's an example of a few different kinds of scrolling behavior:

Image

### Expected behavior

- Scrolling down on the page and then opening the sidebar doesn't change anything about how the sidebar is displayed.
- The sidebar is always from bottom to top of the screen on mobile (at least if we get in https://github.com/jupyter-book/myst-theme/pull/794)

choldgraf:
This cleans up the UI around the sidebar modal. It removes the "gap" between the pop-out and the top of the screen on mobile, and it also ensures that the "banner" is behind the "grey box" that is just behind the modal that you click to dismiss it.

Todo

- [x] Double check that the CSS rules are correct, I think they are but just need to double check
- [x] Make sure we didn't introduce other unexpected behavior out of mobile etc

### Before this PR

Note how the sidebar is both not at the top of the screen (so you lose screen real estate) and also has an extra gap because it's incorrectly calculating the height because of the banner)

https://github.com/user-attachments/assets/7223f07c-613b-4030-9e64-61b597ff95e5

### After this PR

https://github.com/user-attachments/assets/f21ca78d-f6ce-4ac1-992c-93af61ec2e9e

---

- closes #792 
- closes #797

### DIFF
--- a/packages/providers/src/ui.tsx
+++ b/packages/providers/src/ui.tsx
@@ -10,13 +10,16 @@ type UiContextType = [UiState, (state: UiState) => void];
 
 export const UiContext = createContext<UiContextType | undefined>(undefined);
 
-// Create a provider for components to consume and subscribe to changes
+// Tracks viewport width and mobile nav open state.
 export function UiStateProvider({ children }: { children: React.ReactNode }) {
-  // Close the nav
   const wide = useMediaQuery('(min-width: 1280px)');
   const [state, setState] = useState<UiState>({ isNavOpen: false });
   useEffect(() => {
-    if (wide) setState({ ...state, isNavOpen: false, isWide: wide });
+    setState((prev) => ({
+      isWide: wide,
+      // Close the mobile nav when switching to wide viewport (e.g. if user goes full-screen)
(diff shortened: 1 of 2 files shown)


choldgraf:
I merged from main and I also added more comments throughout the logic here, because I found it hard to wrap my head around which components were being defined where, and which behavior we were triggering with the CSS classes. If the preview looks good then I think this is good to go :-) 

IMO it is nice and standard to have the sidebar modal take up all the vertical height, rather than to be pushed down by the navbar!

PR #795, opened 2026-02-05
✨ Add `anywidget` directive rendering and model logic

stevejpurves:
This PR implements the node renderer for any widget directive. It's implemented as a separate package -- this is going to be more complex than your average node renderer because it's going to have, once fully implemented, various model runtime code to manage state and complete the rest of the any widget interface contract. 

Right now, this is capable of **rendering** varied content using hosted JavaScript modules, passing data into them. What's not implemented yet is the `initialize` and the full model synchronization cycle. Will be in the future, PR. See https://github.com/jupyter-book/mystmd/issues/2603 for overall progress and discussions on extensions ot make on the `mystmd` side.

https://github.com/user-attachments/assets/3bafb614-3751-45bd-b085-42cb0a2d8eb6

### Testing
Test this PR using the content here: https://github.com/jupyter-book/example-widgets
Ensure you are running `mystmd` build from the `sp/any-widget` headlessly from that folder, along with this branch running `npm run theme:book`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,10 +14,10 @@ jobs:
         with:
           fetch-depth: 2
           submodules: recursive
-      - name: Use Node.js 18.x
+      - name: Use Node.js 20.x
         uses: actions/setup-node@v3
         with:
-          node-version: 18.x
+          node-version: 20.x
       - run: npm install
       - run: npm run lint:format
       - run: npm run lint
@@ -28,10 +28,10 @@ jobs:
         with:
           fetch-depth: 2
           submodules: recursive
-      - name: Use Node.js 18.x
+      - name: Use Node.js 20.x
         uses: actions/setup-node@v3
         with:
-          node-version: 18.x
+          node-version: 20.x
       - run: npm install
       - run: npm run build
       - run: npm run compile # check the types of the site and themes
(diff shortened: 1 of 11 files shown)


choldgraf:
I pushed a commit that added documentation about this feature in the myst-theme. It tries to keep it more "developer-focused" since we use these more as mini test docs, and as a reference.

As part of that, did a few things:

- Added a little anywidget local module. I showed off two examples: a local and a remote anywidget module.
- Showed how you can use a local CSS file to over-ride the widget output styles
- Showed a way to use the shadow-dom to inject a `<style>` block that won't be clobbered by react (so that a user could over-ride styles of a widget more easily than if we hard-coded styles into the element).
- As part of that, I realized that it's possible to do all kinds of funky stuff with the `document`, and so added a warning that this is not supported or recommended.
- I added a mini test for the core get/set/update logic that we can build on later, just so we have something there.

I think there's a *lot* more we can do with the local file setup, so thanks @agoose77 for setting that up. Here's what the output looks like for me (since this won't work in our demo build)

CleanShot 2026-02-24 at 15 17 21@2x

IMO this is pretty much good to go

### What I couldn't figure out

I am not sure how to document behavior with an actual jupyter kernel, but my feeling was that it's OK if we leave that to a future improvement for now, unless @stevejpurves really wants to get that in as well?

stevejpurves:
> I am not sure how to document behavior with an actual jupyter kernel, but my feeling was that it's OK if we leave that to a future improvement for now, unless @stevejpurves really wants to get that in as well

that is all future @choldgraf and the docs look great. I want to get this in asap and my last snag is that the local dev experience (via content server) is not working, whilst maybe the static site builds are( ?) if this.is working in the docs.

PR #802, opened 2026-02-11
✨ Add navbar_end part in navbar UI

### LINKED ISSUE #801: Extend the navbar via parts configuration (similar to the sidebar and footer)
We allow users to extend the primary sidebar, as well as the footer, by defining markdown as a "part" and teaching the theme to look for that part and add it to the relevant UI space accordingly. In order to allow users to add extensibility to the navbar (e.g., icons or badges) we could add follow the same pattern.

### Proposal

If the user configures a `navbar_right` part, it will be parsed as AST and inserted just after the theme light/dark button (the right-most part of the screen).

A user configures the part like so:

```yaml
site:
  parts:
    navbar_right: icon-links.md  # Or in-line markdown
```

And in the file:

```
# in icon-links.md
![](markdown-badge.svc) | {button}`CTA`
```

And that file will be parsed and inserted here:

Image

### Implementation

This would basically follow exactly what we do with footers and primary sidebar.

### Considerations

- This would let users extend different UI parts of their site via the same structure we've already started using for footers etc.
- Plugins that define new roles and directives (or PRs like #752 ) could be used in these spaces without needing new configuration etc.

choldgraf:
This adds support for a markdown "part" that is added to the end of the navbar. See the docs for an example of this in action with two `{button}` roles. If we merge #752 we could replace those with icon links!
                                                                           
- Adds a new `navbar_end` site part, following the same pattern as footer, banner, and primary_sidebar_footer.
- On narrow screens, the content moves into the sidebar menu just above the "horizontal line"
- On wide screens, it's on the far end of the navbar, after the theme button.
- I did a little normalizing for spacing between right-side navbar elements (some were padding on both sides, and I made them all pad on one side so it's consistent)
- I added docs noting that site-level parts parsing isn't supported in the article theme.

---

- closes #801 
- supercedes #787 
- big chunk of https://github.com/jupyter-book/myst-theme/issues/362


packages/site/src/components/Navigation/PrimarySidebar.tsx
```diff
@@ -204,6 +208,16 @@ export const PrimarySidebar = ({
...
             </nav>
           )}
+          {navbarEnd && (
+            <div
+              className={classNames(
+                'article myst-primary-sidebar-navbar-end xl:hidden p-2 my-1 flex flex-wrap gap-2 [&_p]:contents',
+                sidebarSectionInsetClass,
+              )}
+            >
+              <MyST ast={navbarEnd} />
```
rowanc1: I am not sure - but for coherence with the `footer` part, this could be passed in as a react node (doing the `<MyST />` a level higher.

packages/site/src/components/Navigation/TopNav.tsx
```diff
@@ -152,8 +162,17 @@ export function TopNav({ hideToc, hideSearch }: { hideToc?: boolean; hideSearch?
...
           <NavItems nav={nav} />
           <div className="flex-grow block"></div>
+          {/* Search bar */}
           {!hideSearch && <Search />}
-          <ThemeButton />
+          {/* Light/Dark theme button */}
+          <ThemeButton className="w-10 h-10 ml-3" />
+          {/* Custom part at end of navbar */}
+          {navbarEnd && (
+            <div className="article myst-navbar-end hidden xl:flex items-center ml-3 [&>*]:m-0">
```
parmentelat: why hidden ?

stefanv (approved):
Same question as @parmentelat about `hidden` and `xl:hidden`, otherwise LGTM, and a neat addition!

packages/site/src/components/Navigation/TopNav.tsx
choldgraf: It's hidden because we have to render the navbar items in two places:

- Once in the navbar for wide screens
- Once in the *sidebar drawer* for narrow/mobile screens

So here we hide it until we hit `xl` width, at which point it becomes `flex`

packages/site/src/components/Navigation/PrimarySidebar.tsx
```diff
@@ -204,6 +208,16 @@ export const PrimarySidebar = ({
               <SidebarNav nav={nav} />
             </nav>
           )}
+          {navbarEnd && (
+            <div
+              className={classNames(
+                'article myst-primary-sidebar-navbar-end xl:hidden p-2 my-1 flex flex-wrap gap-2 [&_p]:contents',
```
choldgraf: Here we *show* the navbar items up until we hit `xl` at which point it becomes hidden

packages/site/src/components/Navigation/Navigation.tsx
```diff
@@ -15,13 +16,15 @@ export const PrimaryNavigation = ({
...
   footer,
+  navbarEnd,
 }: {
   children?: React.ReactNode;
   projectSlug?: string;
   sidebarRef?: React.RefObject<HTMLDivElement>;
   hide_toc?: boolean;
   mobileOnly?: boolean;
   footer?: React.ReactNode;
+  navbarEnd?: GenericParent;
```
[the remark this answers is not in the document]
choldgraf: I can't tell if this is a suggestion to do something different or not - but it's a good point that footer already uses `ReactNode`....I can try to make this follow the same pattern and see if I can get the outcome we want. Is that what you'd like me to do?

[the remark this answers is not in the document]
stefanv: Sorry, I was musing to myself that the typing feels less than optimal here. But this requires a systematic review to address, nothing for you to take care of here!

choldgraf:
I asked in the chat and Stefan said his comment was non-blocking, so I'm gonna merge this since we have an approve so we can close this out, thanks for the review folks :-)

rowanc1:
Amazing. Nice work! 🚀

PR #812, opened 2026-02-25
Fix linked images inlined

choldgraf:
this document, and fixes, a bug where inline images that were in links didn't properly get inlined:

CleanShot 2026-02-24 at 18 50 29@2x

### DIFF
--- a/docs/reference/blocks.md
+++ b/docs/reference/blocks.md
@@ -58,6 +58,14 @@ Second card.
 
 ::::
 
+:::{card} Linked card with image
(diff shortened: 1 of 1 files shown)


choldgraf:
I think I just pushed a fix for this - let's see if it works in the demo build. The issue was that we had a strict CSS rule of `a > img` but the linked span was generating `a > span > img`. So this removed the strict `a > img` and just makes it `a img`. I think that's safe - let's see how the docs look

PR #816, opened 2026-02-26
Support single domain to treat as internal URL

choldgraf:
This adds the ability for a user to define internal domains for links that will trigger "internal-link" behavior even if the URL is not to an internal myst document. For example.

```yaml
# In myst.yml
site:
  options:
    internal_domains: "mystmd.org"  
```

will make the theme treat a link like `https://mystmd.org/guide` as internal (open in same tab, no 'external link' icon).

---

- closes https://github.com/jupyter-book/mystmd/issues/2719#issuecomment-3968099854

### DIFF
--- a/docs/permalinks.md
+++ b/docs/permalinks.md
@@ -57,8 +57,8 @@ site:
 
 For example:
 
-- <https://mystmd.org> - treated as *internal* URL because the domain matches
-- <https://docs.mystmd.org> - treated as external URL because of the `docs.` subdomain
-- <https://jupyterbook.org> - treated as external URL because it doesn't match at all
+- `<https://mystmd.org>` - treated as *internal* URL because the domain matches
+- `<https://docs.mystmd.org>` - treated as external URL because of the `docs.` subdomain
+- `<https://jupyterbook.org>` - treated as external URL because it doesn't match at all
 
 You can match an exact domain (e.g. `mystmd.org`) or use a wildcard to match a single subdomain level. Matches will only be for that subdomain level, not deeper ones (e.g. `*.mystmd.org` matches `docs.mystmd.org` but not `a.b.mystmd.org`).


packages/site/src/components/Navigation/TableOfContentsItems.tsx
```diff
@@ -106,7 +111,9 @@ function LinkItem({
         <span className="inline align-middle">
           {`${heading.enumerator ? `${heading.enumerator} ` : ''}${heading.title}`}
         </span>
-        <ArrowTopRightOnSquareIcon className="inline h-4 w-4 align-middle ml-[0.2rem]" />
+        {!treatAsInternal && (
```
choldgraf: We need to add this extra conditional now because we might have an external URL in the TOC but that is now treated as internal w/ this new logic (before, this function would only be triggered when an external URL was listed in the toc)

docs/myst.yml
```diff
@@ -44,3 +44,4 @@ site:
     logo: _static/myst-logo-light.svg
     logo_dark: _static/myst-logo-dark.svg
     logo_alt: The logo for the MyST Markdown Project
+    internal_domains: "mystmd.org *.mystmd.org"
```
agoose77: I'd prefer not to introduce workaround syntax for the lack of array support in config. I think we should fix that first, and then update this PR to pull it in.

choldgraf: fiiiinnnneeee 😅

I was thinking the same thing haha - just didn't want to make this harder than it needs to be. But I do agree with you it's the "right" way to do it.

I am pretty sure that's as simple as making the template validation allow for `array` or `object` or whatever JS calls "lists of things" right?

choldgraf:
@agoose77 do you have any thought on the UX and the feature at a user-facing level? I want to make sure others think this would be valuable and that this is the right way to expose configuration/control over this functionality

docs/myst.yml
choldgraf: Does this describe the enhancement we'd need to get this in the right way?

- https://github.com/jupyter-book/mystmd/issues/2729

What do you suggest between:

1. Fix the issue, then come back to this PR
2. Restrict this PR to only allow a single string/domain. (I actually suspect this will cover 95% of the use-cases)

stefanv:
One test case to consider is [protocol-relative URLs](https://en.wikipedia.org/wiki/Wikipedia:Protocol-relative_URL) (starting with `//`).

And, rare, but multiple-level domains won't get caught: `isExternalUrl("https://mystmd.org.github.io", "mystmd.org")` returns false.

PR #830, opened 2026-03-05
👉 By default select the first project

rowanc1:


agoose77:
I think I can intuit what this does. It moves the "select the first project" logic to the consumer, not the provider, so that provider-free contexts work. Nice.

PR #841, opened 2026-03-17
Set netlify build branch from GitHub project

stefanv:
This PR adds the ability to adjust the branch of `mystmd` that is used to build a PR.
This is useful for testing `myst-theme` PRs that also modify `mystmd`.

It gets the branch information by querying a custom field on a [github project board](https://github.com/orgs/jupyter-book/projects/4).

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -5,10 +5,15 @@
   environment = { PYTHON_VERSION = "3.12" }
   publish = "docs/_build/html"
   command = """
-    # Determine which mystmd branch to build against.
+    # Determine which mystmd branch to build against. By default, use `main`.
+    # You specify a different branch in the left-hand column of the PR:
+    # under Projects, expand the fields and fill out `mystmd-branch`.
+    #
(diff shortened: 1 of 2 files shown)


stefanv:
Thanks for updating the Netlify token, @choldgraf—seems to work as intended now. Also moved the logic to Python, which leaves the Netlify config a bit cleaner.

choldgraf (approved):
This looks nice to me! I think we should merge and see how it works. I added a few quick comments that i think are safe to just merge in, the left right thing confused me a bit haha. I'll merge this so we can test it on your new PR

netlify.toml
```diff
@@ -5,13 +5,22 @@
   environment = { PYTHON_VERSION = "3.12" }
   publish = "docs/_build/html"
   command = """
-    # Clone and build mystmd from main branch
-    git clone --depth 1 https://github.com/jupyter-book/mystmd.git /tmp/mystmd && \
+    # Determine which mystmd branch to build against. By default, use `main`.
```
stefanv, line 8: Thanks for fixing; that was indeed a typo!

PR #842, opened 2026-03-22
🧑‍🧑‍🧒‍🧒 Add `myst#children` unidirectional anywidget model interface

agoose77:
This PR prototypes a reserved `myst#` interface for MyST-specific anywidget properties. In this first implementation there is support for passing in `children` from the `anywidget` node to the model, and subsequently reflecting changes to the model state into the AST.

See https://github.com/agoose77/example-anywidget-table/pull/1 for a demo of this:

image

The goal is to facilitate _portable_ node renderers such as gallery widgets that leverage MyST rendering but enhanced with runtime behaviour like filtering and sorting. It is already possible to do this by vendoring myst-to-react:

https://github.com/agoose77/example-anywidget-table

But this requires the anywidget to deal with the consequences of being isolated in its own React context and shadow DOM. By adding an interface for `children` rendering, we can ensure that the children renderer lives outside the shadow DOM (and make AST-tweaking widgets smaller and more robust). For example, here's a naive vendoring of `myst-to-react` alternative:

image

## To Do
- [x] Fix up types

### DIFF
--- a/packages/anywidget/src/renderers.tsx
+++ b/packages/anywidget/src/renderers.tsx
@@ -14,6 +14,9 @@ import * as React from 'react';
 import classNames from 'classnames';
 import type { AnyWidget } from './types.js';
 import { MystAnyModel } from './models.js';
+import { MyST } from 'myst-to-react';
+
+export const MYST_PROP_CHILDREN = 'myst#children';
 
 export function AnyWidgetRenderer({ node }: { node: AnyWidget }) {
   // basic validation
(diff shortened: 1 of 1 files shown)


packages/anywidget/src/renderers.tsx
```diff
@@ -14,6 +14,9 @@ import * as React from 'react';
 import classNames from 'classnames';
 import type { AnyWidget } from './types.js';
 import { MystAnyModel } from './models.js';
+import { MyST } from 'myst-to-react';
+
+export const MYST_PROP_CHILDREN = 'myst#children';
```
agoose77, line 20: This is a myst-reserved prop.

rowanc1:
Lots of rich discussion here:
https://discord.com/channels/1083088970059096114/1485410506490515618

The summary from the conversation is that an alternative approach using node-renderers would be useful to see before merging.

Issue #612, opened 2025-06-15
Hover colors are inconsistent for download, launch, copyright, jupyter buttons, etc


artoftheblue:
![Image](https://github.com/user-attachments/assets/bbdc6c87-35c0-464a-96b0-04e0e42a4713)

As shown in the gif above, the behavior of buttons is inconsistent:
* Download button doesn't change color on hover, although it's clickable;
* Copyright buttons become less opaque on hover;
* Jupyter Notebook mark doesn't change color on hover, although since it's non-interactable, it might be intended;
* Launch binder button changes color to a shade of orange on hover.

> [!NOTE]
> This PR is inspired by @stevejpurves's comment in #606:
>
> _We need to remove the hover color from the download button, this is a long standing library component and introducing a color here is imposing that on downstream theme builders. If we want that, we need to apply it in the theme/* and not packages/*._

I suggest either making the hover behaviors of these buttons similar, allowing users to customize these behaviors, or both. What are the conventions regarding this issue?

choldgraf:
Agreed we should have consistent behavior for buttons in the same area or class.

My first preference would be a black or white ish default (depending on dark mode) that changes to a color on hover that is configurable by the user in their theme.

Issue #227, opened 2023-08-25
Add option to choose accent color


eurunuela:
### Context

Currently, the themes use Tailwind's blue as the accent color with different tints and shades used throughout the themes.

### Proposal

To give users another customization option, the themes could have an `accent-color` option that would only accept Tailwind colors. The themes would then use this accent color, but wouldn't change the tint and shades to make it easier for devs to develop the feature.

### Tasks and updates

- [ ] Add `accent-color` option to config file.
- [ ] Read the user-given color in the different `jsx` files.
- [ ] Change all instances of `text-blue-600` and others to `text-{accentColor}-600`.

agoose77:
Thanks for opening this! Although this is the first issue, it's a specific example of a general problem that we later opened an issue, so I'm going to close it as a duplicate of that.

PR #843, opened 2026-03-23
Expose colors as css custom props to allow override

### LINKED ISSUE #567: Move colour choices to theme variables
It's not very easy to rebrand / reskin a MyST theme for different colour palettes. We might want to make this easier by exposing colours in a more centralised location.

pollomarzo:
Replaces ~200 "hardcoded" `blue-*`, `gray-*`, `slate-*`, `stone-*` Tailwind classes with **46 CSS custom properties** across semantic color roles, so users can theme the entire UI without rebuilding (or playing whack-a-mole with Tailwind classes).

- closes https://github.com/jupyter-book/myst-theme/issues/567
- Related to #227. 
- Touches #612, #689.

This PR aims for a fairly complete token system.

```jsx
className="text-myst-link hover:text-myst-link-hover"
```

To override:
```css
:root { --myst-color-link: #3a3988; }
.dark { --myst-color-link: #8b8ad0; }
```

Token values match original Tailwind defaults, so existing sites see no changes unless they override.

**Prose theming**: Removed `prose-stone dark:prose-invert` from `.article`: all 18 `--tw-prose-*` color variables now pass through `--myst-color-*` tokens in the DEFAULT typography config. Still using `prose` for spacing etc.

A new `/theming` page includes a live color picker widget (AnyWidget) that reads current CSS var values via `getComputedStyle` and applies overrides in real time. "Copy CSS" outputs `:root` or `.dark` blocks depending on current mode.

#### opacity modifiers

Tailwind v3 opacity modifiers (e.g., `bg-myst-bg/80`) don't work with CSS variables. A few translucent UI elements remain hardcoded:

- `bg-white/80` and `dark:bg-black/80` in `Headers.tsx` (banner backdrop-blur)
- `hover:bg-slate-300/30` in `PrimarySidebar.tsx` and `TableOfContentsItems.tsx` (translucent hover states)

These would require either RGB channel format tokens, `color-mix()`, or Tailwind v4 to migrate. Feedback?

#### action buttons
currently, there's no way to invert colors (ask for an element to be colored with dark scheme while on light scheme). action buttons and tooltip used this; i don't think tooltips should use it, but i see why action buttons did. the theme selector button falls into this category as well. maybe we can include duplicated variables under `.invert` or similar?

<details><summary>Complete token reference</summary>
<p>

**Accent tokens (14)**

| Token | Light | Dark | Role |
|---|---|---|---|
| `myst-primary` | blue-500 | blue-500 | Primary button bg |
| `myst-primary-hover` | blue-700 | blue-400 | Button hover |
| `myst-heading` | blue-900 | blue-100 | Accent headings (dt, TOC) |
| `myst-link` | blue-700 | blue-400 | Link text |
| `myst-link-hover` | blue-500 | blue-300 | Link hover |
| `myst-link-underline` | blue-400 | blue-500 | Underline decoration |
| `myst-active` | blue-600 | blue-400 | Selected UI elements |
| `myst-active-bg` | blue-300/30 | blue-500/20 | Active item bg overlay |
| `myst-active-surface` | blue-50 | slate-800 | Active item solid bg |
| `myst-focus-ring` | blue-500 | blue-400 | Focus ring |
| `myst-focus-outline` | blue-200 | blue-800 | Focus outline |
| `myst-accent-surface` | blue-50 | blue-900/30 | Banner/callout bg |
| `myst-accent-surface-border` | blue-200 | blue-800 | Banner/callout border |
| `myst-accent-surface-text` | blue-900 | blue-50 | Banner/callout text |

**Neutral tokens (11)**

| Token | Light | Dark | Role |
|---|---|---|---|
| `myst-bg` | white | stone-900 | Page background |
| `myst-bg-alt` | gray-50 | stone-800 | Sidebar, nav panels |
| `myst-surface` | gray-100 | slate-800 | Cards, dropdowns |
| `myst-surface-hover` | gray-200 | slate-700 | Hover on surfaces |
| `myst-text` | gray-900 | white | Primary text |
| `myst-text-secondary` | gray-600 | gray-300 | Body, labels |
| `myst-text-tertiary` | gray-500 | gray-400 | Captions, metadata |
| `myst-border` | gray-200 | gray-600 | Subtle dividers |
| `myst-border-strong` | gray-300 | gray-500 | Inputs, prominent dividers |
| `myst-code` | pink-700 | pink-400 | Inline code text color |
| `myst-kbd-shadow` | stone-900/10% | white/10% | Keyboard element shadow |

**Admonition tokens (21)** - 7 colors x 3 roles (border, header bg, header/icon text)

| Color | Border | Background | Text |
|---|---|---|---|
| Info | `myst-info` | `myst-info-bg` | `myst-info-text` |
| Success | `myst-success` | `myst-success-bg` | `myst-success-text` |
| Warning | `myst-warning` | `myst-warning-bg` | `myst-warning-text` |
| Danger | `myst-danger` | `myst-danger-bg` | `myst-danger-text` |
| Orange | `myst-orange` | `myst-orange-bg` | `myst-orange-text` |
| Purple | `myst-purple` | `myst-purple-bg` | `myst-purple-text` |
| Gray | `myst-gray` | `myst-gray-bg` | `myst-gray-text` |

</p>
</details>

<details><summary>Remaining hardcoded colors</summary>
<p>

- **GitHub icons**: `green-700`, `purple-700`, `red-700` for PR/issue state semantics
- **Error states**: `red-500` etc. for semantic errors (independent of brand)
- **Error/validation dialogs**: `text-white bg-red-500` in `InvalidBlock`/`UnknownBlock`, `border-red-500` in `Search.tsx` validation. Not sure if these should be themable
- **Spinner**: `gray-200`/`gray-600` track colors (neutral UI component)
- **LaunchButton**: `orange-500` jupyter orange
- **Opacity modifiers**: `bg-stone-200/10` in `code.tsx` (Tailwind v3 can't do opacity with CSS vars)

Verify with:
```bash
grep -rnE '\b(text|bg|border|ring|fill)-[a-z]+-[0-9]+(/[0-9]+)?\b|\b(text|bg|border|ring|fill)-(white|black)(/[0-9]+)?\b' packages --include="*.tsx" | grep -vE 'node_modules|/dist/|\.d\.ts|(text|bg|border|ring|fill)-myst-'
```

</p>
</details>


agoose77:
Thanks for opening this PR @pollomarzo! We've been keen to do this for a while, so it's great to see someone picking it up. Please ping us here or on the Discord if you get stuck!

choldgraf:
I would love to see this happen, the pydata theme had a lot of css variables that let users control the look and feel of the site and I think that's a great pattern to allow for authors to provide their own branding etc

Here's a link in case it helps for inspiration 

https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/styling.html#base-variables

docs/myst.yml
```diff
@@ -22,34 +22,48 @@ project:
...
-      url: /developer
+  - title: Reference
+    url: /reference
+  - title: Developer page
+    url: /developer
+    children:
+    - title: Local Development
+      url: /developer/local
+    - title: Architecture
+      url: /developer/architecture
```
choldgraf: Why did you add these extra dropdowns to the navbar? It feels a bit out of scope for this PR. If there's a good reason I think it'd be OK though

docs/myst.yml
```diff
@@ -22,34 +22,48 @@ project:
     primary_sidebar_footer: _site/primary_sidebar_footer.md
     navbar_end: _site/navbar_end.md
   toc:
-    - file: index.md
-    - file: ui.md
-    - file: authorship.md
-    - file: anywidgets.md
-    - file: permalinks.md
```
choldgraf, line 34: There's a bunch of formatting changes here that I think make it harder to review the PR. Can you revert anything that's not related to solving the one problem that this PR solves?

docs/src/color-picker.mjs
```diff
@@ -0,0 +1,142 @@
+/**
```
choldgraf, line 1: I'd recommend that we put this in a dedicated page like docs/color-picker and link to it from docs/theming (or have this as a child of the theming docs) so that we can have more user-facing documentation in docs/theming

docs/myst.yml
```diff
@@ -22,34 +22,48 @@ project:
...
+    url: /developer
+    children:
+    - title: Local Development
+      url: /developer/local
+    - title: Architecture
+      url: /developer/architecture
+  actions:
+  - title: GitHub
+    url: https://github.com/jupyter-book/myst-theme
+  - title: MyST Docs
```
choldgraf: For some reason I see duplicated buttons in the actions area:

Image

choldgraf (commented):
This is just a quick review, I think we should do a deeper look at the built docs and make sure that nothing is broken there. I'd suggest going through all the "reference docs" section and just visually look at as much as we can to see if anything changed.

In general, I really like the idea of using semantic class names, and CSS variables associated with those names, rather than hard-coding CSS via tailwind classes. I think it'll make it much easier for people to over-ride things and control the look and feel. So directionally I'm +1 on this.

There were some changes I didn't understand the rationale for in our docs that I flagged, and e.g. some formatting changes that we shouldn't have in this PR. We are already changing a *lot* of LOC here, and I think it'd be best to keep this as focused as we possibly can.

Am I correct that, right now, this would still leave some behavior hard-coded via the tw classes, and this is more like "move what we can into CSS variables, while recognizing some stuff remains to do this in the future"? I think I'd be fine with that, since right now the TW classes basically mean a user can't do _anything_ with the things they're applied to, and this is a step towards a bit more customizability. @agoose77 wdyt?

pollomarzo:
Thanks a lot for this first review. Glad to hear the approach seems sensible.

> There were some changes I didn't understand the rationale for in our docs that I flagged, and e.g. some formatting changes that we shouldn't have in this PR. We are already changing a _lot_ of LOC here, and I think it'd be best to keep this as focused as we possibly can.

I see what you mean. I should've mentioned this in my top-comment: all changes to the dev docs besides the "theming" page are components/pieces that were not shown in the docs, so I made changes to allow for visual inspection, fully expecting to revert after a review. Maybe I can squash changes into [actual PR] - [theming page] - [docs changes for visual inspection]? Or branch off in my fork, deploy and link here? Or I can absolutely just remove them.

> I'd suggest going through all the "reference docs" section and just visually look at as much as we can to see if anything changed.

While doing this, I created a small test to navigate to pages and capture rendered content so that I could compare before/after. I could contribute that, or post them somewhere. This misses hover states.

> Am I correct that, right now, this would still leave some behavior hard-coded via the tw classes, and this is more like "move what we can into CSS variables, while recognizing some stuff remains to do this in the future"?

Correct. For the opacity matter especially, to support opacity on colors we'd have to use one of the (sub-optimal) proposals I made in the top comment. For the  "action buttons invert in dark mode", I mostly wasn't sure whether it was intended in the first place; it looks cool, but it's a bit strange to only have it in dark mode, and it's the only element group that behaves this way (dark mode button + action buttons).

choldgraf:
I'd be happy to quickly review and merge the docs config / formatting stuff in a new PR if you want to quickly spin that off, assuming those are changes you think would improve our setup in general. Wanna do that so we can keep this PR tightly focused on the CSS variables?

choldgraf:
I spent a bit of time trying to get this PR cleaned up so that it is mergeable - there were a bunch of merge conflicts with main but I think they're all taken care of now in a way that didn't break anything. I also reverted some of the formatting changes that weren't functional, so we can keep the diff a bit cleaner.

@stefanv maybe this makes the PR a bit easier to review and think about? It seems reasonable to me in general, I think my intuition is to make sure there's nothig obviously wrong, and then just merge and try to iterate quickly as we spot problems. What do you think? (pinging you because I think you mentioned this one was important to SP!)

stefanv:
Awesome, thanks very much, Chris! This became a priority as of today, so I'll review tomorrow 😊

pollomarzo:
Sorry everyone for my absence. Meanwhile I found a couple missing spots (`hover-text` CSS rule, abbreviations). One pain point that I'm not sure how to handle: the `inverse`d components (the two buttons in the dev docs and the dark mode button). Should we have repeated, inversed rules for light/dark mode? Otherwise there's no way to say "color this component with the light mode colors" (since we override the variables in `.dark`)
Thanks a lot for the time you dedicated @choldgraf

choldgraf:
I updated a few more classes so that we use these new centrally-defined colors. Check those out in the latest commit and let me know if I got something wrong. I also noticed a few CSS classes that were mixing our new class names with opacity and I think (?) those were gonna break because of the opacity restriction.

For the opacity stuff - right now it's probably not critical to fix, and my vote would be we just get this to a mergeable state, and add that later if there's a key element that users want to over-ride that also uses transparency. I think my vote would be to try TW v4, and if that doesn't work, do `color-mix`

packages/frontmatter/src/downloads.tsx
```diff
@@ -143,16 +143,16 @@ export function DownloadsDropdown({ exports }: HasExports) {
         <ArrowDownTrayIcon
           width="1.25rem"
           height="1.25rem"
-          className="myst-fm-downloads-icon"
+          className="myst-fm-downloads-icon stroke-myst-text-tertiary hover:stroke-myst-text"
           aria-hidden="true"
           title="Download"
         />
       </Menu.Button>
-      <Menu.Items className="myst-fm-downloads-menu absolute z-10 overflow-hidden bg-white rounded-sm shadow-lg -right-1 dark:bg-slate-800 ring-1 ring-black ring-opacity-5 focus:outline-none">
```
stefanv, line 151: These colors just went away?

docs/theming.md
```diff
@@ -0,0 +1,36 @@
+---
+title: Theming
```
stefanv: Well, this is going to be problematic unless we change out concept labels :)

Maybe this is "Styling" for now?

docs/theming.md
```diff
@@ -0,0 +1,36 @@
...
+---
+
+`myst-theme` exposes its color palette as CSS custom properties (CSS variables), so you can re-skin the entire interface from a single stylesheet without rebuilding the theme.
```
stefanv, line 5: Yes, it also attached a CSS class to each element.

docs/theming.md
```diff
@@ -0,0 +1,36 @@
...
+
+To explore the tokens interactively, try the [Live color picker](./color-picker.md).
+
+## Applying overrides in your own site
+
+Add the declarations you want to override to a stylesheet loaded by your site:
```
stefanv: Point to where we show how to do that?

docs/theming.md
```diff
@@ -0,0 +1,36 @@
...
+
+```css
+:root {
+  --myst-color-info: #3b82f6;
+  --myst-color-info-bg: #eff6ff;
+  --myst-color-info-text: #2563eb;
+  /* ...override only what you need */
+}
+
+.dark {
```
stefanv, line 22: This is confusing: `:root` vs `.dark`. Shouldn't it be:

```
:root {
    .light {
       ...
    }
    .dark {
    }
}
```

Or some other similar nesting structure?

Maybe we don't parse SCSS (what?! surely we do). But, even if we don't, I expect there to be an equivalent light and dark definition.

docs/theming.md
```diff
@@ -0,0 +1,36 @@
...
+  --myst-color-info: #60a5fa;
+  --myst-color-info-bg: #0f172a;
+  --myst-color-info-text: #93c5fd;
+}
+```
+
+## Known limitations
+
+A few colors are intentionally not themable yet:
```
stefanv: What is listed below isn't a color.

docs/theming.md
```diff
@@ -0,0 +1,36 @@
...
+
+- **Opacity modifiers** (e.g. `bg-myst-bg/80`) don't work with CSS variables under Tailwind v3, so a few translucent surfaces remain hardcoded (the top navbar backdrop, some hover overlays).
+  Migrating these requires either RGB channel tokens, `color-mix()`, or Tailwind v4.
```
stefanv: I think we can do:

```
<p className="text-[var(--brand-color)]/80">Hello World</p>
```

If not, then we can still work around the issue by using something like:

```ts
const MyComponent = () => {
  const style = {
    color: 'rgb(from var(--myst-color) r g b / 0.8)'
  };

  return <p style={style}>Some text</p>;
};
```

docs/theming.md
```diff
@@ -0,0 +1,36 @@
...
+- **Inverted color schemes** — action buttons and the theme toggle previously rendered with inverted colors on hover.
+  There's currently no token for "use the opposite scheme's colors here", so these now follow the standard tokens.
```
stefanv: How about `invert`? https://v3.tailwindcss.com/docs/invert

docs/theming.md
```diff
@@ -0,0 +1,36 @@
...
+## Applying overrides in your own site
+
+Add the declarations you want to override to a stylesheet loaded by your site:
+
+```css
+:root {
+  --myst-color-info: #3b82f6;
+  --myst-color-info-bg: #eff6ff;
+  --myst-color-info-text: #2563eb;
+  /* ...override only what you need */
```
stefanv, line 19: Where can the user find the full list of color variables?

stefanv (commented):
1. This is great! Happy to see custom styling arrive.
2. I'm not going to be able to check each individual occurrence, but I presume an LLM has had a look.
3. Some comments around documentation / edge cases / usability.

docs/theming.md
choldgraf: we could do a `literalinclude` of the place where we define them maybe?

stefanv: Yes, or just link to that file.

docs/theming.md
stefanv: Invert does too much (takes over the entire element, including images etc.).

stefanv:
I've added a few more commits. How do we review this to ensure we didn't break anything?

choldgraf:
I think we'll need to just do a thorough look-through, and use the demo page for inspiration, and then also just merge it and see if anybody stuff breaks and be ready to fix things.

Usually this is where I just go page by page through the visual reference to see if anything looks off haha.

I'll try and do this tomorrow if somebody else doesn't beat me to it first! Maybe @pollomarzo has a moment to look?

stefanv:
Changes spotted on initial scan in dark mode:

- Button hover color has changed (now light, was dark)
- Background for hover box (over, e.g., a PR number) is now black (should be blue)
- Top of code blocks now are black (the part that says the languge, e.g. `yaml`); used to be gray
- Hover highlight on right-hand-side navigation bar is now working (was broken!)
- Footer colors have changed: background was dark blue, is now dark gray
- Admonitions are brighter (looks better, actually!)
- Pandas tables: alternating rows are now white (wrong)

stefanv:
I got most of the ones I identified; there are a few (the buttons, due to inverse) that cannot be fixed easily.

I left the brighter admonitions in place, but strictly we'd have to revert that to be perfectly backward compatible.

choldgraf:
Another thing we should do is make sure that the naming of variables actually is the right name (eg semantically makes sense given the things it changes.) I wonder if there is a standard nomenclature across CSS frameworks or something

pollomarzo:
I'd created a simple test that creates screenshots of all pages to allow for quick comparisons. I'll start with that, then check hover states on most components.
Thank you both for the time you've dedicated! Stefan, glad you like the few improvements and sorry about all other regressions :grin: 
I think now that you've looked into it you see that some decisions between dark and light mode were different so some choices were motivated by an attempt to merge them into a single path... plus a couple fixes I made along the way (and a few regressions). This makes me say that aiming for 100% pre/post is not the right target, but of course you are the maintainers so that's your decision.
I'll dedicate some time and look it over and list them first just in case I can't tackle all of them

pollomarzo:
Ended up taking me all the time i had. 
I marked all the differences I view as potentially problematic as `(!)`. Some are present both before and after. For testing I opened a split view in firefox, ran main and this with a shared content server and went through all pages switching light/dark mode.
I included open things from @stefanv so that we don't lose track of them

> Changes spotted on initial scan in dark mode:
>     * Button hover color has changed (now light, was dark)

- top nav, now highlights differently (blue font color)

>     * Pandas tables: alternating rows are now white (wrong)
- lighter alternation, both light and dark mode

- side nav darker background for current heading in dark mode
- top nav second level elements are now dark as background (used to be white)

### admonitions
- dropdowns are lighter
- toc (fine but different)
>     * Hover highlight on right-hand-side navigation bar is now working (was broken!)

    - sub-headings are darker
    - siderail is darker
- download box content is different color, both are readable


### blocks
- abstract has grey background in dark mode (was blue, looks good though)

- less syntax highlight in dark mode
- cards grey border (was blue)
- topic heading no different background (!)
- hover card remains dark mode (used to be light mode). but text (subtitle?) is black on grey (!)
- on ^, switching to dark mode gives issues with some logos (e.g. wikipedia logo is a black png!)
- hover card has surface background (didn't use to)

### computation

- binder box is same color as background (used to be white)
- copy button goes light->dark in light mode, dark->light in dark mode. looks good.
- no hover style on heading anchors (wasn't there before either)
- error outputs not readable in dark mode (white on pink) (!), sorry i don't know the state of the accessible colors

### typography
- math is cut off (!) both versions
- abbreviations were reverted to non css-vars state https://github.com/pollomarzo/myst-theme/blob/a5251c13733d1e105d3983a47555adcb89beee70/packages/myst-to-react/src/components/HoverPopover.tsx#L53 (!)
- footnotes and references are darker in dark mode
- links in references are blue, used to be black (i'd say it's... correct?) (!)

### developer
- underline of "current page" in top nav is thinner/less visible (!)
"developer tips"
- underline in popover is blue, used to be black in light mode
- header (for filename) in popover is same "surface color" as popover background (!)

also had claude make a tiny [tmux script](https://gist.github.com/pollomarzo/84a63ffb4f9e9cc6bc186a6beb5520fc) to set up the three servers after the third time i did it manually, if it's of any use to someone else...

choldgraf:
found another one - I think we are making the hover card backgrounds grey now
 
CleanShot 2026-06-11 at 11 02 35@2x

pollomarzo:
I think it's better to have a color for "surfaces" that isn't the same as the normal background color, and this also allows to keep dark/light more aligned (since there's no shadows in the dark version, you need a different background color to distinguish the two). As always your call though

choldgraf:
I think you might be right, @pollomarzo , but I don't want to conflate purposes of this PR. We should keep this PR to implementing variable naming, and only _changing_ things if it is a _very obvious_ thing to do (e.g. correcting a bug). For design decisions like "should the card background be white or grey" I think that's better in dedicated issues.

I'll action that and a few things I see and will push a few commits

stefanv:
I think we should focus on ensuring there are mechanisms for adjusting these things; then, minor changes can go either way—easy to fix by simply tweaking a variable!

stefanv:
(I should add that, it would be good for us to have a very clear record of what changed, with screenshots, so we can review as a team.)

choldgraf:
A few things that I noticed:

- myst-heading -> myst-accent, since the thing modified by this variable is the "active" nav item, not any headings on the page
- bg-alt -> bg-secondary, since we use "secondary" in several other variable names, this feels more consistent.
- I reverted the prose body color to be a bit "softer" again since its a lot of text, and on dark mode it was pure white which was a bit harsh. That brings this PR back in-line with the text color of prose on `main`
- There were a few other little UI things that I caught and updated (e.g. the background of the close button on the banner  was the same color as the banner so I made it slightly different etc)

I split it up into a bunch of commits for review - I don't feel particularly strong about them if anybody wants to revert something

choldgraf:
OK I found a few more little things to clean up, and some stuff that @stefanv told me he noticed in a visual review. I also gave it a visual review and didn't see anything that stood out to me.

Given that we've taken multiple passes at this, how about @stefanv merges this whenever he's confident there isn't anything _obvious_ wrong with it? IMO it is good to go, and we can always spot-check improvements in a follow-up PR that isn't as complex as this one!

choldgraf (approved):

choldgraf:
stefaaaaan https://www.youtube.com/watch?v=6b9ci_z4v7M&list=RD6b9ci_z4v7M&start_radio=1

pollomarzo:
@stefanv  is there something more you need in this PR before merge? Additional tests I can help with?

choldgraf:
I don't want this one to go stale, and we've already had multiple rounds of review on this, so I'm just gonna merge it in. I am happy to iterate with somebody if they spot a new problem and want to fix it in a PR. Thanks @pollomarzo for the help on this

PR #844, opened 2026-03-26
🐛 Fix netlify branch script for our theme docs

choldgraf:
We are exiting `1` when there's no netlify branch target, but I think we want to exit `(0)` instead because on `main` we don't expect that target to exist, and currently it's causing netlify build failures

### DIFF
--- a/.github/tools/get-mystmd-branch.py
+++ b/.github/tools/get-mystmd-branch.py
@@ -99,11 +99,11 @@ def main():
     branch = DEFAULT_BRANCH
 
     if review_id is None:
-        print("No REVIEW_ID set by Netlify; exiting.")
-        sys.exit(1)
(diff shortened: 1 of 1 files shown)


choldgraf:
I think the only way to test this is to merge it - so if the netlify script works on this PR, I'll merge and then see if it works on main

choldgraf:
ok I think that fixed it

stefanv:
This would work, but now we would fail to catch missing REVIEW_ID--maybe fine, since Netlify should always provide that. 

PULL_REQUEST is another an env var we can check to see whether we are building on a PR or not.

choldgraf:
oh I see you mean just add a check for `PULL_REQUEST == True` and if so, just skip all the following logic? That does seem simpler!

PR #845, opened 2026-03-30
Add an authorship example in the documentation

choldgraf:
This adds a simple example that defines a bunch of authorship metadata and adds a page so we can see what does and does not show up, and so we have something to iterate on if we want to extend or add functionality.


choldgraf:
I'm just gonna merge this because it's a little docs example.

PR #847, opened 2026-04-02
Document grid system without storybook

### LINKED ISSUE #829: Port the grid system storybook page to MyST Theme docs
In #827 we're discussing removing storybook because it's extra maintenance burden without a huge benefit. However, the [structural elements](https://jupyter-book.github.io/myst-theme/?path=/docs/components-grid-system--docs) page is quite helpful at navigating the grid and layout of a page, and how the CSS classes impact it.

We could try porting that page to our myst-theme docs, but just doing it in a systematic way via our own CSS classes to show off how users and trigger different layouts with various classes.

### References

- This [page from PST](https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/layout.html) could be used for inspiration

choldgraf:
This adds a `layout.md` page to our reference documentation that (I think) uses all of the CSS classes that we demo in the grid system on storybook. It basically just puts them all in divs so you can see how they look, with a little bit of styling to make them look nice.

[Demo here](https://deploy-preview-847--myst-theme.netlify.app/reference/layout/)

---

- closes #829

### DIFF
--- a/docs/reference/layout.md
+++ b/docs/reference/layout.md
@@ -1,0 +1,233 @@
+# Page Layout
+
+The myst-theme page layout is built on a CSS grid system with three variants.
+Each bar below uses the actual CSS class as its label.
+Each is generated with the `{div}` directive.
+
+We intentionally leave the primary and secondary sidebars visible so that you can see how they relate to one another in size, but normally you'd remove one or the other if you wanted to use many of these classes.
+
+## `article-grid`
+
+:::::{div}
+:class: col-page not-prose
+
+::::{div}
+:class: article-grid grid-gap
+
+:::{div}
+:class: layout-demo layout-demo-col-screen col-screen
+col-screen
+:::
+
+:::{div}
+:class: layout-demo layout-demo-col-page col-page
+col-page
+:::
+
+:::{div}
+:class: layout-demo layout-demo-col-page-inset col-page-inset
+col-page-inset
+:::
+
+:::{div}
+:class: layout-demo layout-demo-col-body-outset col-body-outset
+col-body-outset
+:::
+
+:::{div}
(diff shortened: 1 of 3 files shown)


docs/reference/layout.md
```diff
@@ -0,0 +1,233 @@
+# Page Layout
+
+The myst-theme page layout is built on a CSS grid system with three variants.
+Each bar below uses the actual CSS class as its label.
+Each is generated with the `{div}` directive.
+
+We intentionally leave the primary and secondary sidebars visible so that you can see how they relate to one another in size, but normally you'd remove one or the other if you wanted to use many of these classes.
+
+## `article-grid`
+
```
rowanc1, line 10: It is possible that we might need a:

```
+++ {"class": "col-screen"}
```

Here?

I am suspect that the screen does not extend to the edges, when it should ... 

Image

docs/reference/layout.md
```diff
@@ -0,0 +1,233 @@
...
+:::{div}
+:class: layout-demo layout-demo-col-gutter-page col-gutter-page-right
+col-gutter-page-right
+:::
+
+::::
+
+:::::
+
+## `article-left-grid`
```
rowanc1, line 106: I am not sure if this is going to work on the same page (or at all .. ) unless we wrap with a "article-left-grid".

The grid is set on the page at the moment.

rowanc1:
#897 has the original documentation on the grids.

choldgraf:
now that static files are supported, I wonder if we could have a hand-made HTML file that we bundle as a static file to show this off.... that way we can strip away all of the DOM and just have a blank page + the js/css for tailwind spacing classes that we support

PR #848, opened 2026-04-03
Fix color contrast for erroring code cells and add ANSI colors to stream outputs

### LINKED ISSUE #744: Accessibility: Color Contrast for Code Cells with Errors
Code cells that produce errors do not meet color contrast standards. An example of a `NameError` can be found on [this page](https://inferentialthinking.com/chapters/08/functions-and-tables/) with output from axe accessibility checker [here](https://github.com/data-8/textbook/actions/runs/20148056555/job/57833764033#step:9:62). The checker is configured to meet the standards required by ADA Title II.

I'm more than happy to help with this fix, and am filing this issue for tracking purposes.

I recently fixed this for a quarto site and am happy with the [chosen colors defined here](https://github.com/data-6-berkeley/notes/blob/main/assets/custom-error-colors.css)

pancakereport:
Closes #744 

I updated the colors used by `.jupyter-error` to meet WCAG 2.1 AA color contrast standards. The colors here are now also responsive to the light and dark modes.

UC Berkeley (who deploys multiple online textbooks with Jupyter Book) is required by law to meet this standard by [April 24](https://dap.berkeley.edu/ada-title-ii-updates/ada-title-ii-updates-what-does-it-mean-us), so I'd LOVE to get this merged by then. I'm planning to attend the collaboration cafe on April 7 to at the very least touch base about this PR, if not merge it.

Visual changes made by this PR:
Screenshot 2026-04-03 at 1 27 58 PM
Screenshot 2026-04-03 at 1 28 10 PM

Previous look:
Screenshot 2026-04-03 at 1 30 33 PM

Screenshot 2026-04-03 at 1 30 43 PM

I tried without the `!important` override, but the changes didn't render. If there's a better way to ensure that the changes take, let me know.


choldgraf:
I think that we might be able to get classes with these ansi things so we don't have to manually style them if we use a `useClasses` parameter for the react component, it is documented there though I'm not super familiar with it:

https://github.com/curvenote/ansi-to-react#classes

@stevejpurves can you provide guidance?

choldgraf:
OK I think I figured out how we can use the CSS classes - the latest two commits do that. The first uses CSS classes and adds rules for them, so we aren't hard-coding. The second adds a little padding because I think we lacked that as an oversight, and this is closer to how JB1 styles text/error outputs. Let's see how the outputs look.

Here's how they look in light mode

CleanShot 2026-04-04 at 12 14 04

Dark mode

CleanShot 2026-04-04 at 12 15 18

pancakereport:
Okay! Your changes make sense to me. I haven't run into the ansi stuff before, but I'm looking into it now!

rowanc1 (approved):
This looks solid.

agoose77:
We need to resolve the conflicts against the scrollable PR (easy to do), and I _think_ also generalise the ANSI handling for the non-error output too. Regular streams can include things like colours.

agoose77:
I tried to make some progress here. First, I noticed that myst-theme doesn't yet handle `stderr` stream outputs similarly to error outputs. That's OK for now, but we should.

Next, I noticed that the class-based approach drops unknown palette colours:
image

This is a regression vs the original:
image

JupyterLab does handle these with discrete classes,
image

so it should be do-able.

agoose77:
Hmm, I actually don't think this is entirely the issue. 

Something in the stack (ipykernel, I think) has changed recently regarding TB handling. One solve is producing raw colours, which are formatted into class names as e.g. `ansi-palette-124-fg`, whilst another solve just uses nice simple `ansi-yellow-bg`.

I'm out of time for this evening, probably, but my intention is to start by computing the entire set of palette styles and adding them to the stylesheet so that we see _something_.

What do we need to do? I think we need to look at what JupyterLab does in the `ansi-palette-124-fg` case, which I suspect is display _those_ colours. As such, by using classes we'll need to define the extended palette colours, which might not be a11y friendly. But, that's OK I think because we can't control it.

agoose77:
@pancakereport how did you compute the colours here? I am wondering how to adjust the other colours from the wider palette to follow suit.

pancakereport:
> @pancakereport how did you compute the colours here? I am wondering how to adjust the other colours from the wider palette to follow suit.

One tool that I like is [this color contrast checker](https://webaim.org/resources/contrastchecker/). You'll put in the foreground (text) color and the background color and it'll tell you the contrast between the two. MINIMUM contrast we need is 4.5 for WCAG 2.1 AA. It also has a color picker tool so you can see what making the foreground lighter/darker/some other color does for contrast.

Will be looking at your other comments in a bit...

choldgraf:
Here's the problem as I understand it:

- In some cases, there are *extra* classes that include a number, and we're "dropping them" because our CSS rules here are just for colors, not numbers+colors (e.g., `<span class="ansi-palette-28-fg ansi-bold">def</span>`)
- Here's a list of all the ansi colors that are possible here: https://github.com/IonicaBizau/anser/blob/d300b23541d831741b872340706848a013a3ccdc/lib/index.js#L5-L26

What we'd to do to unblock this:

1. @pancakereport will provide an example of the code we want to demonstrate for our error report. We'll use this as a reference example for the CSS classes we need to special case.
2. One of us will take a look at the ANSI class list above and see if it's easier to either programmatically generate CSS classes using the logic there for _all possible_ classes, or to special-case just the classes that are generated by the example we use.
3. Once the example in the myst-theme docs looks good, we can merge this PR, even though we know we probably are missing some other classes.

choldgraf:
OK merged in main so we can at least be working from a mergeable state now

pancakereport:
I tried out approx. 10 different ways to get errors with various datatypes and I'm only seeing colors that come with `add_one("hello")`; I haven't been able to get additional colors to show.

choldgraf:
I played around with this a little bit and pushed a commit that does the following:

- Makes the example a little bit more complex (uses a few other syntax pieces, though honestly don't know how much difference this makes in the error output)
- Added a CSS class hard-code for each class emitted there
- In choosing colors, I tried to pull from the accessible pygments project from quansight since that was developed for the pydata sphinx theme
- I made the red background of light mode a bit "lighter" so that the colors contrast a bit more
- I'm fine tweaking any of those colors, they are semi-arbitrary, so if y'all think of better ones that are a11y friendly just go for it

I thought about programmatically generating all possible ansi colors, but that felt like overkill since most of them will probably never be used.

Light mode:

CleanShot 2026-04-07 at 15 03 17

Dark mode:

CleanShot 2026-04-07 at 15 03 14

pancakereport:
From the screenshots the only one that I don't like is the dark mode magenta; it's very close to the white(?) so it's hard to tell the difference there imo. Let me see if there's a better purple-magenta color.

pancakereport:
Okay it's actually `.ansi-palette-28-fg` I wasn't liking. If we want to keep it purple, `#D8B4FE` is good:
Screenshot 2026-04-07 at 3 20 43 PM

Or I think a pink like `#F472B6` is actually more distinguishable from the white:
Screenshot 2026-04-07 at 3 22 19 PM

choldgraf (approved):
Your version looks good to me @pancakereport , thanks for updating the colors! IMO this looks good to go.

agoose77:
I've pushed the final changes I wanted to make, which is to ensure that we have colours for all palette indices. I've also fixed a bug in `stream` which was preventing stderr streams from displaying like error outputs.

image

image

I think we should aim for the _named_ colours to be accessible, but I'm not sure that it's possible (by definition) to dot he same for the entire palette. @pancakereport do you have any capacity to tweak these named (1-16) colours? I've defined them in `ansi.css`

choldgraf:
I think the new ansi.css is helpful - it feels like an increase in scope for this PR beyond the initial issue goal, but if the output looks good then I'm fine merging it as is as well. It feels like we've now added a _feature_ (ANSI colors in stream output) rather than just fixing a bug (make the error colors a11y friendly). Thanks for putting in the effort to solve a deeper problem @agoose77 !

agoose77:
> it feels like an increase in scope for this PR beyond the initial issue goal, 

This part — adding ansi.css — is preventing this PR introducing a regression!

> but if the output looks good then I'm fine merging it as is as well. It feels like we've now added a feature (ANSI colors in stream output) rather than just fixing a bug (make the error colors a11y friendly).

Yes, fixing ANSI colours in stream outputs is a separate bugfix!

choldgraf:
Something tells me this is not a11y-friendly:

CleanShot 2026-04-08 at 10 26 38@2x

I suggest that we:

- Run another color check on this and @pancakereport adjust colors as they wish, I trust your judgment there Silas!
- Fix the linting errors if those are legit (see cicd)
- Merge this

agoose77:
@choldgraf @pancakereport it looks like these are more examples of things that we need to handle — the commits that I added basically surfaced things that were previously unstyled.

I think our goal should be to make accessible the colours that we regularly see in the code outputs

pancakereport:
@agoose77 For the example error message in light mode I see a yellow with low contrast, but it doesn't seem to be one of the ones you define, it's `ansi-yellow-bg` NOT `ansi-yellow-fg`.

This should be fixed before merge. Are you okay with me defining this variable? `#707000` should work for light background; contrast of 4.79 with background `#FEF2F2`.

Screenshot 2026-04-08 at 5 59 25 PM

Also have this one I don't see a definition for: `ansi-palette-140-fg`. I think `#8A4FC4` is a good color for this on light mode.

And then for the other colors I'm going to ensure they're high contrast with the stderr stream. There will probably be a few violations still present against the white or black background.

pancakereport:
Not sure what to do about the `ansi-bright-white-fg`, `ansi-white-fg`. They're problematic on stderr stream light mode. Just make these gray? (But then the name doesn't make sense...)

Similarly for the `ansi-bright-black-fg`, `ansi-black-fg` on stderr stream dark mode. Should I just make them lighter (more gray-white)?

choldgraf:
I think it's safe to just manually change the ones you've listed, to the colors you suggested!

pancakereport:
I pushed colors that are high contrast with stderr stream. However, I did have to use `!important` to get some to take. I doubt this is the best approach here, but I'm not sure what's preferable.

choldgraf:
If you're doing overrides, esp w !important, I'd recommend adding a comment about why for each. Once we do that, if the colors in the preview look good to you imo we can go for it

pancakereport:
Is there a quick way to see if these are used elsewhere? I don't want to end up messing things up because of the `!important`

choldgraf:
Mmm I'd just do a find in the search bar in vscode to see if it is defined elsewhere. I think Angus auto generated his ansi.css file so I think it'd be reasonable to override in a different file like you did here

agoose77:
@pancakereport you shouldn't need to use !important to override these :frowning:. I didn't dig into why it was needed though, because I added a level of indirection via CSS variables. 

---

I think we might need to take a step back here.

TL;DR (below) — I think we should only patch `ansi-red` etc. here, and leave `ansi-palette` for downstream projects.

Details:

Stating my understanding for explicitness — It's not possible to make colours "accessible" — the accessibility is determined by the context in which they appear. In this case, the context is driven by the kernel, which is IPython (and Jupyter Book supports other kernels). So, the only option is to make a subset that we see accessible.

From the game of whack-a-mole, I don't think the approach we've tried here works — the colours vary too much by software environment and source material. 


There are two kinds of palette — the restricted 16-colour conceptual palette (red, white, bright-red, etc.), and the broader 256 - 16 colours that are effectively a spectrum.

I think myst-theme _can_ provide a base level of accessiblity for the 16 named colours if:
1. We choose the white that matches our page background
2. We choose other colours to have high contrast to each other
3. We don't support bright vs non-bright (I think IPython console does this already)

Then, for UCB I think there are two options:
1. Take the same approach as this PR, and define custom `ansi-palette-XXX` colours that fix the textbooks you have. Use https://mystmd.org/guide/website-style#style-sheet to apply these overrides.
2. Patch IPython to use a restricted colour palette.

(2) is by far the best approach, but I didn't have much luck when trying it myself.

@pancakereport @choldgraf how does this approach sounds? 

To nerd-snipe someone, I think it should be possible to patch IPython's colour themes to use the Pygments `TerminalFormatter` rather than `Terminal256Formatter`. But this didn't work when I munged the `theme_table` values, so I might be missing something.

choldgraf:
I think we should change the absolute minimal lines of code that makes error messages visually accessible. I'll defer to your judgment about what that is.

Is this a commit we can just revert to? This one only defined over-rides for the things we _knew_ were in the error outputs:

https://github.com/jupyter-book/myst-theme/pull/848/changes/bca30fe324a4e3857b9d09acf2db0bfc0605f210

agoose77:
@choldgraf I don't think so. The later PRs don't change what we're receiving -- my point is that this is so fragile that if the goal is to get UCB to pass, then I think this needs to be vendored on the UCB side, possibly with pinning for the Python packages.

choldgraf:
> this is so fragile that if the goal is to get UCB to pass, then I think this needs to be vendored on the UCB side, possibly with pinning for the Python packages.

I don't understand this bit - we're talking about a standardized theme background (light red in error outputs) with standardized error outputs coming from a specific package (IPython). If one user has accessibility problems then all other users will have the same problems, no? If you're worried about screwing up the ansi colors in non-error context, then why couldn't we:

- Use `ansi.css` just to make sure that all ansi classes are picked up and rendered _at all_.
- Use the final part of that file (or another called `ansi-a11y.css` that only includes CSS *overrides* for ANSI classes that we know break a11y under certain conditions)

Is the issue the fact that it's a moving target based on the IPython version? If so, then IMO we should just pick the version that we want to make work (e.g. latest stable release) and make sure the myst-theme always meets a11y standards for that version.

pancakereport:
I think that we all agree that what exists without this PR isn't a great user experience; it's not responsive to light/dark modes and in dark mode I personally can't read the error output. 

If we don't want to include everything that was worked on for this PR, ok. I can definitely make custom changes for Berkeley textbooks. But I do think changing something to ensure error outputs are readable (putting specific a11y standards to the side) is needed, at least for dark mode.

choldgraf:
OK I just pushed a commit w/ my proposal. Here's basically how it works:

- `ansi.css` has pre-computed ansi colors using Angus' original calculation. It defines the CSS variable defaults, and applies them to the fg/bg classes. (note: we're still only importing `ansi.css` in the jupyter file... I think it'll still apply everywhere bc of how the rules are defined so maybe we should change that?)
- `ansi_overrides.css` has *our* over-rides for specific colors, where we can make more assumptions about them. There are currently two kinds of over-rides:
  - Default overrides (assume our standard theme light/dark background, and adjust just a few colors that are obviously not visible, like yellow-on-white)
  - Jupyter error overrides (assumes our red error background, and is the fix for this original issue)
- A user could thus override any of these ANSI colors by just defining their own CSS variables

I looked at the meeps and they look good, I also looked at the error message and it looked good to me 

I think this means our `ansi.css` will mean any ansi-style classes will "just work", and the overrides are more targeted for contexts where we know we need the overrides. Here's how it looks

CleanShot 2026-04-09 at 15 28 10

pancakereport:
Just pulled and ran tests. Error messages are good on the color contrast front. Happy with this.

agoose77:
@pancakereport and @choldgraf I'm not sure that my argument is coming through clearly enough, so I'll try to clarify. Apologies for the wall of text, but I think it's needed. Thank you both for your patience and hard work on this.

## We are aligned on the a11y problem

> I think that we all agree that what exists without this PR isn't a great user experience;

Yes. We are all agreed that the _user story_ of this PR is valid:

> [!Important]
> Error outputs are not readable in dark mode, and tracebacks often are not high-contrast.

My position is that we might need to do this in two separate places (I'll explain below).

## We don't agree on the specific solution
> I don't understand this bit [...] standardized error outputs coming from a specific package (IPython). [...]
> 
> Is the issue the fact that it's a moving target based on the IPython version? 

Yes! The observation that I'm making is that the error outputs clearly _aren't_ "standardised":
- A different Jupyter Book author runs the `%colors` magic in their book to change the theme colours
- A different Jupyter Book author uses a Java, or R kernel instead of IPython
- A different Jupyter Book author uses IPython <9 instead of IPython 9.

For posterity, there are _a lot_ of Jupyter Kernels in the wild: https://github.com/tariqueus/jupyter-kernels

---

The approach we've taken in `ansi_overrides.css` is to say:

> We set our ANSI colour codes to nicely render tracebacks from the IPython kernel version XXX

To compute the set of CSS variables that we need to tweak, we have done the following:
1. Run a particular kernel (IPython) with a particular version (XXX) and generated a particular traceback which is lexed to tokens, e.g. `Name.Constant`
2. IPython has a user-chosen Pygments theme, that has particular hex code colours e.g. [`Name.Constant = #880000`](https://github.com/pygments/pygments/blob/c001c2521f0d1b8fc8db2f2150c3a0a195e36e92/pygments/styles/default.py#L46)
3. Pygments defines an arbitrary _colour map_ for the ANSI palette e.g. [`xterm_colors = ...`](https://github.com/pygments/pygments/blob/c001c2521f0d1b8fc8db2f2150c3a0a195e36e92/pygments/formatters/terminal256.py#L152-L186)
4. Pygments maps theme colours to ANSI colours by [finding the closest colour](https://github.com/pygments/pygments/blob/c001c2521f0d1b8fc8db2f2150c3a0a195e36e92/pygments/formatters/terminal256.py#L188-L203)
5. We render HTML with those classes, and assign them a _different_ colour palette.

This can break in the following ways:

- A different Jupyter Book author runs the `%colors` magic in their book to change the theme colours
- A different Jupyter Book author uses a Java, or R kernel instead of IPython
- A different Jupyter Book author uses IPython <9 instead of IPython 9.
- Pygments change their theme colours
- Pygments change their ANSI palette
- IPython changes the default theme

Given that (a) this is so fragile, and (b) we're not "IPython-9-linux-theme-book" but rather "Jupyter Book", I think we should not try to solve the a11y problem in this way.

## What we should do instead

### 1. Define a11y-friendly named colours

Here's the xterm named colour palette:

image

This is arbitrary — any terminal / renderer can choose its own. We **should** choose `ansi-black` through to `ansi-white` such that they are a11y friendly in console output and error output for the text-on-white/black scenario. We can _choose that_ — there are only eight colours if we do not distinguish between standard and high intensity.

### 2 (Option a). Project the remaining colour space into named colours
The remaining palette `ansi-palette-16 → ansi-palette-255` is a colour "space" rather than a discrete palette like the named colours:

image

We can't have all these colours be distinct whilst also supporting a11y contrast — contrast requires that there is enough distance between two colours.

To solve this, we could project these colours onto the 8 a11y-friendly colours. Here's an example of where I've done that (see this script https://gist.github.com/agoose77/5a3da28ac3b6244201fd79c4e43d527d):
image

This approach _sounds_ like what we're already doing in this PR except:
- It's exhaustive — every ANSI colour is mapped to something we deem "accessible" in a prose context
- It's deterministic — we are quantising the colour space, which is a realistic interpretation of what we're trying to achieve.

We can possibly make it possible to add a special CSS class to opt-out of these colours in certain outputs if it is really important.

### 2 (Option b). The end-user sets a11y-friendly IPython config
The issue here is that IPython is not a11y friendly — they have an arbitrary colour map for tracebacks. We could determine the configuration (which is possible) to restrict IPython to use our a11y-friendly named colours, and require UCB text-books to set this in the config.


### 2 (Option c). The end-user sets a11y-friendly CSS
If we can't easily do 2.b., another option is to take the same approach in this PR — cherry picking colours — but to do so on a per-textbook basis. Because UCB controls their build environment, they can pin the IPython version etc. to ensure this doesn't break.

agoose77:
The lack of reply from me here is a reflection of the fact that I've been trying to find the right design for this. It's 90% there, but I'll take another stab later today or tomorrow.

choldgraf:
I don't think I have the mental bandwidth to really dig into this any more. Personally:

- I am totally fine with a hacky solution that special-cases IPython. I suggest we use "the latest version of IPython" or something, but IMO python is common-enough to special case.
- If you have an implementation that you like and that makes jupyter errors accessible, let's just go with it. I trust your judgment here.
- What I don't think we should do is require _users_ to write special CSS just to make the theme accessible, that is the thing I feel more strongly about.
- So `tl;dr`: Any solution that doesn't require users to alter their behavior for a11y purposes is one I will support. Just go for it

agoose77:
Yeah, it's a gnarly one.

> What I don't think we should do is require users to write special CSS just to make the theme accessible, that is the thing I feel more strongly about.

I agree in principle. I think the difficulty is when you're an intermediary rather than a content provider. We can make myst-theme accessibly, but we can't make the user-provided tools accessible without a lot more work (see this PR!)

In any case, I think we can find a compromise in the direction I'm moving.

choldgraf:
> I think the difficulty is when you're an intermediary rather than a content provider. We can make myst-theme accessibly, but we can't make the user-provided tools accessible without a lot more work (see this PR!)

I agree it's a whack-a-mole thing. Personally, I think that's fine. Even if we ended up with an `ansi_overrides.css` monstrosity that was 3000 lines long and we just kept adding to it whenever somebody showed up with a new ansi class we hadn't heard of before, that'd be fine. In practice, I suspect that if we fix these errors as they pop up, we'll quickly stop getting error reports about it because all the theoretical cases we don't support (e.g. random kernels or versions) aren't problems in _practice_

pancakereport:
Thank you both so much for this. Let me know if there's anything I can do to help with this.

choldgraf:
FYI I've written up my understanding of the deeper problem in this issue:

- https://github.com/jupyter-book/mystmd/issues/2812

@agoose77 please edit it to make it better if I got something wrong! I think it does make it clearer why this is more complex than we thought. In looking at VSCode's approach, it seems like the combination of "css classes (or variables) for named colors, but use inline styles for palette colors, is a reasonable approach if we can figure that out)

agoose77:
This week has been a lot busier than expected for various reasons, so I'm only getting to this now. I'll finish this up over the next few days, although anyone is welcome.

Here's a toy example of what I'm planning to do: https://play.tailwindcss.com/SeluKcu3j9
image


It does several things:

1. Two palettes (full, restricted)
2. Two palette modes (full (bg + fg) vs restricted (bg→fg, fg only)
3. Dark + light variants
4. Error vs stream variants

I suspect that naked CSS is actually better in the long run than Tailwind components, after my previous efforts in this PR make it feel a bit hard to override.

agoose77:
Thanks for this PR @pancakereport, and @choldgraf for reviewing! I've closed it in favour of #855.

PR #849, opened 2026-04-03
Make stderr and stdout from code cells scrollable if they overflow

### LINKED ISSUE #743: Accessibility: Scrollable Regions
Scrollable regions (like wide code cells) are not accessible via keyboard. An example is the cell defining `biggest_difference()` [here](https://inferentialthinking.com/chapters/08/functions-and-tables/) with output from axe accessibility checker [here](https://github.com/data-8/textbook/actions/runs/20148056555/job/57833764033#step:9:62). The checker is configured to meet the standards required by ADA Title II.

I'm more than happy to help with this fix, and am filing this issue for tracking purposes.

The way to fix this is simply to add `tabindex="0"` (or the appropriate number).

pancakereport:
Closes #743 

This PR ensures that users who navigate with a keyboard are able to view (potentially) scrollable code/code output regions. I've tested for the code cells themselves, normal output, and output with errors.

UC Berkeley (who deploys multiple online textbooks with Jupyter Book) is required by law to meet this standard by [April 24](https://dap.berkeley.edu/ada-title-ii-updates/ada-title-ii-updates-what-does-it-mean-us), so I'd LOVE to get this merged by then. I'm planning to attend the collaboration cafe on April 7 to at the very least touch base about this PR, if not merge it.

This is a narrow fix for scrollable regions relating to code cells and their output. Other potentially scrollable regions (like tables) likely still present accessibility violations. 

Tested with [axe devTools chrome extension](https://chromewebstore.google.com/detail/axe-devtools-web-accessib/lhdoppojpmngadmnindnejefpokejbdd?hl=en-US)

I'm not the most familiar with React, so please point out any potential problems with this (bonus points if you link me to resources).

### DIFF
--- a/.changeset/forty-parents-ring.md
+++ b/.changeset/forty-parents-ring.md
@@ -1,0 +1,7 @@
+---
+'myst-to-react': patch
+'@myst-theme/providers': patch
+'@myst-theme/jupyter': patch
+---
+
+Make overflowing content focusable
--- a/docs/reference/blocks.md
+++ b/docs/reference/blocks.md
@@ -32,6 +32,15 @@ def calculate_sum(a, b):
     return result
 ```
 
+### Wide Code Block
+
+This code block has long lines and should be horizontally scrollable. Keyboard users should be able to tab to it and scroll with arrow keys.
+
+```python
+def print_a_really_long_string():
+    print("MyST is really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really neat!")
+```
+
 ## Blockquotes
 
 > Basic blockquote for quoted text.
--- a/docs/reference/computation.md
+++ b/docs/reference/computation.md
@@ -20,6 +20,7 @@ These cells produce wide outputs to test that scrollable regions are keyboard-ac
 ```{code-cell} python
 def print_a_really_long_string():
     print("MyST is really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really neat!")
+print_a_really_long_string()
 ```
 
 ## Pandas
--- a/docs/reference/computation.md
+++ b/docs/reference/computation.md
@@ -13,16 +13,6 @@ This page contains example content for previewing or demonstrating computational
 The [Example Outputs Site](https://jupyter-book.github.io/example-outputs/) has more comprehensive examples to show off MyST generating outputs with computation.
 :::
 
-## Wide cell inputs and outputs
-
-These cells produce wide outputs to test that scrollable regions are keyboard-accessible (tab to focus, arrow keys to scroll).
-
-```{code-cell} python
-def print_a_really_long_string():
-    print("MyST is really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really neat!")
-print_a_really_long_string()
-```
-
 ## Pandas
 
 ```{code-cell}
@@ -98,4 +88,14 @@ p.yaxis.axis_label = 'Column B'
 p.grid.grid_line_alpha = 0.3
 
 show(p)
+```
+
+## Wide cell inputs and outputs
+
+These cells produce wide outputs to test that scrollable regions are keyboard-accessible (tab to focus, arrow keys to scroll).
+
+```{code-cell} python
+def print_a_really_long_string():
+    print("MyST is really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really neat!")
+print_a_really_long_string()
 ```
--- a/docs/reference/computation.md
+++ b/docs/reference/computation.md
@@ -13,6 +13,15 @@ This page contains example content for previewing or demonstrating computational
 The [Example Outputs Site](https://jupyter-book.github.io/example-outputs/) has more comprehensive examples to show off MyST generating outputs with computation.
 :::
 
+## Wide cell inputs and outputs
+
+These cells produce wide outputs to test that scrollable regions are keyboard-accessible (tab to focus, arrow keys to scroll).
+
+```{code-cell} python
+def print_a_really_long_string():
+    print("MyST is really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really neat!")
+```
+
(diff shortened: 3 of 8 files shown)


choldgraf:
Thanks @pancakereport - where do you think we should look in the docs to understand if this has introduced a bug. It sounds like this is particularly for long code outputs right? Do we have an example here that you were using to compare against?

pancakereport:
Yes, this fix works for code/code output that is longer (wider?) than the window. If the code/code output is bigger left-right than the window size allows for, it becomes left-right scrollable. Without this PR those scrollable regions are not accessible via keyboard so any keyboard-only users won't be able to see/access what's being cut off to the right.

While working on this, I was looking at the [page](https://inferentialthinking.com/chapters/08/functions-and-tables/) referenced in #743 which includes a particularly wide code cell by default (beginning with `def biggest_difference(array_x):`). Changing the size of my window let me test for code output that is left-right bigger than the window. This page has both erroring and non-erroring cell output.

If it's helpful I can provide the html generated with/out this fix for this page. Hopefully this is roughly what you were asking for, but let me know if I'm misunderstanding.

choldgraf:
I just pushed a couple of examples that I think we could use to test this. For each of them, we should tab-onto the output/code block then high the "right arrow" and I think that should test out whether this works.

choldgraf:
OK I think I figured out how to clean this up a little bit by moving the "is scrollable" function into a shared module. I believe that the way to test this is to:

- Click a code cell
- Hit tab
- If the cell contents is wider than the cell, then it should _highlight the code cell_ and hitting right-arrow will scroll to the right.
- If not wider, then it just skips the code cell
- Same for various outputs

@pancakereport want to test again if Netlify is happy?

choldgraf:
@pancakereport is it worth creating a parent issue about this upcoming Berkeley accessibility deadline? It could be a helpful way to focus attention on accessibility for a bit

pancakereport:
`useIsScrollable` is much cleaner! Thanks for that. I checked it out on the netlify preview and everything seems to be working correctly. I think this can be merged now. 

I can also make another PR for tables that are wide/scrollable using this PR as an example (use the shared module, update reference docs). Are there other things that you can think of that are possibly left-right scrollable?

Also, happy to make a parent issue. Is there a particular jupyter-related GH org/repo that would find it most useful? (Maybe multiple?)

choldgraf:
I think jupyter-book/mystmd is the most helpful place for a parent issue!

If you'd like I think it'd be fine to add tables for this PR as well, since it'll hopefully just be the exact same `useIsScrollable` fix it should just be a one-liner

pancakereport:
Hmmm, I put a 40-column table (below) in and it doesn't end up with the same scrolling behavior as code/code output. The table expands into portion of the screen normally reserved for the secondary sidebar (page's "Contents"), making the whole page scrollable, not the table itself. This also means that when you scroll to the right, what was anchored in the content window overlaps with the primary sidebar.

This is visually messy but doesn't present an accessibility violation because the left-right scroll (applied to the whole page) is keyboard accessible. 

If we want to change this, I'd be happy to take a first pass at it, but **I think it's a separate issue**. 

```
| Col 1 | Col 2 | Col 3 | Col 4 | Col 5 | Col 6 | Col 7 | Col 8 | Col 9 | Col 10 | Col 11 | Col 12 | Col 13 | Col 14 | Col 15 | Col 16 | Col 17 | Col 18 | Col 19 | Col 20 | Col 21 | Col 22 | Col 23 | Col 24 | Col 25 | Col 26 | Col 27 | Col 28 | Col 29 | Col 30 | Col 31 | Col 32 | Col 33 | Col 34 | Col 35 | Col 36 | Col 37 | Col 38 | Col 39 | Col 40 |
|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|:---|
| Data A1 | Data A2 | Data A3 | Data A4 | Data A5 | Data A6 | Data A7 | Data A8 | Data A9 | Data A10 | Data A11 | Data A12 | Data A13 | Data A14 | Data A15 | Data A16 | Data A17 | Data A18 | Data A19 | Data A20 | Data A21 | Data A22 | Data A23 | Data A24 | Data A25 | Data A26 | Data A27 | Data A28 | Data A29 | Data A30 | Data A31 | Data A32 | Data A33 | Data A34 | Data A35 | Data A36 | Data A37 | Data A38 | Data A39 | Data A40 |
| Data B1 | Data B2 | Data B3 | Data B4 | Data B5 | Data B6 | Data B7 | Data B8 | Data B9 | Data B10 | Data B11 | Data B12 | Data B13 | Data B14 | Data B15 | Data B16 | Data B17 | Data B18 | Data B19 | Data B20 | Data B21 | Data B22 | Data B23 | Data B24 | Data B25 | Data B26 | Data B27 | Data B28 | Data B29 | Data B30 | Data B31 | Data B32 | Data B33 | Data B34 | Data B35 | Data B36 | Data B37 | Data B38 | Data B39 | Data B40 |
```
Screenshot 2026-04-06 at 4 27 21 PM

choldgraf:
Agreed it is a separate issue! Let's leave the demo in this PR, then we can open an issue that points to the docs for an example of this bug 

Btw it's there anything that has a vertical scroll complement? (Eg a really long code cell?) If so then we might need the same behavior for a11y purposes

pancakereport:
Just pushed the wide table example. Happy to open that issue once this is merged.

Thinking about vertical scrollable regions and I'm only really coming up with sidebars (which haven't been flagged) and smaller, intentionally scrollable regions (like long terms and conditions embedded on a webpage), but I'm not sure how this would happen in JB. I think really long code cells don't trigger this, they just stay vertical scrolling with the webpage ([this](https://data8.org/materials-sp26/hw02/) has a code cell thats ~50 lines).

agoose77:
I've tested this using the helpful examples, and I'm happy that it improves things. Thanks for driving this forward @pancakereport and review @choldgraf

PR #853, opened 2026-04-07
Destructure aria-controls and only use it when its target exists in search

### LINKED ISSUE #808: Missing matching element for `aria-controls` value
A WCAG audit of pages generated with book theme is reporting an invalid value with the `aria-controls` attributes on the page. If I understand the [documentation on that attribute](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-controls), the value is intended to match the `id` attribute of another element in the document. The search button element at the top right of the page is an example where there is not a matching element as far as I can see. 

```
<button type="button" aria-haspopup="dialog" aria-expanded="false" aria-controls="radix-:R75cp:" data-state="closed" class="myst-search-bar flex items-center h-10 aspect-square sm:w-64 text-left text-gray-600 dark:text-gray-300 border border-gray-300 dark:border-gray-600 rounded-lg bg-gray-50 dark:bg-gray-700 myst-search-bar-disabled hover:ring-blue-500 dark:hover:ring-blue-500 hover:border-blue-500 dark:hover:border-blue-500">...</button>
```

This is the error I see when running the `axe-selenium-python` package against the JupyterBook pages. I was not able to reproduce this with axe DevTools in Chrome.

```
- 🛑‼️  ARIA attributes must conform to valid values (https://dequeuniversity.com/rules/axe/3.1/aria-valid-attr-value?application=axeAPI):

  Fix all of the following:
  Invalid ARIA attribute value: aria-controls="radix-:R75cp:"

  HTML fragment: <button type="button" aria-haspopup="dialog" aria-expanded="false" aria-controls="radix-:R75cp:" dat
```

Thanks for looking into this!

choldgraf:
This is a quick fix for the aria-controls bug that @erikamov reported!

I think that we might have similar bugs (that are less-common to trigger) in other places where we have "pop up" menus (e.g. launch buttons) but didn't have time to investigate and just wanna get this fix in

---

- closes https://github.com/jupyter-book/myst-theme/issues/808

### DIFF
--- a/.changeset/silent-kids-start.md
+++ b/.changeset/silent-kids-start.md
@@ -1,0 +1,8 @@
+---
+'myst-to-react': patch
+'@myst-theme/providers': patch
+'@myst-theme/jupyter': patch
+'@myst-theme/site': patch
+---
+
+Fix aria-controls in search popup
--- a/.changeset/silent-kids-start.md
+++ b/.changeset/silent-kids-start.md
@@ -1,7 +1,4 @@
 ---
-'myst-to-react': patch
-'@myst-theme/providers': patch
-'@myst-theme/jupyter': patch
 '@myst-theme/site': patch
 ---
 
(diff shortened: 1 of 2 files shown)


choldgraf:
to test this out I:

- Looked at the HTML of the search button. It doesn't have `aria-controls`
- I opened the search dialog.
- Checked that the HTML now **does** have `aria-controls`

agoose77:
LGTM!

PR #854, opened 2026-04-07
Improve search accessibility

### LINKED ISSUE #823: Accessibility of search
Search has some broken accessibility features, it seems like it is too verbose.

- ht: https://github.com/the-turing-way/the-turing-way/issues/4534

choldgraf:
This adds some aria metadata to our search results in the hopes that it makes it a bit more screen-reader-friendly. This is not my expertise, but these seemed like straightforward things to improve the UX a bit for screen readers!

Major things to note:

- `aria-label` is now explicitly set (though cut-off at 120 characters since some of the descriptions were super long)
- the `<a>` below now has `aria-hidden`, which I think means that screen readers will just focus on the `li` instead of the underlying `a`, and the label should give the useful text we need
- In other parts i added a `role` which I think should make it stop reading like a big blob of text

Here's what each item spits out.

```html
<li data-index="2" role="option" aria-selected="true" aria-label="Long Content - This document demonstrates how the theme handles long-form content. It includes multiple sections, subsections, and vari…" class="myst-search-result-item group">
  <a class="block px-1 py-2 text-gray-700 rounded shadow-md dark:text-white group-aria-selected:bg-blue-600 group-aria-selected:text-white dark:shadow-none dark:bg-stone-800" aria-hidden="true" href="/reference/really-long#introduction">
     [unchanged stuff inside]
  </a>
</li>
```

It would be really helpful if a person that has a screen reader could test this out and give any pointers - I don't have a ton of cycles to put into this, but am happy to iterate a bit if it can give an incremental improvement! cc @JimMadge and @LizHareDogs maybe?

### To test this

- Go to the docs preview here: [Docs preview](https://deploy-preview-854--myst-theme.netlify.app/)
- Click the search bar or click `cmd+K`
- Type something, scroll through search results, report back if they are useful on screen readers!

---

- closes https://github.com/jupyter-book/myst-theme/issues/823 (somebody please tell me if it does not!)

### DIFF
--- a/.changeset/busy-games-find.md
+++ b/.changeset/busy-games-find.md
@@ -2,4 +2,5 @@
 '@myst-theme/site': patch
 ---
 
-Add aria-label and other annotations to search items
+Re-order search entries to better behave under screen-readers.
+Add missing aria-expanded attribute.
--- a/.changeset/busy-games-find.md
+++ b/.changeset/busy-games-find.md
@@ -1,0 +1,5 @@
+---
+'@myst-theme/site': patch
+---
+
+Add aria-label and other annotations to search items
--- a/packages/site/src/components/Navigation/Search.tsx
+++ b/packages/site/src/components/Navigation/Search.tsx
@@ -525,7 +525,6 @@ function SearchForm({
               { 'border-red-500': !enabled },
             )}
             id={searchInputID}
-            role="combobox"
             aria-labelledby={searchLabelID}
             aria-controls={searchListID}
             aria-expanded={!!searchResults}
--- a/packages/site/src/components/Navigation/Search.tsx
+++ b/packages/site/src/components/Navigation/Search.tsx
@@ -361,9 +361,15 @@ function SearchResults({
     },
     [onHoverSelect],
   );
+  const searchResultsWithAria = useMemo(() => {
+    return searchResults.map((result) => {
+      const ariaLabel = getResultAriaLabel(result);
+      return { result, ariaLabel };
+    });
+  }, [searchResults]);
   return (
     <div className="myst-search-results mt-4 overflow-y-scroll">
-      {searchResults.length ? (
+      {searchResultsWithAria.length ? (
         <ul
           // Accessiblity:
           // indicate that this is a selectbox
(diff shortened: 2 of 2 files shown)


agoose77 (changes requested):
I took a look at this — I think we need to tweak it.

`aria-label` is for short names whereas search matches are long-form. Possibly `aria-description` here, except that we _do_ have short search text in the DOM, so `aria-describedby`.

`aria-hidden` on the `<a>` is wrong — the `a` _is_ needed to be interacted with, which is why the `<li>` has `tabindex=-1`. We're mostly following best practices here (I think) which is to treat a link like a link. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-hidden#description:~:text=Warning%3A%20Do%20not%20use%20aria%2Dhidden%3D%22true%22%20on%20focusable%20elements

--- 

My suggestion:

We swap around the "context" part of the match so that it's at the top. This both fixes the a11y ordering and I'm happy personally with both appearances: Image

I looked at MDN which I expect to have decent accessibility, and they do the same thing -- their links are a fusion of document title and content. So, I think that's ok.

FYI I tested this in Talkback on Android.

.changeset/busy-games-find.md
```diff
@@ -0,0 +1,5 @@
+---
+'@myst-theme/site': patch
+---
+
+Add aria-label and other annotations to search items
```
agoose77:
```suggestion
Re-order search entries to better behave under screen-readers.
Add missing aria-expanded attribute.
```

agoose77:
OK, I made the suggested changes and also removed some combobox roles. The `input` already has an implicit role from `type=search`, and we don't want to clobber that. aria-roles for search is actually really confusing, and I remember spending a lot of time thinking about it when we first built this. I'm not sure there's great consensus on the pattern for search-as-you-type.

If we later learn this needs changing, let's do it then?

choldgraf:
IMO just go for it - my primary concern is whether this is an iterative improvement over what is on `main`, so if you think it is then I'm +1 on merging

PR #855, opened 2026-04-19
Change ANSI palette to restricted

agoose77:
This PR replaces #848 (so that I can force-push to rebase things). It implements the "restricted" strategy in https://github.com/jupyter-book/myst-theme/pull/848#issuecomment-4267406898

It has only the 8 ANSI colours, and background colours are mapped onto foreground:

image

image

Non-dim modifiers are supported. This ensures that we can curate an accessible palette that is visible against the background in all contexts.

> [!Important]
> In future, we'll want to add support for a non-restricted palette. However, given that to enable such a palette will be opt-in, I think we should do that separately to iron out UX challenges there.

<details><summary>Example of full palette, now removed</summary>
<p>

The full palette supports all named and unnamed colours. It supports BG and FG colours, and all style modifiers.

image
image

</p>
</details>


agoose77:
Given that we've already had a lot of back and forth (i.e. some level of eyes-on), I'm going to merge this PR _solely_ on the basis of timing for UCB. I don't like doing that, and this might be a mistake, but I think we should be able to walk things back sometimes!

I am certain there's a more elegant way to achieve the goals set out in this PR description, but time is of the essence.

rowanc1 (commented):
Looked over on my phone. If it is working, good to go from my perspective.

agoose77:
@rowanc1 sorry -- need to refresh with the latest build -- forgot the bg color (and had to tweak some dark styles).

I think we'll need to redo the colours, but they _are_ WCAG friendly on my machine.

packages/jupyter/src/stream.tsx
```diff
@@ -17,9 +20,12 @@ export default function Stream({ output }: { output: MinifiedStreamOutput }) {
           tabIndex={isScrollable ? 0 : undefined}
           role={isScrollable ? 'region' : undefined}
           aria-label="cell output stream"
-          className="myst-jp-stream-output text-sm font-thin font-system overflow-auto"
+          className={classNames('myst-jp-stream-output text-sm font-thin font-system p-3', {
```
agoose77:
```suggestion
          className={classNames('myst-jp-stream-output text-sm font-thin font-system overflow-auto', {
```

styles/jupyter.css
```diff
@@ -60,3 +56,60 @@ html.dark {
   --jp-layout-color3: var(--md-grey-700);
   --jp-layout-color4: var(--md-grey-600);
 }
+
+/* ======================================================================
+ * Regular stdout
+ * ====================================================================== */
+/* Light mode */
+.jupyter-output {
+  --ansi-black: rgb(0 0 0);
```
agoose77, line 65:
```suggestion
  @apply bg-red-100;
  --ansi-black: rgb(0 0 0);
```

packages/jupyter/src/error.tsx
```diff
@@ -17,9 +17,9 @@ export default function Error({ output }: { output: MinifiedErrorOutput }) {
           tabIndex={isScrollable ? 0 : undefined}
           role={isScrollable ? 'region' : undefined}
           aria-label="cell error output"
-          className="myst-jp-error-output text-sm font-thin font-system jupyter-error overflow-auto"
+          className="myst-jp-error-output text-sm font-thin font-system overflow-auto p-3 jupyter-error ansi"
```
agoose77:
```suggestion
          className="myst-jp-error-output text-sm font-thin font-system overflow-auto jupyter-error ansi"
```

styles/jupyter.css
```diff
@@ -60,3 +56,61 @@ html.dark {
   --jp-layout-color3: var(--md-grey-700);
   --jp-layout-color4: var(--md-grey-600);
 }
+
+/* ======================================================================
+ * Regular stdout
+ * ====================================================================== */
+/* Light mode */
+.jupyter-output {
+  @apply bg-red-100;
```
agoose77:
```suggestion
```

agoose77:
I've revised this PR to do _one_ thing — change the current palette to restricted. This means that existing outputs with rich ANSI will lose some of their richness. 

I think that's OK — this PR would change the default _anyway_. By not shipping a mechanism for enabling "true" ANSI colours in this PR, we have more time to figure out the right path. 

I'll open an issue / Discord.

PR #857, opened 2026-04-20
Add full-colour support to ANSI renderers

agoose77:
This PR adds support for full-color (8-bit) rendering of ANSI escape codes in code-cell outputs. It introduces support for two classes in the DOM:

- `full-color-error` 
- `full-color-output`

These two classes, appearing anywhere in the DOM above the `.jupyter-error` and `.jupyter-output` classes, will change the ANSI palette, render mode, and formatters to the full 8-bit set.

### DIFF
--- a/docs/reference/computation.md
+++ b/docs/reference/computation.md
@@ -100,7 +100,7 @@ print("Hello, world!")
 print("Result:", 2 + 3)
 ```
 
-A simple cell with ANSI codes, such as color, emphasis, etc.:
+A simple cell with ANSI codes, such as color, emphasis, etc. This uses a restricted palette, which is designed for maximum readability at the expense of styling richness. For example, dim and bright styles are disabled, and only foreground colours are used. Background colours are remapped to the foreground:
 
 ```{code-cell} python
 colour_palette = (
@@ -120,7 +120,7 @@ print(colour_palette, file=sys.stderr)
 print(colours_named, file=sys.stderr)
 ```
 
-We can also set the `full-color-output` class to show raw colours:
+We can also set the `full-color-output` class to show raw colors. Unlike the earlier example, raw colors include the full 8-bit ANSI palette, dim and bright styles, and both foreground and background colors:
 
 ```{code-cell} python
 :class: full-color-output
--- a/docs/reference/computation.md
+++ b/docs/reference/computation.md
@@ -189,7 +189,7 @@ Embedding a full-color cell loses the annotation:
 
 ![](#code-cell-dataloader)
 
-But we can restore it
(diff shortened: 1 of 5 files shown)


styles/ansi.css
```diff
@@ -1,34 +1,16 @@
...
-  text-decoration-line: underline;
-}
-.ansi-hidden {
-  visibility: hidden;
-}
 /* 
  * Restricted colour mapping (FG only) 
  */
-:is(.jupyter-error, .jupyter-output) span[class*='ansi-'] {
+:where(.jupyter-error, .jupyter-output) span[class*='ansi-'] {
```
agoose77, line 4: This could be `is` or `where`. `where` has lower precedence, and I'm using it here only for consistency with later appearances, where it is needed.

styles/ansi.css
```diff
@@ -288,6 +270,300 @@
...
+:where(.full-color-error .jupyter-error, .full-color-output .jupyter-output) span[class*='ansi-'] {
+  /* Styling */
+  color: var(--fg-color);
+  background-color: var(--bg-color);
+}
+
+/*
+ * Unrestricted palette computed from https://en.wikipedia.org/wiki/ANSI_escape_code#8-bit
+ */
+:where(.full-color-error .jupyter-error, .full-color-output .jupyter-output) {
```
agoose77, line 307: Here we _need_ `:where` to lower the specificity, so that `jupyter.css` can set palettes via `.jupyter-output`. Otherwise, we'll need to use specificity hacks in `jupyter.css` to override the palette. This is nicer, and allows users to more easily tweak these.

agoose77: By using `:where`, we're relying on last-wins to displace the earlier restricted palette definition.

docs/reference/computation.md
```diff
@@ -144,6 +164,41 @@ class DataLoader:
...
+        return data.split("\n")[:limit]
+
+DataLoader.load(42)
+```
+
+Embedding a full-color cell loses the annotation:
+
+![](#code-cell-dataloader)
+
+But we can restore it
```
rowanc1:
```suggestion
But we can restore it with a `full-color-error` class in a wrapping div or figure.
```

rowanc1 (approved):
I think small tweaks on dev-docs to make it more obvious what is happening when you can't see the source would be helpful.

As a follow on to this opening a PR into mystmd with a small amount of explaining this class would be helpful.

docs/reference/computation.md
agoose77: Nice. I was fairly exhausted getting this PR finished over the weekend, and could have used a few more words!!

PR #860, opened 2026-04-21
Fix equations scrollable 

### LINKED ISSUE #858: Fix scrollable region focusable (equations)
"Wide" Math equations should be scrollable region focusable, similar to code #849 and tables #852.

I'll submit a PR with a fix and an example of a "wide" equation later today.

An example page with some wide equations: https://inferentialthinking.com/chapters/18/1/more-likely-than-not-binary-classifier/

pancakereport:
Closes #858 

This PR replaces an example of display math with a wide (scrollable) one and implements the accessibility fix `tabindex="0"` following what Chris did in #852.

This PR and associated issue are part of the ongoing work in https://github.com/jupyter-book/mystmd/issues/2802

### DIFF
--- a/packages/myst-to-react/src/math.tsx
+++ b/packages/myst-to-react/src/math.tsx
@@ -59,7 +59,13 @@ const mathRenderer: NodeRenderer<MathLike> = ({ node, className }) => {
           ref={ref}
           tabIndex={isScrollable ? 0 : undefined}
           role={isScrollable ? 'region' : undefined}
-          aria-label={isScrollable ? (node.enumerator ? `Equation ${node.enumerator}` : 'Mathematical Equation') : undefined}
+          aria-label={
+            isScrollable
+              ? node.enumerator
+                ? `Equation ${node.enumerator}`
+                : 'Mathematical Equation'
+              : undefined
+          }
           dangerouslySetInnerHTML={{ __html: node.html }}
           className="flex-grow overflow-x-auto overflow-y-hidden"
         />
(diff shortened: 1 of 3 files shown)


agoose77:
Thanks @pancakereport!

FYI to avoid the https://github.com/jupyter-book/myst-theme/pull/860#issuecomment-4285848883 complaint, you can run `npm run changeset` to generate a changeset file. Most changes, especially a11y that don't touch the interface, are `patch` changes.

packages/myst-to-react/src/math.tsx
```diff
@@ -54,6 +56,16 @@ const mathRenderer: NodeRenderer<MathLike> = ({ node, className }) => {
...
+          ref={ref}
+          tabIndex={isScrollable ? 0 : undefined}
+          role={isScrollable ? 'region' : undefined}
+          aria-label={
+            isScrollable
+              ? node.enumerator
+                ? `Equation ${node.enumerator}`
+                : 'Mathematical Equation'
+              : undefined
+          }
```
agoose77, line 68: This pattern feels like something we can generalise at some point. Let's not do it now though!

rowanc1 (approved):

pancakereport:
> FYI to avoid the https://github.com/jupyter-book/myst-theme/pull/860#issuecomment-4285848883 complaint, you can run npm run changeset to generate a changeset file. Most changes, especially a11y that don't touch the interface, are patch changes.

Just to check the workflow on this, it should only be run one time after all the work for a PR has been done, right?

rowanc1:
You can run changesets any time, it generates a markdown file that lists the specific packages that need to be bumped, and the release note that should be added.

Once you create it, you can edit it like normal in another commit, or delete it if you no longer change that package. You can also have multiple changeset files that target different sets of changes if you are working in multiple packages in the same PR.

Basically whenever and however you make them, they should be present before we merge the PR. The github comment above will state the list of package changes once they are found.

In this case the myst-theme bumps (almost) all packages at once, so your changeset will trigger a bump to all packages.

PR #863, opened 2026-04-21
Allow navbar to grow in height

### LINKED ISSUE #862: Long navbar entries cause it to expand partially but overflow
If you have a lot of navbar entries, then the entry links will spill onto a second line, but the navbar won't grow to accommodate it. Also it looks like lines don't have any whitespace between them if this happens (see the buttons)

Image

choldgraf:
This sets the height of the navbar to `min-h-` instead of `h-` so that it can grow if the content inside grows. It also gives it a bit more padding and makes sure it doesn't get squeezed if the navbar grows.

This doesn't totally fix the problem of content overflow, because some downstream stuff still assumes that the header height is a fixed amount. So I added a comment w/ a TODO for what we'll need to do eventually just to avoid blocking this incremental bugfix.

With this PR (an extreme example to show the navbar grows and the buttons look their normal size):

CleanShot 2026-04-27 at 09 49 04@2x

---

- closes #862

### DIFF
--- a/packages/site/src/components/Navigation/ThemeButton.tsx
+++ b/packages/site/src/components/Navigation/ThemeButton.tsx
@@ -1,6 +1,5 @@
 import { useThemeSwitcher } from '@myst-theme/providers';
-import { MoonIcon } from '@heroicons/react/24/solid';
-import { SunIcon } from '@heroicons/react/24/outline';
+import { MoonIcon, SunIcon } from '@heroicons/react/24/solid';
 import classNames from 'classnames';
 
 export function ThemeButton({ className = 'w-10 h-10 mx-3' }: { className?: string }) {
@@ -8,15 +7,15 @@ export function ThemeButton({ className = 'w-10 h-10 mx-3' }: { className?: stri
   return (
     <button
       className={classNames(
-        'myst-theme-button theme rounded-full aspect-square border border-stone-700 dark:border-white hover:bg-neutral-100 border-solid overflow-hidden text-stone-700 dark:text-white hover:text-stone-500 dark:hover:text-neutral-800',
+        'myst-theme-button theme shrink-0 rounded-full border border-stone-700 dark:border-white hover:bg-neutral-100 border-solid overflow-hidden text-stone-700 dark:text-white hover:text-stone-500 dark:hover:text-neutral-800',
         className,
       )}
       title={`Toggle theme between light and dark mode`}
       aria-label={`Toggle theme between light and dark mode`}
       onClick={nextTheme}
     >
-      <MoonIcon className="myst-theme-moon-icon h-full w-full p-0.5 hidden dark:block" />
-      <SunIcon className="myst-theme-sun-icon h-full w-full p-0.5 dark:hidden" />
+      <MoonIcon className="myst-theme-moon-icon h-full w-full p-[18%] hidden dark:block" />
+      <SunIcon className="myst-theme-sun-icon h-full w-full p-[18%] dark:hidden" />
     </button>
   );
 }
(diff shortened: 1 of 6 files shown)


agoose77:
In the long run, I *think* I'd prefer to wrap the trailing items in some kind of dropdown. This is a good incremental fix in the meantime.

agoose77:
I've looked at this before — something we need to do is handle the icon distortion that happens (see the theme selector).

choldgraf:
I agree with all of your takes! I've pushed a commit that should fix the "theme button gets squeezed" issue. For the navbar links themselves, I opened this issue to track that:

- https://github.com/jupyter-book/myst-theme/issues/866

packages/site/src/components/Navigation/ThemeButton.tsx
```diff
@@ -1,22 +1,21 @@
 import { useThemeSwitcher } from '@myst-theme/providers';
-import { MoonIcon } from '@heroicons/react/24/solid';
-import { SunIcon } from '@heroicons/react/24/outline';
+import { MoonIcon, SunIcon } from '@heroicons/react/24/solid';
```
choldgraf, line 2: I'm not sure why these were pulling from different styles before, but I figured it would be simpler / more visually consistent if we used the same style. Happy to change it back if it's too much of a change

packages/site/src/components/Navigation/TopNav.tsx
```diff
@@ -128,8 +128,12 @@ export function TopNav({
   const config = useSiteManifest();
   const { title, nav, actions } = config ?? {};
   const { logo, logo_dark, logo_text, logo_url, logo_alt } = config?.options ?? {};
+  // TODO: when the nav wraps to multiple lines the header grows past 60px, but
+  //   there are downstream consumers of DEFAULT_NAV_HEIGHT and this will
+  //   cause a mismatch if the navbar grows. Here we set it to `min-h` to let
+  //   it grow, but we'll need to revisit downstream height consumers eventually.
   return (
-    <div className="myst-top-nav bg-white/80 backdrop-blur dark:bg-stone-900/80 shadow dark:shadow-stone-700 p-3 md:px-8 sticky w-full top-0 z-30 h-[60px]">
+    <div className="myst-top-nav bg-white/80 backdrop-blur dark:bg-stone-900/80 shadow dark:shadow-stone-700 p-3 md:px-8 sticky w-full top-0 z-30 min-h-[60px]">
```
choldgraf, line 136: this is the part that makes the navbar grow

packages/site/src/components/Navigation/ThemeButton.tsx
```diff
@@ -1,22 +1,21 @@
...
 import classNames from 'classnames';
 
 export function ThemeButton({ className = 'w-10 h-10 mx-3' }: { className?: string }) {
   const { nextTheme } = useThemeSwitcher();
   return (
     <button
       className={classNames(
-        'myst-theme-button theme rounded-full aspect-square border border-stone-700 dark:border-white hover:bg-neutral-100 border-solid overflow-hidden text-stone-700 dark:text-white hover:text-stone-500 dark:hover:text-neutral-800',
+        'myst-theme-button theme shrink-0 rounded-full border border-stone-700 dark:border-white hover:bg-neutral-100 border-solid overflow-hidden text-stone-700 dark:text-white hover:text-stone-500 dark:hover:text-neutral-800',
```
choldgraf, line 10: this will force the theme button to not shrink or get squeezed

choldgraf:
I just realized that we _do_ already make it possible to have dropdowns in those nav links. So I've added an example and docs about that. I'll merge this!

PR #864, opened 2026-04-22
Make cell outputs focusable if they aren't stdin/stderr

### LINKED ISSUE #859: Code cell outputs that overflow are not scrollable (non stdin / stdout)
Note: I originally filed this issue specifically looking at code outputs that are tables. This has since expanded to anything that isn't a standard output (basically a print) or a standard error (i.e. not covered by https://github.com/jupyter-book/myst-theme/pull/849)

Wide regions that have a scroll bar need to be accessible to keyboard only users. I found that this is not already happening for wide table/data frame output from code, as seen [here](https://ds100.org/course-notes/eda/) after 
```
congress_df = pd.DataFrame(congress_json['members'])
congress_df.head()
```

This is also a problem for other kinds of wide output, like with plotly.

I tried updating packages/jupyter/src/outputs.tsx without success (it wasn't able to recognize isScrollable; only hardcoding worked to add the tabindex="0").

To try and disambiguate what's going on, I've added examples with wide table output, wide plotly output, and multiple outputs in #864. In the case of multiple cell outputs, wide regions are separately scrollable. If only one is wide, only that one is scrollable; if both are scrollable, they are separately scrollable. However, both are housing in the outer div with the following elements
```
div.max-w-full.overflow-y-visible.overflow-x-auto.m-0.group.not-prose.relative.text-left.mb-5
```

I'm wondering if the overflow-x-auto needs to be moved into a nested div? And I don't understand what's going on with the thebe stuff.

### References

- https://github.com/jupyter-book/myst-theme/pull/849 fixes this for `stderr` and `stdout` but generic other cell outputs are still broken.

pancakereport:
Does two things:

- Adds a new `MutationObserver` callback to check cell output size and inject scrollability metadata if so
- Documents some of the jupyter cell output functions we've got for better maintainability

This should make the cell outputs more keyboard accessible friendly if they scroll

---
Closes #859

_note this has been edited a lot by @choldgraf ! So if anything breaks don't blame silas blame him :-D_


choldgraf:
OK I've pushed a few changes to this branch that _should_ fix this:

1. I improved the docstrings for our output renderers w/ Jupyter. I found it pretty confusing to figure out what was doing what, so this was an attempt at making this easier to understand for a dev in the future. That's here: [`82c900a` (this PR)](https://github.com/jupyter-book/myst-theme/pull/864/commits/82c900aa2e6dfccb440279f16a59d8bf68bdd869)
2. I added the fix for a11y-friendly cell outputs here: [`ab52b06` (this PR)](https://github.com/jupyter-book/myst-theme/pull/864/commits/ab52b0657541c67432cd41423d34341ab1f9dd08)

I think this should fix it, let's look at the docs preview to confirm though, once it's done

choldgraf:
the behavior looks good to me, [test it out here](https://deploy-preview-864--myst-theme.netlify.app/reference/computation/#wide-cell-inputs-and-outputs)

pancakereport:
I agree that it _looks_ good, in that I can tab into all the cell outputs, but actually running the axe devtools browser extension on the preview is still flagging the table output:
Screenshot 2026-04-23 at 9 37 49 AM

I'm also not immediately seeing the `tabIndex` for these in the html (but maybe it's just kinda early for me and it really is there).

choldgraf:
oh interesting - I didn't actually test if tabIndex was there, I just tested if I could actually tab in via the keyboard and scroll via the arrow keys. Can you explain why that _would_ be possible but it would still not be accessible? That could help us debug this more

pancakereport:
I wish I had manually tested, because your changes look a lot like the changes I tried initially (where I couldn't see `tabIndex="0"` unless I hardcoded it). 

As for why it might be accessible even without `tabIndex="0"`, there are some regions that are automatically focusable--like links, buttons, and ARIA landmarks. I think there's a new aria label there: `aria-label: "cell output"`. And note, that ARIA landmarks and labels are different.

To be accessible, I believe it needs both the `aria-label` and `tabIndex="0"`

choldgraf:
OK I think this one might be an upstream thing, because **thebe** is what controls the scrollability of each output. I opened an issue there:

- https://github.com/jupyter-book/thebe/issues/794

One thing we could do is explicitly *over-ride* that scrollability, and in this theme, make the whole cell output area scrollable. I think that would let us add the `tabIndex` on our own.

I've updated this to do two things:

- Over-ride the CSS that thebe provides, so that we can make scrolling happen at the output area level, not the individual output level.
- Added a MutationObserver to the cell outputs, so that we re-compute whether we need to make it scrollable as the content inside changes, because thebe seems to change stuff asynchronously.

I think that this should solve the problem, but I can't tell if this is too hacky or conservative - we're going to hit a lot of mutationobserver events, I suspect, but maybe that's not a huge deal since the function they call is very simple? @stefanv what do you think?

edit: I **do** now see the tabIndex:

CleanShot 2026-04-23 at 12 32 23@2x

pancakereport:
Yes, is this now passing the tests. If/when a change is made to thebe, will this approach end up making things focusable twice? Would we expect this to slow down sites with a large number of cell outputs (or in other instances)?

pancakereport:
Looks like the multi-output is now scrollable in one. This is fine (and probably correctly because that's the div where the overflow tag is mentioned) but I did kind of like having them separately scrollable. (And of course, I prefer accessible over separately scrollable.)

choldgraf:
OK I _think_ that this might work to keep the outputs scrollable individually, instead of making the whole area do it. In this case we're adding a dedicated `isScrollable` style function just for the thebe outputs. I think that's safe because thebe is the only way that we render cell outputs. This means we don't have to "over-ride" thebe behavior, but we are special-casing it in the JS. I tested it locally and this got me per-cell output scrolling that was focusable. Does it work for you?

### Alternatively we could just add `tabIndex` to each cell output

All of this complexity is just to dynamically only label some cell outputs as focusable when they scroll. I think the simplest thing would be to just always add tabIndex. It means we never accidentally miss one, though it'd be a minor UX annoyance (arguably) if all of them could get focus state.

So is it better to add `tabIndex` to all of them so we never miss one, or dynamically add `tabIndex` but probably missing some now and then?

pancakereport:
It's my understanding that you shouldn't add tabIndex to a region that is NOT scrollable, so I think dynamically adding tabIndex (even if it misses some, which we can hopefully catch and address at some point) is the right way to go. 

Always adding tabIndex isn't necessarily a WCAG violation, but it could confuse and frustrate keyboard users.

pancakereport:
The deploy preview looks good to me! (each output is scrollable independent of other outputs)

choldgraf:
OK I realized that this might break when thebe is _active_ even though it doesn't break when it's not active. I updated this so that we:

- Check for cell scrollability on initial load
- Re-check when thebe executes a cell and it becomes active

I think this means that we'll cover all of our bases here, and we're using a thebe event hook so it should happen more dependably. I merged in a PR to let us test out Binder workflows here, so to test out the behavior, do this:

- Go here: https://deploy-preview-864--myst-theme.netlify.app/reference/computation/#wide-cell-inputs-and-outputs
- Confirm that both the dataframe and the plotly output are scrollable (for me they are)
- Click the thebe button (little power button in the top right)
- When it's done executing, confirm _again_ that the plotly plot is scrollable.
  - The dataframe won't be, because it seems like thebe is rendering the cell output differently with a live kernel so it won't scroll. I don't know what's up with that, but it feels out of scope for this. 

I think that we are good to go if this seems correct on the user side.

_Yes you will see the bug https://github.com/jupyter-book/jupyter-book/issues/2614 too but that's taken care of in #869_

pancakereport:
Looking at the output after active execution is confusing to me. 

The output for stdout and error isn't scrollable, the two TypeErrors and printed "MyST is really, really..." are flagged for me. 

I'm also confused because the wide plotly example at the very end isn't getting a `tabIndex=0` (when I CTRL-F for tabindex the last one I see is the cell that generates that output) BUT I can tab in/out of it and axe isn't flagging it.

choldgraf:
OK I think I figured out what was going on - there was another CSS rule that was over-riding the auto-scroll behavior of the thebe cells. I _think_ the latest commit fixes it, try again?

pancakereport:
What I'm seeing now, both with and without execution:
First pandas output doesn't seem to match the code. Where are the rest of the columns?
Screenshot 2026-04-27 at 9 28 14 PM

TypeErrors and printed "MyST is really, really..." have `tabIndex=0` as they should. And wide pandas output does have `tabIndex` as appropriate (the one that is showing as wide has it; idk what's happening above). 

I don't see `tabIndex=0` for plotly, not when I make my window small so both should be scrollable and not normally (only the one in multiple inputs should be scrollable). But I can tab into both. I actually think you can tab into the single plotly when it ISN'T scrollable (ghost stop). The other thing with this is that the screen doesn't automatically scroll the page so the plotly output is shown on the page, I can scroll back and forth but I can barely see the region so that seems like a problem to me.

choldgraf:
re: that dataframe example, I think you might be looking at the first example on the page, not the last. The first doesn't show the wide output (note the `iloc[:, :5]`). How does the second look?

So it sounds like the big remaining issue is the plotly one then? I have been tab-ing into things and haven't been reliably checking for `tabIndex` so I think that is part of what I've missed here

choldgraf:
OK I just swapped this to use a `MutationObserver` on the output area, this is a bit "expensive" because it'll run any time the contents of the output change, which might be a lot for an interactive widget, but the callback it uses is very simple, and I think this means it'll arbitrarily add/remove the `tabIndex` pretty reliably. I couldn't figure out a way to reliably insert scrollability for interactive content like Plotly using my "thebe event" approach.

Test it out again @pancakereport ?

pancakereport:
Okay yeah, missed the `iloc`. 

The preview seems to be good. `tabIndex` added appropriately, and tests pass

docs/reference/computation.md
```diff
@@ -16,12 +16,14 @@ The [Example Outputs Site](https://jupyter-book.github.io/example-outputs/) has
...
 
+The following cell shows the output of a scrollable table.
+
 ```{code-cell}
 import pandas as pd
 import numpy as np
 
-df = pd.DataFrame(np.random.randn(10, 4), columns=['A', 'B', 'C', 'D'])
-df
+df = pd.DataFrame(np.random.randn(10, 26), columns=['A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M', 'N', 'O', 'P', 'Q', 'R', 'S', 'T', 'U', 'V', 'W', 'X', 'Y', 'Z'])
```
choldgraf, line 25:
```suggestion
df = pd.DataFrame(np.random.randn(10, 26), columns=['A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M', 'N', 'O', 'P', 'Q', 'R', 'S', 'T', 'U', 'V', 'W', 'X', 'Y', 'Z'])
# Here we'll just display the first few columns, we demonstrate wide tables below
```

pancakereport:
Is there anything more to do here or is this ready to be reviewed and merged?

choldgraf:
I went ahead and merged this - I took another pass at the code and think it looks safe, the outputs looked good so let's iterate from there if there is more to fix!

PR #872, opened 2026-05-10
Fix vertical height for stacked buttons

### LINKED ISSUE #871: buttons don't have vertical padding when they stack
An example from the snap documentation at UC Berkeley 

Image

choldgraf:
Sets button `role` items to be inline-block so that their height is respected for the line they're on, and gives a consistent margin so they stack / are spaced consistently.

There's a trade-off here, which is that moving to `inline-block` means that the button will create extra whitespace if it's used with inline text (see before/after below). It feels like a reasonable trade-off since this is really a "UI component" not a piece of text. I'm curious if others disagree though!

| Before | After |
|--------|-------|
| CleanShot 2026-05-10 at 15 42 49@2x | CleanShot 2026-05-10 at 15 42 54@2x |

---

- closes #871

### DIFF
--- a/.changeset/bright-hairs-provide.md
+++ b/.changeset/bright-hairs-provide.md
@@ -1,0 +1,5 @@
+---
+"@myst-theme/styles": patch
+---
+
+Fix vertical height for stacked buttons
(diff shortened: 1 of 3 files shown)


choldgraf:
We discussed this in the collab cafe and everybody was +1 on the tradeoff, so I'll merge this!

PR #875, opened 2026-05-14
Fix hydration mismatch on static builds with BASE_URL

chad-earthscope:
Similar to [mystmd issue 2851](https://github.com/jupyter-book/mystmd/pull/2851
), sites built with a BASE_URL display a white blink (flash) when switching pages.

When the static HTML is built, it first writes the HTML, and when a user loads it in the browser React compares the new DOM to the HTML version initially loaded. The on-disk HTML does not have baseurl in it, while the client version did, and this would cause mis-matches and flickers. This PR strips the baseurl from the client version, which makes it properly match to the on-disk version.


packages/site/src/components/Navigation/TableOfContentsItems.tsx
```diff
@@ -44,13 +44,19 @@ function nestToc(toc: Heading[]): NestedHeading[] {
...
+  // does not include the site's `baseurl`; on the client it does (Remix does
+  // not strip the basename here).
+  //
+  // Also strip any trailing slash, since static builds end up with one and
+  // `heading.path` is slashless.
+  let normed = pathname.endsWith('/') ? pathname.slice(0, -1) : pathname;
+  if (baseurl && normed.startsWith(baseurl)) normed = normed.slice(baseurl.length);
+  const headingPath = heading.path;
+  if (normed && headingPath === `${normed}/index`) return true;
+  return headingPath === normed;
```
agoose77, line 59: This looks great!

So the change is, rather than trying to build a base URL relative path, we try to produce a base URL absent path? Great idea.

choldgraf (approved):
This looks good, I updated the PR body to include a little explanation, since that's what @stefanv was looking at in a recent PR and I think it'll get into the commit log :-) thanks @chad-earthscope !

parmentelat:
for what it's worth, I have tried this in the wild and had to backpedal because apparently it was causing the infamous flickering syndrom

so I can't be 100% positive this is the culprit, but in my use case
- I see a flicker when using a custom theme built on top of aac7f50 aka main
- this goes away if I the same customization on top of be74efb aka 1.3.0

in my case edit: ~~there's no base_url~~ this site has a `base_url`; it's published as a static html build, not an app
surprisingly enough some other of my sites, still without a base_url, did not experience the same regression

unfortunately I'm not quite comfortable opening the content yet as it's work in progress, and comes with confidential data; but just in case this resonates with someone else's experience...

choldgraf:
@parmentelat another candidate might be #884 - that removed a smoother transition and replaced it with a hard one, which might show up as a flicker?

One way you could test this is whether dark mode is an important part of experiencing this flicker...

parmentelat:
I can confirm the flicker comes in dark or light mode

I need to amend my initial statement though (too many repos...):
the site in question does have a base_url set when built for publication, through an `export BASE_URL=/p25`
and the sites that I said were running fine don't

so there's that..

choldgraf:
So in your case,

- sites that do not have baseurl worked fine before and fine now.
- sites that do have a baseurl worked fine before and now show a flicker on main

Is that right? It would be really helpful to nail down the specific scenarios where this breaks

parmentelat:
I'm probably going to need some more time to gather further data, but so far yes, that's what I think I have been witnessing

choldgraf:
Tracking this here:

- https://github.com/jupyter-book/myst-theme/issues/889

PR #878, opened 2026-05-22
soft code THEME_REPO_OWNER

joequant:
This change soft codes THEME_REPO_OWNER in the Makefile to make it easy to generate custom themes.


rowanc1:
Thanks!

PR #882, opened 2026-05-23
Confirm #1 no longer a problem

### LINKED ISSUE #1: Mermaid diagrams use duplicate IDs
You can not cross-reference the diagram on the same page, it disappears, likely because it uses the same ID for the cross-ref.

image

choldgraf:
This is just a little docs PR to confirm that #1 is no longer an issue. I tried it out and it all seems to appear fine, so assuming this is now fixed!

---

- closes #1

### DIFF
--- a/docs/reference/blocks.md
+++ b/docs/reference/blocks.md
@@ -125,12 +125,17 @@ $$
 
 ## Mermaid
 
-```mermaid
+```{mermaid}
+:label: my-mermaid
 graph LR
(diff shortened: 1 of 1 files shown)


rowanc1:
Starting from issue #1 and working through them?!!!?!

Thanks.

choldgraf:
Ha it popped up in a search for mermaid when I was working on the dark mode PR 🙂

PR #883, opened 2026-05-27
Syntax highlight TOML

stefanv:
In highlight.js, TOML/ini are interchangeable

/cc @henryiii

### DIFF
--- a/.changeset/shaggy-ghosts-hide.md
+++ b/.changeset/shaggy-ghosts-hide.md
@@ -1,6 +1,6 @@
 ---
-'myst-to-react': minor
-'@myst-theme/site': minor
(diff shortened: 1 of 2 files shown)


choldgraf (approved):
Makes sense to me - I'll approve and can you merge as long as you can confirm this fixes the problem for you in testing it?

.changeset/shaggy-ghosts-hide.md
```diff
@@ -0,0 +1,6 @@
+---
+'myst-to-react': minor
+'@myst-theme/site': minor
```
rowanc1:
```suggestion
'myst-to-react': patch
'@myst-theme/site': patch
```

PR #884, opened 2026-05-27
Prevent theme change animation to avoid confusing transitions

stefanv:
~Hide the page until a theme class (.light or .dark) is set on <html>, so the body never renders in the wrong theme.~ 

The bit above was reverted, this just does the following:

Remove the body transition-colors that was animating the light→dark switch.


packages/site/src/components/theme.tsx
```diff
@@ -15,5 +15,14 @@ export function BlockingThemeLoader({ useLocalStorage }: { useLocalStorage: bool
...
 `;
 
-  return <script dangerouslySetInnerHTML={{ __html: CLIENT_THEME_SOURCE }} />;
+  // The <style> hides the page until the theme class (.light or .dark) is set on <html>.
+  // The blocking <script> sets that class synchronously before the browser paints, so the
+  // page is never visibly hidden — but if anything temporarily removes the class (e.g. a
+  // React hydration edge case), the page goes invisible rather than flashing the wrong theme.
+  return (
+    <>
+      <style dangerouslySetInnerHTML={{ __html: 'html:not(.light):not(.dark){visibility:hidden}' }} />
```
rowanc1: Would this mean that if JS doesn't load we would see nothing?

stefanv: Not sure if we'll get any content without JS anyway, but we can revert this part of the patch—the removal of the transition should, by itself, help.

rowanc1 (approved):

choldgraf (approved):
I updated the PR body to reflect the new diff, because I was a bit confused at first given the logic we'd removed. I tried the demo and it worked OK (no slow transition of theme), so I'm merging this!

PR #886, opened 2026-05-29
Allow env vars to be passed to `npm run theme:*`

stefanv:
Otherwise, you can't do

CONTENT_CDN_PORT=8000 npm run theme:book


choldgraf (approved):
makes sense to me!

PR #891, opened 2026-06-02
Add a @main workflow for releases

### LINKED ISSUE #890: Publish a `myst-to-react@main` release that we update with theme assets for each commit
In https://github.com/jupyter-book/myst-theme/pull/659 we added a github workflow that builds the theme assets and attaches them to new releases. It would also be useful if users could (optionally) pin their themes to `main`, so that they can try bleeding-edge functionality. This could be a way give us more canaries to try out the theme and report problems, regressions, feedback ,etc.

To do this, we should be able to re-use the same workflow infrastructure, and create a new process that:

- triggers on every commit to `main`
- *builds* release assets using `main`
- *updates* the assets attached to a special `myst-to-react@main` release (without publishing a _new_ release)

Then we can document this as a lightweight "use at your own peril, this is a way to try the latest version of the myst-theme" workflow for power users

choldgraf:
Follow-up from the collab cafe discussion. This adds an extra branch to our "release assets" workflow that updates a `myst-to-react@main` release with newly-built assets. It runs on each commit to main, so the assets are re-built each time we merge.

Similar to the PR we just merged, I'm not sure how to test this out other than to just merge it and see if it works. Any objections?

---

- closes #890

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -99,3 +99,14 @@ jobs:
     with:
       tag_name: ${{ needs.get-tag.outputs.tag_name }}
     secrets: inherit
+
+  # Refresh the rolling `myst-to-react@main` prerelease on every push to main,
+  # so power users can pin their themes to the bleeding edge. Runs regardless of
+  # whether a version was published. See README "Track the latest main".
+  refresh-main-assets:
+    name: Refresh myst-to-react@main assets
+    uses: ./.github/workflows/theme-assets.yml
+    with:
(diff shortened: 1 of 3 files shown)


parmentelat:
you might be able to increase your test surface by using a fork
within a fork there are limitations with secrets of course, but at first sight testing this change may be possible..

agoose77:
I acknowledge @parmentelat's suggestion, but I think I'm just going to merge this in the interest of moving things forward!

PR #892, opened 2026-06-03
Document static files behavior

choldgraf:
This documents the static files behavior so we can confirm how it works! can @agoose77 or @stefanv give a quick look?

### DIFF
--- a/docs/permalinks.md
+++ b/docs/permalinks.md
@@ -92,7 +92,7 @@ How you link a file determines its URL. See the [MyST downloads guide](https://m
   - [/downloads/example-data.csv](/downloads/example-data.csv)
 :::
 
-A static link only works if its path doesn't match a source file. Declare the file under `static_files` in `myst.yml`, then link to the resulting URL. The live app redirects such unmatched routes to the served file.
+A static link only works if its path doesn't match a source file. Declare the file under `static_files` in `myst.yml`, then link to the resulting URL.
 
 Static files are copied to the site root. To link to them, you can provide a link relative to root: `/my_file.pdf` (no need to include `BASE_URL`). Folder structure is preserved, so if `my_folder` is added to `static_files`, you can link to `/my_folder/my_file.pdf`. This is the configuration powering the examples above:
 
--- a/docs/permalinks.md
+++ b/docs/permalinks.md
@@ -63,3 +63,41 @@ For example:
 
 You can match an exact domain (e.g. `mystmd.org`) or use a wildcard to match a single subdomain level. Matches will only be for that subdomain level, not deeper ones, and will not match the root domain (e.g. `*.mystmd.org` matches `docs.mystmd.org` but not `a.b.mystmd.org` or `mystmd.org`).
 
+## Linking to static files
+
+How you link a file determines its URL. See the [MyST downloads guide](https://mystmd.org/guide/website-downloads) for the full reference.
+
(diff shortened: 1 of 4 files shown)


docs/permalinks.md
```diff
@@ -63,3 +63,41 @@ For example:
...
+:::{list-table} Static file link behavior
+:header-rows: 1
+
+* - Link structure
+  - URL you get
+  - Notes
+  - Example
+* - `` {download}`path/to/file.csv` ``
+  - `/file-<hash>.csv`
+  - Cache-busted; URL changes with the file
```
stefanv:
```suggestion
  - Hashed so that URL changes with the contents of the file
```

stefanv: Cache-busting is technical jargon that I think a lot of people don't know about.

docs/permalinks.md
```diff
@@ -63,3 +63,41 @@ For example:
...
+  - `/file.csv`
+  - Stable (parent folders dropped)
+  - [/standalone.csv](/standalone.csv)
+* - `[text](/folder/file.csv)` - `static_files` **folder** entry
+  - `/folder/file.csv`
+  - Stable (folder name kept, parent dropped)
+  - [/downloads/example-data.csv](/downloads/example-data.csv)
+:::
+
+A stable link only works if its path doesn't match a source file. Declare the file under `static_files` in `myst.yml`, then link to the resulting URL. The live app redirects such unmatched routes to the served file.
```
stefanv:
```suggestion
A static link only works if its path doesn't match a source file. Declare the file under `static_files` in `myst.yml`, then link to the resulting URL. The live app redirects such unmatched routes to the served file.
```

You probably did mean "stable" here, but I don't think it's clear from the context what "stable" means.

docs/permalinks.md
```diff
@@ -63,3 +63,41 @@ For example:
...
+
+Each entry is copied to the site root by its **basename**, so parent folders are dropped. You do not need to add a `BASE_URL`, just provide a link relative to your site root. This is the configuration powering the examples above:
```
stefanv:
```suggestion
Static files are copied to the site root. To link to them, you can provide a link relative to root: `/my_file.pdf` (no need to include `BASE_URL`). Folder structure is preserved, so if `my_folder` is added to `static_files`, you can link to `/my_folder/my_file.pdf`. This is the configuration powering the examples above:
```

stefanv (approved):

docs/permalinks.md
```diff
@@ -63,3 +63,41 @@ For example:
...
+  - `/file.csv`
+  - Stable (parent folders dropped)
+  - [/standalone.csv](/standalone.csv)
+* - `[text](/folder/file.csv)` - `static_files` **folder** entry
+  - `/folder/file.csv`
+  - Stable (folder name kept, parent dropped)
+  - [/downloads/example-data.csv](/downloads/example-data.csv)
+:::
+
+A static link only works if its path doesn't match a source file. Declare the file under `static_files` in `myst.yml`, then link to the resulting URL. The live app redirects such unmatched routes to the served file.
```
stefanv:
```suggestion
A static link only works if its path doesn't match a source file. Declare the file under `static_files` in `myst.yml`, then link to the resulting URL.
```

I think this is an implementation detail.

stefanv:
Thanks!

PR #893, opened 2026-06-03
Fix a11y check workflow

choldgraf:
This does a few follow-ups:

- Documents the a11y-check workflow
- Makes it not "error" if a11y problems are found, instead it just opens the issue
- Adds permissions so that it can edit issues


choldgraf:
I'm gonna merge this one so that it unblocks the workflow and we can see if it works

PR #901, opened 2026-06-10
Update from upstream

ZeroEcks:
Sync with upstream, rebased by me, and include my commit that adds umami analytics.


choldgraf:
I think this might be opened on accident

ZeroEcks:
🤦‍♀️ sorry! I had three repositories open and opened the PR against the wrong one!

rowanc1:
Really cool to see you taking the theme and modifying it to your needs. As you go we would really like to hear where things were easy, and where you ran into anything that you would like to be easier!!

PR #902, opened 2026-06-10
Build theme assets with bun, to ensure we use bun.lock

stefanv:


stefanv:
Users may be taking this action, so sticking with npm.

stefanv:
No, something is fishy with that workflow; we cannot install using npm. Fine to install the THEME + its dependencies using npm, but NOT the build dependencies, since we only keep a `bun.lock` for those.

stefanv:
Our themes `package.json` files contain:

```
    "prod:build": "bun run prod:copy && bun run build:thebe && bun run build:css && bun run remix build",
```

Where do we generate the npm lock file?

stefanv:
Yeah, so we used to generate this with `npm install`, but `npm install` doesn't honor the `bun.lock`, so this isn't correct AFAICT.

choldgraf:
I'm so confused 😅

stefanv:
Yeah, took me a while to figure it out; should be good now.

stefanv:
To summarize the issue that is being fixed here: the theme was being *built* using npm install, instead of bun install, meaning that we were relying on `package.json` for dependencies, instead of the more restrictive `bun.lock` that matches our CI.

To make the theme installable by users, we must distribute a `package-lock.json` generated by npm. Theme distribution generation is distinct from (a follow-up to) theme building: we take the theme, install *its* dependencies, and capture the package lock. For consistency (and speed), I've made `bun` do that install as well, and then just generated the `package-lock.json` using npm (there's no `bun.lock` at play here, so could also do this step using only npm).

So, the first change (build using bun, not npm) is a fix, the second (generate package-lock.json from a bun-installed `node_modules`) is an optimization.

choldgraf (approved):
Gonna merge this one because it seems to address the problem and also improves our docs a bit! It'll also address an issue that @parmentelat ran into and tried to address in this sibling PR:

- #907

Issue #906, opened 2026-06-22
zip creation on tag depends on bun but doesn't set it up


pollomarzo:
[theme-assets](https://github.com/jupyter-book/myst-theme/blob/main/.github/workflows/theme-assets.yml) expects bun (`npm run build` -> `turbo run build` will want bun) but bun isn't set up in the workflow, so it will fail on tag creation

PR #907, opened 2026-06-22
theme-assets.yml adapted to using bun

parmentelat:
addresses issue #906

when building zip artefacts for publishing through gh releases, we need to install and invoke bun

manually tested: I have successfully used this code to produce a custom theme in my own orga

### DIFF
--- a/.github/workflows/theme-assets.yml
+++ b/.github/workflows/theme-assets.yml
@@ -44,12 +44,17 @@ jobs:
         with:
           node-version: 24
 
+      - name: Setup Bun
+        uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: latest
+
       - name: Install dependencies
-        run: npm install
+        run: bun install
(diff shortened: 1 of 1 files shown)


parmentelat:
I just came across @stefanv branch `built-with-bun` which apparently addresses the same issue, sorry to not have noticed before

I guess it's still interesting to know this has run through on a real release though

choldgraf (approved):
I believe that this only impacts our "upload release artifact to the github release" workflow, and it looks reasonable to me, so I'll merge this in and we can test it at the next release!

choldgraf:
Actually, I just saw your last comment haha. I am going to close this, and merge this one:

- https://github.com/jupyter-book/myst-theme/pull/902

mostly because that one has a bit of extra docs and optimization in it, but thanks for nudging here @parmentelat !

parmentelat:
fair enough - thanks !

PR #912, opened 2026-06-25
Use just system-ui for font

### LINKED ISSUE #911: Fonts are not reliably loaded and sometimes default to a monospace font for prose
Chris diagnosed my problem as MyST using my system font. I think it should not do that -- I have never seen this behavior on a non-MyST-book-theme site.

My original report assumed that book-theme was intentionally using a monospace font. Original report below based on that mistaken understanding :)

---

Of course, I think monospaced fonts are great for certain use cases like data entry or programming, where characters must be aligned in columns.

But for prose, the default case in a book-theme project (i.e. where users have not explicitly delimited code/data with code fences), I would argue that monospace fonts reduce readability and are less accessible than variable space fonts. Prose written in monospace has an "alien" feel as well, it's just uncommon. And for a "book" theme, I believe it's rare to encounter a book written in fixed width typeface. I struggle to think of an argument in support of monospace prose, aside from the history of use in JupyterBook (and lack of affordances for "pinning" the theme), which is a valid argument, but I feel it doesn't overcome the arguments against.

I personally struggle to read pages rendered using book-theme and without exception override the font in CSS immediately on starting a MyST project.

Maybe an anecdote isn't useful :) Some references:

* [APA "Accessible typography" myth \#4](https://apastyle.apa.org/style-grammar-guidelines/paper-format/accessibility/typography): "monospace fonts are never a good choice for readability"
* Study: [Typeface features and legibility research](https://www.sciencedirect.com/science/article/pii/S0042698919301087). "readers with normal vision read Times Roman 5% faster than Courier Bold at moderate print sizes" (this didn't hold for readers with low vision).
* Study: [A Comparison of Popular Online Fonts: Which is Best and When?](https://web.archive.org/web/20051125062841/http://psychology.wichita.edu/surl/usabilitynews/3S/font.htm). Courier was the 2nd-worst performer on reading time, outperformed significantly by Comic Sans (though Courier performed well on perceived legibility, I think that is not a meaningful metric for the prose use case). Courier scored low on subjective measures of personality, elegance, and fun, but scored well on "business-like".  🕴️

choldgraf:
This defines an explicit font for prose, to see if it renders more nicely on ubuntu.

@mfisher87 want to test this out and see if it looks different? (see the netlify preview when it's done).

---

- closes #911

### DIFF
--- a/styles/theme-colors.css
+++ b/styles/theme-colors.css
@@ -79,6 +79,9 @@
     --myst-color-gray: #6b7280; /* gray-500 */
     --myst-color-gray-bg: #f9fafb; /* gray-50 */
     --myst-color-gray-text: #4b5563; /* gray-600 */
+
+    /* Prose font */
+    --myst-font-sans: system-ui, sans-serif;
   }
 
   .dark {
--- a/styles/index.js
+++ b/styles/index.js
@@ -72,6 +72,9 @@ const themeExtensions = {
     'myst-gray-bg': 'var(--myst-color-gray-bg)',
     'myst-gray-text': 'var(--myst-color-gray-text)',
   },
+  fontFamily: {
+    sans: ['var(--myst-font-sans)'],
+  },
   gridTemplateColumns: {
     'article-sm':
       '[screen-start screen-inset-start] 0.5rem [page-start page-inset-start body-outset-start body-start gutter-left-start body-inset-start middle-start] 1fr 1fr [gutter-left-end] 1fr 1fr [gutter-right-start] 1fr 1fr [middle-end body-inset-end body-end gutter-right-end body-outset-end page-inset-end page-end] 0.5rem [screen-inset-end screen-end]',


mfisher87:
Not sure what's up with Netlify. I tested this on mystmd.org/guide using browser dev tools and I still get monospace. I will add another comment to #911 with a proposed resolution.

styles/theme-colors.css
```diff
@@ -79,6 +79,9 @@
     --myst-color-gray: #6b7280; /* gray-500 */
     --myst-color-gray-bg: #f9fafb; /* gray-50 */
     --myst-color-gray-text: #4b5563; /* gray-600 */
+
+    /* Prose font */
+    --myst-font-sans: system-ui, sans-serif;
```
mfisher87, line 84:
```suggestion
    --myst-font-sans: sans-serif;
```

https://github.com/jupyter-book/myst-theme/issues/911#issuecomment-4802992429

mfisher87:
@choldgraf what do you think? :point_up:

choldgraf:
Why don't we use the sans defaults from tailwind? I noticed that doesn't have the system UI either do might solve your program

https://tailwindcss.com/docs/font-family

agoose77:
I've added a query here: https://github.com/jupyter-book/myst-theme/issues/911

If we want to move away from system-ui, upgrading to Tailwind 4 would do that, I think.

mfisher87:
Ahhh I see, this is coming from tailwind v3 (https://v3.tailwindcss.com/docs/font-family). What a strange decision to use `system-ui` 2nd, before `sans-serif`, in their `font-sans` family:

```
font-family: ui-sans-serif, system-ui, sans-serif, ...
```

It's strange that they only fixed this in a major release; I would consider it a bug if the tailwind user specifies sans serif and the end-user gets [pijamas](https://www.fontspace.com/pijamas-font-f11716) :shrug:

choldgraf:
How complex would it be to upgrade to Tailwind v4? I think out of principle it's good to upgrade, but would this either be disruptive to downstream users, or complex to implement ourselves?

PR #915, opened 2026-06-27
Add Jupyter-compatible notebook cell deep-links

Darshan808:
### Description
Renderer half of the fix for jupyter-book/mystmd#518 (compiler half is [in this PR](https://github.com/jupyter-book/mystmd/pull/2975)).

Fixes unstable notebook cell links by using the stable `nbformat` cell ID instead of the per-build `node.key`.

### Changes
- Use `html_id ?? identifier` for notebook cell anchors instead of `node.key`.
- Add support for `#cell-id=<id>` fragments (while keeping `#<id>` working).
- Add a unit test for fragment parsing.


krassowski:
@agoose77 is the netlify build failure expected? I see this on many other PRs but I also know that the related PR https://github.com/jupyter-book/mystmd/pull/2975 is not merged yet so I am not sure if there needs to be some kind of coordination between these two PRs.

krassowski:
It looks like there is "1 workflow awaiting approval" :)

agoose77:
This looks good! We can ship this before the `mystmd` PR, so I'll merge it now.

PR #917, opened 2026-06-30
Clean ups to color variables in CSS and picker following #843

stefanv:
This tidies up the CSS class groupings, fixes a couple of cases where CSS tokens weren't used, improves the color theme experiments page to have a more complete showcase, removes specialized colors (orange, purple, etc.) from proofs, and slightly expand the theming docs.


stefanv:
Need to update color picker.

stefanv:
Still some fixing to do on the color picker page.

stefanv:
Should be in good shape for a review now.

choldgraf (approved):
I like the move towards semantic color schemes rather than using color names, so +1 on that from me. I looked at the built docs and I couldn't find anything there that looked off! The color picker page I'm a little worried about going out-of-date since there is so much hard-coded variable naming etc in there...but I also feel like it's useful to have a lot of clarity there, and I guess we can run an LLM over it now and then to make sure it hasn't drifted from reality too much. IMO this is OK to merge, and we can keep following up with more clean-ups as necessary. WDYT?

stefanv:
Sorry, @choldgraf, I dropped that last commit of yours, but I think you'll like this: we now discover the color picker variables from the loaded page's computed CSS.

choldgraf:
Goodbye little commit, we hardly new ye

choldgraf (approved):

stefanv:
Great, thanks for taking a look; hopefully these changes do not proof to be disruptive. I think any follow-up fixes should be easy, so let's see how it goes.

choldgraf:
Thank you for this clean up!

PR #919, opened 2026-07-01
Honor bun lockfile in CI and Netlify

stefanv:


choldgraf:
does this update the lockfile if we don't use this flag or somethign?

stefanv:
@choldgraf Yes, exactly; which is not what we want in CI.

choldgraf (approved):
LGTM - let's merge and see if that fixes things :-) 

thanks for these cleanups!

stefanv:
Follow-up to fix the underlying issue with the new version of `@jupyterlab/services`: https://github.com/jupyter-book/mystmd/pull/2981

PR #920, opened 2026-07-01
📂 Allow renderer/runtime extensions with Module Fedaration

### LINKED ISSUE #2620: Support MyST Theme node renderer plugins
### Proposal

Similar to #1801, this initiative tracks the work to add plugins to myst-theme that enable users to override renderers of existing (or new) nodes like `text` or `card`.

These plugins are effectively of the form
```tsx
function renderCard(node: Card): Component {
  return <div class="card"><MyST ast={node.children} /></div>
}
```

Custom renderers are specific to each MyST site. As such, they should avoid materially changing the interpretation of nodes such that when embedded in other sites these nodes lose their meaning.

agoose77:
This is a tiny PR that lets MyST load renderers from e.g. https://github.com/myst-contrib/myst-theme-demo-renderer.

> [!Note]
> **TLDR**; Module Federation lets extension authors pre-compile code with dependencies that can integrate into a bigger app. We can share core modules like React, etc.

It uses [Module Federation](https://module-federation.io/), a standard pattern for decomposing monolithic applications into smaller units. There are different versions of MF. This PR uses 1.5 via `rspack`. MF is backwards compatible, so the changes are more about tooling than runtime. 

> [!Important]
> JupyterLab uses Module Federation for its extension system. They were solving exactly the same problem that we have — how to bring in new functionality without recompiling.

Given our tooling in myst-theme is likely to change, I haven't bothered trying to get MF working via a build plugin. Instead, I'm manually loading the runtime in the app. In future, we can use build plugins that make the dynamic loading invisible to the source code (e.g. `import "foo"` just works).

> [!Note]
> Why this approach? The extension side of things is easy — just use `rspack` (or other supported bundlers). The rest is mostly invisible (see https://github.com/myst-contrib/myst-theme-demo-renderer/tree/master/basic-div). 
> We can build small machinery or a copier template to simplify the process.

We could update our docs to replace the anywidget in https://github.com/jupyter-book/myst-theme/blob/babfb2fb5fa2e598442ecd695f3506d75db02880/docs/src/color-picker.mjs#L2 with a react plugin.

Closes https://github.com/jupyter-book/mystmd/issues/2620

### To Do

- Consider how to simplify tooling for plugin authors.
  - We should provide shared package information, for example.
- Consider how to robustly support extensions for both article and book themes.
  - Both apps needs to declare which packages are shared and which are not.
- Consider which deps to share with extensions. 
  - The more we add here, the harder it will be to re-use extensions in different contexts. 
- Confirm that one extension can expose a shared dep to another extension.
  - I think this can be done simply by not specifying `import: false` on the shared object in the provider that brings the implementation. We should test this.


themes/book/app/root.tsx
```diff
@@ -133,10 +135,31 @@ function NoCSSWarning() {
...
+type SetRenderers = (renderers: NodeRenderers) => void;
+
+async function loadRenderPlugins(setRenderers: SetRenderers): Promise<void> {
+  const mf = createInstance({
+    name: 'host',
+    remotes: [
+      {
+        name: 'custom_heading_renderer',
+        entry: 'http://localhost:8081/custom_heading_renderer.js',
+      },
```
agoose77: We should adapt this to query the MyST config and generate an array of entry items.

stefanv:
xref: discussion at today's [working meeting](https://www.google.com/url?q=https://hackmd.io/KZCfkPZyQHObuPLYC6NsMg&sa=D&source=calendar&usd=2&usg=AOvVaw2qK2LWHsnHP7IZGsMD5dFx), with a prototype dynamic render loader in https://github.com/jupyter-book/myst-theme/tree/wip-frontend-renderers (@rowanc1, @stefanv during in-person hacking session)

@agoose77 raised the idea today of using Module Federation (v1.5) instead. I like this idea, because it (a) uses a standard mechanism (b) paves the way for sharing common dependencies like react and (c) aligns with the way Jupyter Lab does it.

Thanks for implementing this POC @agoose77!

stefanv:
@agoose77 The branch I mention above shows how to add renderers to the config for dynamic loading; not much code, but probably very similar to what we need here next.

agoose77:
I've updated this spike to share the built-in singletons like React and the various MyST deps. Here's an example from https://github.com/myst-contrib/myst-theme-demo-renderer:

image

themes/book/app/bootstrap.ts
```diff
@@ -0,0 +1,270 @@
+import type { NodeRenderers } from '@myst-theme/providers';
+import { createInstance } from '@module-federation/runtime';
+
+/**
+ * Generated modules from 
+ * jq '.dependencies | to_entries | map({key: .key, value: {get: "() => import(\"\(.key)\").then(mod => () => mod)", scope: "default", shareConfig: {singleton:true,import:false,requiredVersion: .value | sub("\\^(?<v>.*)"; "~\(.v)")}}}) | from_entries' package.json | wl-copy
+ * This would be handled by the build system typically (invisibly)
+ */
```
agoose77: JupyterLab does this at build-time in a custom webpack / rspack plugin. We would do the same, but I didn't want to work around our existing Remix stack given that we'll be upgrading to react-router soon.

agoose77:
This spike is finished. What's next?

I'd like to find a less manual way to define `bootstrap.js`. Without using a build plugin (not easy with remix I think), we probably need manual imports for the libs but we could resolve the versions of shared modules etc from package.json and the lock file. Maybe that's a pre-build step to compute that information into a single JSON file.

I'd like to support chunked (multi-file) extensions. That means copying them into `static` / `public` under a dir, probably.

I also want to test this / think about how we would move it forward for HTML builds.

Then we need to establish templates for extensions.

I also want to review whether an `effect` is the right place for for this — I feel that we should block the main app whilst we load exts to avoid a flash.

stefanv:
With the plugin branch I mentioned before, we built the files and put them at a known location (via config.yaml) on the CDN. 

Not sure if that is the discovery you mentioned above, or whether there is something else missing?

agoose77:
@stefanv I copied the implementation from your suggested branch to use `config.json`. Now, the theme can consume plugins from the content server.

stefanv:
Very nice, Angus!!

stefanv:
CI failing because of type declarations.

themes/book/scripts/bootstrap.mjs
```diff
@@ -0,0 +1,84 @@
...
+    const constraint = pkgJSON.dependencies[name];
+    return [
+      name,
+      {
+        scope: 'default',
+        version,
+        get: `{{ ${name} }}`,
+        shareConfig: {
+          singleton: true,
+          requiredVersion: constraint.replace('^', '~'),
```
agoose77, line 42: JupyterLab uses this technique — relax constraints to permit plugins to load even if they strictly say they don't work.

themes/book/app/root.tsx
```diff
@@ -133,10 +140,15 @@ function NoCSSWarning() {
...
 
+
 export default function AppWithReload() {
   const { theme, config, CONTENT_CDN_PORT, MODE, BASE_URL } = useLoaderData<SiteLoader>();
 
   const searchFactory = useCallback((index: MystSearchIndex) => createSearch(index), []);
+  const [renderers, setRenderers] = useState(RENDERERS);
+  useEffect(() => {
+          loadExtensions<Extension>((config as any)?.remotes ?? []).then(extensions =>           setRenderers(mergeRenderers([RENDERERS, ...extensions.map(ext=> ext.renderers)])))
+  }, []);
```
agoose77, line 151: The return type of `loadExtensions` is an `Extension` object. This gives us scope to support more kinds of extensions besides those which add new node renderers.

choldgraf:
This is super cool! I love being able to define our own renderers for nodes, and the ability to bring reactiness with us, ability to embed CSS, etc.

I was trying to add a little commit to this PR to show off an example in these docs, but I noticed that your demo is doing some hacky patching to the config here:

https://github.com/myst-contrib/myst-theme-demo-renderer/blob/master/patch.mjs

Is there a way we can get this PR to a place where we don't need hacky config patching in order to demonstrate the functionality in these docs? I think it'd be pretty easy to throw together a little demo here that shows how to over-ride a node, and how to define a new node + define a custom HTML renderer for it

That said, using the same config hack you did, I was able to get this working locally and it is neat :-)

CleanShot 2026-07-04 at 22 18 03@2x

agoose77:
I think there are two ways we could specify these MF extensions.

1. In the environment so that you can `pip install jupyter-book-theme jupyter-book-custom-renderer` and have it Just Work™.
2. In the content, so that the theme can pull it in from the project.

(2) is what this PR implements, and allows the content provider to populate `config.json`. Right now, the `mystmd` engine doesn't know about this, so it can't do anything. 

We could merge e.g. https://github.com/jupyter-book/mystmd/pull/2910 and put this under a template option too.

choldgraf:
My first intuition was that you'd enable this by adding something to the list of plugins, and this would just be another type of thing a plugin can define. Did you explore that approach? Or maybe this would be a renderer-specific plugin and that's why it wouldn't make sense for the core config list?

Either way, that arrays PR seems like a no brainer so if it'd help unblock one pathway here, I can focus some review time there

agoose77:
> My first intuition was that you'd enable this by adding something to the list of plugins, and this would just be another type of thing a plugin can define. Did you explore that approach? Or maybe this would be a renderer-specific plugin and that's why it wouldn't make sense for the core config list?
> 
> Either way, that arrays PR seems like a no brainer so if it'd help unblock one pathway here, I can focus some review time there

Yes — for now, this extension system is outside of the myst plugin system. I see us potentially unifying them down the road (at least, by UX), but that's a bit further away. The main technical reason is that render plugins need to work in different runtimes (e.g. the browser) than the core plugins (Node.js and/or browser, in future).

choldgraf:
That makes sense to me - as long as it's simple to load renderers for authors I'm happy 🙂... It just felt like hand modifying that config file felt a bit much

themes/book/app/bootstrap.ts
stevejpurves: The big plus for me about the modern MF is that we can avoid webpack completely, and still use vite with MF

themes/book/scripts/bootstrap.mjs
```diff
@@ -0,0 +1,84 @@
...
+
+ return await Promise.all(
+        remotes.map(r => (mf.loadRemote(r.name) as Promise<{default: T}>).then(m => m.default))
+
+ );
+}
+
+`;
+
+writeFileSync(path.join(import.meta.dirname, '..', 'app', 'bootstrap.ts'), bootstrap);
```
stevejpurves, line 84: Instead of bootstapping in by manual code injection, has anyone tried to define the host via the vite build system? https://github.com/module-federation/vite

stevejpurves:
@agoose77 in your example https://github.com/myst-contrib/myst-theme-demo-renderer the basic-dev has a long list of shared modules, is that necessary for any module that is built to have the full list? and does that need to be kept in sync with the host? (i'm assuming that not and that was just the boilerplate for the demo?)

stevejpurves:
I'd love for us to be explicit here, just for clarity in the discussion (andd psrry if Inhave missed that elsewhere), what we see as having MF brings above us just having a mechanism loading plain ESM modules in bootstrap.* (which we could do ourselves to achieve the same for basic rendering), and I realize MF has a lot of features/advantages but which ones are actually going to matter for our use case(s)?

Personally, since MF has grow beyond webpack only tooling (yay) i'm positive and I am intrigued/curious on hit can help us use jupyter lab front end code in a more consistent and future proof way (bearing in mind the one-of point solutions we have for plotly and bokeh in the thebe/jupyter passive rendering code and the fact that the iWidgets system use the old legacy notebook extension code).

@agoose would love your take on that

stevejpurves:
Having noderenders flow from our render plugins at build time is an essential that we should work towards imo

themes/book/package.json
```diff
@@ -40,7 +41,9 @@
     "node-fetch": "^2.6.11",
     "thebe-core": "^0.5.0",
     "react": "^19.1.0",
-    "react-dom": "^19.1.0"
+    "react-dom": "^19.1.0",
+    "@module-federation/runtime-tools": "^2.6.0",
+    "@module-federation/runtime": "^2.6.0"
```
stevejpurves, line 46: no webpack! 😅

themes/book/scripts/bootstrap.mjs
agoose77: I haven't tried it here, because we don't currently use Vite! Once we look to upgrade the stack to React Router etc., then we'd replace this bootstrap with a build plugin (as per https://github.com/jupyter-book/myst-theme/pull/920#issue-4789514788)

agoose77:
> @agoose77 in your example https://github.com/myst-contrib/myst-theme-demo-renderer the basic-dev has a long list of shared modules, is that necessary for any module that is built to have the full list? and does that need to be kept in sync with the host? (i'm assuming that not and that was just the boilerplate for the demo?)

This is necessary, but I think only for the _used_ dependencies. My understanding is that the bundler and build plugin need to know about the external deps to avoid bundling them into the core app. 

JupyterLab handles this via a `jupyter-builder` package that extensions use to build themselves. It's basically a way to ship this config without needing extensions to copy-and-paste it. We could also ship _just_ the shared data as a package on NPM as `@myst-theme/build-metadata` and import it in the `rspack` config.

themes/book/package.json
agoose77: :pray:

agoose77:
> I'd love for us to be explicit here, just for clarity in the discussion (andd psrry if Inhave missed that elsewhere), what we see as having MF brings above us just having a mechanism loading plain ESM modules in bootstrap.* (which we could do ourselves to achieve the same for basic rendering), and I realize MF has a lot of features/advantages but which ones are actually going to matter for our use case(s)?
> 
> Personally, since MF has grow beyond webpack only tooling (yay) i'm positive and I am intrigued/curious on hit can help us use jupyter lab front end code in a more consistent and future proof way (bearing in mind the one-of point solutions we have for plotly and bokeh in the thebe/jupyter passive rendering code and the fact that the iWidgets system use the old legacy notebook extension code).
> 
> @AGoose would love your take on that

Yes, Module Federation is very similar to just `import("https://.../")` with some differences:
- Shared dependency injection — extensions ask for React, they get the main react runtime used by the parent app.
- Invisible loading of core dependencies — if we use a build-time plugin, we can in time separate parts of the core into these federated modules, without seeing that in the source code (i.e. we still have `import XXX`). This might conflict with the goal of being able to turn off core features after install, though.

I have some ideas for JupyterLab, `mystmd`, and `myst-cli`:
- https://github.com/jupyter-book/mystmd/issues/2962
- https://github.com/jupyter-book/mystmd/issues/2962

Starting with lab extensions, I _think_ that we can consume the mime renderers _without_ needing the full JupyterLab application framework. MIME renderers e.g. Plotly, are built with Module Federation. Here's [an example](https://github.com/plotly/plotly.py/blob/main/js/src/mimeExtension.ts). 

However, loading e.g. an old extension like plotly might require us to do bits by hand — I can't see how to tell the runtime to load the extension from the `_JUPYTERLAB` top level variable. I think it must be possible, because the rspack ecosystem understands how to populate that entry. I'm also aware that rendermime extensions are sometimes bundled in the same federated module as other lab extensions. I think that might be a pain for us, because we have no way of saying "the renderermime ext doesn't need JupyterLab deps". However, I suspect it's good practice to use separate packages to ship rendermime extensions and proper extensions.

agoose77:
> Having noderenders flow from our render plugins at build time is an essential that we should work towards imo

Are you saying "we should split up myst-theme into extensions that provide plugins"?

I like this approach too. It makes it easy to start with a pared down theme, and add things like cards etc.

agoose77:
What are the next steps here?

- I think we can aim to ship this feature _only_ for the book theme, until we migrate to react-router. 
- We should figure out the UX from the mystmd side. I'd prefer to add template options that copy directories over bespoke plugin awareness.
- We should confirm that HTML builds work.

choldgraf:
When I tested this locally it worked nicely, so my suggestion would be:

1. Ask others if there are any hard technical blockers we must answer before merging.
2. The UX so that authors can load a renderer with one line of config and potentially an install step is one such question to ask. Unless we explicitly want to keep this feature in the realm of " prototype in production for testing purposes" in which case we should document that. 

Is it ok if I push a docs example so we show off what this can enable? I do not think it's realistic to point people to other repositories so I'd prefer if we bundled a tiny docs plug-in to show this off

agoose77:
> When I tested this locally it worked nicely, so my suggestion would be:
> 
>     1. Ask others if there are any hard technical blockers we must answer before merging.
> 
>     2. The UX so that authors can load a renderer with one line of config and potentially an install step is one such question to ask. Unless we explicitly want to keep this feature in the realm of " prototype in production for testing purposes" in which case we should document that.
> 
> 
> Is it ok if I push a docs example so we show off what this can enable? I do not think it's realistic to point people to other repositories so I'd prefer if we bundled a tiny docs plug-in to show this off

It would be nice to have some docs! We're moving over from spike to implementation, so it needs to be done at some point. 

As a user I tend to prefer code repositories because I know that they work — docs examples have a habit of breaking as they're not built and tested. But, I'll take a read from you on your preference. The demo above should be simplified to do the smallest thing possible. That means dropping the anywidget demo.

choldgraf:
Oh I totally agree about having docs that are tested. That's why I want this to build an actual plugin with this pattern, so that we can look at the netlify preview in this PR and confirm that it actually works without needing to look at another repo. In my experience, any time you require more than one repo to test something out, the likelihood it happens drops way down. I can try to put something simple together and show you what I mean here.

agoose77:
> Oh I totally agree about having docs that are tested. That's why I want this to build an actual plugin with this pattern, so that we can look at the netlify preview in this PR and confirm that it actually works without needing to look at another repo. In my experience, any time you require more than one repo to test something out, the likelihood it happens drops way down. I can try to put something simple together and show you what I mean here.

I think I'm following — you're proposing building a plugin in-tree and using it in the docs? I think that's functionally fairly similar to having something in a separate repo and pulling it in to our build — and I'm happy either way. 

As long as there's a directory containing a set of files that a user can build, rather than a series of `.md` files with instructions!

choldgraf:
Yes that's exactly what I propose - I can push something soon to demonstrate. I like it when the source repository has all the information self contained to demonstrate that functionality works - and yes that includes actually building and running things not just copy pasting code snippets!

choldgraf:
Argh I played around with this a little while and only got like 70% where I was hoping to get haha. I wrote up a little docs page and a mini extension example using your repo as inspiration - I also realized we can use `static_files` to copy over the `/dist/` folder so the extension can use it.

However, I realized that we can't use this in `--html` mode because we can't do the `config.json` patching that way, so we can't actually preview this with netlify :-( .  Anyway I figured i'd push the commit + the docs just to have something to iterate on - it works locally via the myst server, just not via static HTML outputs.  I promise that I won't be mad if you decide to just not use it haha

agoose77:
Closing this spike in favour of https://github.com/jupyter-book/myst-theme/pull/939

PR #922, opened 2026-07-08
(1) Upgrade to Remix v2 (classic compiler)

### LINKED ISSUE #828: Use React-Router instead of Remix
I've seen a few discussions about using [React Router](https://reactrouter.com/) - creating this issue to track the topic overall, reasons for doing it, and implementation and design discussion.

I don't think this has been "decided" yet, but it does seem some folks are keen to do it so maybe this can be a space to track the "should we?" discussion as well.

agoose77:
This PR does the smallest amount of work to upgrade to Remix v2. The goal is to iteratively upgrade, so that we can ensure success of moving to react-router.

This doesn't switch to Vite yet.

Part of https://github.com/jupyter-book/myst-theme/issues/828


flake.nix
```diff
@@ -19,6 +19,7 @@
           devShells.default = mkShell {
             buildInputs = [
               pkgs.nodejs_22
+              pkgs.bun
```
agoose77: Just including this Angus-specific dev change here.

PR #925, opened 2026-07-10
(4) Use pre-rendering for SSG in React Router 7

### LINKED ISSUE #828: Use React-Router instead of Remix

agoose77:
Needs https://github.com/jupyter-book/myst-theme/pull/924

### Changes

I removed the session-based theme management. There's no benefit to keeping this IMO, because if we want the theme selector to work on static builds, we can't fully lean in to SSR unless we more deeply split the SSG non SSG pathways. If we keep it, I can't see how to prevent react-router from complaining because `action` is not supported without a server.

> [!Warning]
> I want to think through how we handle the CDN in the theme. With moving pre-rendering to the theme, it would make sense to have the theme do all of the work to rewrite `/static/foo` as `${CONTENT_CDN}/static/foo` where possible. We can turn that on as an option in the live server, and _require_ it for the HTML export.
>
> This is also useful for non-prerendering e.g. to serve CDN behind a firewall. I think we need to add static proxying to the production server (via express), dev server (maybe vite?), and the pre-rendering post-build action.

Part of https://github.com/jupyter-book/myst-theme/issues/828


themes/book/react-router.static.config.ts
```diff
@@ -0,0 +1,72 @@
...
+      path: asset,
+      binary: true,
+    })),
+  ].flat();
+}
+export default {
+  ssr: false,
+  async prerender({ getStaticPaths }) {
+    return [...makeRoutes(config).map((r) => (r.url ? r.url : '/')), ...getStaticPaths()];
+  },
```
agoose77, line 109: We'll want to update this to support SSG and non SSG. This spike sets up only SSG

agoose77:
The intention is to use `bun run render` to build a static site for a particular content server. This would require MyST to invoke a new `template.yml#build.render` script that runs `bun run render` under the hood.

This change would, I think, mean that rather than building a theme package that has a tiny set of dependencies at runtime (`express.js`, and we could just bundle that right now), we build more of a full package. I think that's OK, I'm just observing the shift.

agoose77:
This reminds me of a finding in https://github.com/agoose77/jupyter-book-site-renderer — it would be useful to expose a manifest.json or something that declares the contents of public, so that we only need the CDN URL to build the static site.

agoose77:
I am now looking at relative base URLs, as the final feature

choldgraf:
If we get relative urls out of this I will buy you a beer 😅

agoose77:
@choldgraf it looks like react-router (and every other routing framework) doesn't support relative base URLs. I'm going to give up on it for this work because it's going to be hacky if at all viable.

themes/book/react-router.static.config.ts
```diff
@@ -0,0 +1,78 @@
...
+interface Page {
+  slug: string;
+}
+
+function makeRoutes(data: any) {
+  const localProj = data.projects[0];
+  const baseurl = '';
+  const host = '';
+
+  const localProjSlug = localProj.slug ? `/${localProj.slug}` : '';
```
agoose77, line 47: Can we remove project slugs now?

agoose77:
I started to look at how to rebuild this on top of #925. I ran in to a confusing error message that I've queried upstream, but I did some thinking too.

For SSG, we can use build-time MF to bake in the remotes. Whereas, for the dev server we need to use a dynamic remote.

agoose77:
OK, I asked about SSG in the Discord for React Router: https://discord.com/channels/770287896669978684/1545147103846735882

The true MPA approach requires us to set `ssr: true` and disable all hydration. But, we need hydration for things like `useEffect`. So, we're still going to use `ssr: false` and enable hydration.

We can delete the `.data` files and SPA fallback _if_ we're consistent in our navigation behaviour to use the browser routing. We do that in our theme provider with `reloadDocument`.

PR #934, opened 2026-07-18
Add dep back for plotly builds

### LINKED ISSUE #933: Plotly outputs no longer working on main?
I noticed this on [our theme docs](https://myst-theme.netlify.app/reference/computation/#plotly):

Image

choldgraf:
I think this might fix the plotly issue? It seemed like this was an old dep that got lost when we moved to bun? @stefanv does this ring a bell?

this seems to [make plotly happy again](https://deploy-preview-934--myst-theme.netlify.app/reference/computation/#plotly)

CleanShot 2026-07-18 at 12 18 16@2x

---

- closes #933 (maybe)

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -435,6 +435,7 @@
   "patchedDependencies": {
     "@remix-run/dev@1.17.1": "patches/@remix-run%2Fdev@1.17.1.patch",
     "@jupyter-widgets/controls@5.0.12": "patches/@jupyter-widgets%2Fcontrols@5.0.12.patch",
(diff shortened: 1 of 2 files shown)


stefanv:
Is plotly supported by default in Jupyter outputs, or how does this come about?

choldgraf:
I think we were using the jupyterlab-plotly package for this right? If I recall plotly has its own logic for rendering in various notebook interfaces and this has evolved over time - at one point I think it was that jupyterlab-plotly package but it might be deprecated now

agoose77:
LGTM. I don't think this needs a changeset, it's not a single package change. I think we just need to make sure that the themes/ are released.

PR #935, opened 2026-07-18
Fix TOC navigation in static build, and reduce flashing

### LINKED ISSUE #910: Sidebar TOC section collapses after clicking an index page's title
After moving the scientific-python development guide to MystMD, we ran into a bug: clicking on a page causes the animation for expanding the subpages, but then the new page loads and everything is closed. This makes it hard to navigate to the guide pages unless you know about the little `<`/`v` arrow. This only happens for sites that have pages with subpages; pure subpages without parent pages work fine because it's all really just the same thing as the `<`/`v` button.

I've added the output from a Claude investigation below. The suggested fix from that is:

> :robot: _AI text below_ :robot:
>
> Include the computed value in the dependency array so the effect re-applies once the path settles:
> 
> ```js
> useEffect(() => {
>   if (nav.state === 'idle') setOpen(startOpen);
> }, [nav.state, startOpen]);
> ```
> 
> (This is also more correct generally — any path change that flips `startOpen` should re-sync the collapsible.)

I can make a PR for that if you want.

<details>
<summary>:robot: AI issue text :robot:</summary>

**Title:** Sidebar TOC section collapses after clicking an index page's title (should stay expanded)

**Body:**

### Describe the bug

In the book theme, when a TOC entry has children (an index page with subpages), clicking the **name** of that entry navigates to its index page but leaves the section **collapsed**. You can watch the expand animation start and then the section collapses again once the page loads.

Clicking the **chevron/dropdown arrow** works fine (it just toggles open state with no navigation). The problem is specific to clicking the entry's title link.

Expected: when you're on an index page, its subpages should be expanded by default (and stay expanded after navigating to it).

### To reproduce

1. Build a site with a TOC where a top-level entry has children, e.g.:
   ```yaml
   toc:
     - file: index.md
     - file: guides/index.md
       children:
         - file: guides/a.md
         - file: guides/b.md
   ```
2. From a different page (with the `guides` section collapsed), click the **"guides"** title in the sidebar.
3. The section briefly animates open, then collapses once `guides/index` loads.

Seen on `@myst-theme/site@1.3.1` (book theme).

### Root cause

In [`[packages/site/src/components/Navigation/TableOfContentsItems.tsx](https://github.com/jupyter-book/myst-theme/blob/main/packages/site/src/components/Navigation/TableOfContentsItems.tsx)`](https://github.com/jupyter-book/myst-theme/blob/main/packages/site/src/components/Navigation/TableOfContentsItems.tsx), `NestedToc`:

```js
const startOpen = childrenOpen([heading], pathname, baseurl).includes(heading.id);
const nav = useNavigation();
const [open, setOpen] = React.useState(startOpen);
useEffect(() => {
  if (nav.state === 'idle') setOpen(startOpen);
}, [nav.state]);           // <-- only depends on nav.state
```

The intent is correct — `childrenOpen` / `pathnameMatchesHeading` already match an entry against *its own* index path (including the `…/index` case), so `startOpen` is `true` when you're on the entry's index page.

The problem is the effect's dependency array: it re-runs **only** on the `nav.state` transition, not when `pathname` / `startOpen` change. Clicking the title link optimistically opens the section (`onClick={() => setOpen(heading.path ? true : !open)}` — the animation you see), but if `pathname` updates in a separate commit from the `nav.state → 'idle'` transition, the effect fires with the *stale* (pre-navigation) `startOpen` (`false` for the just-navigated section) and never re-runs to pick up the corrected `true`. So it latches collapsed.

### Suggested fix

Include the computed value in the dependency array so the effect re-applies once the path settles:

```js
useEffect(() => {
  if (nav.state === 'idle') setOpen(startOpen);
}, [nav.state, startOpen]);
```

(This is also more correct generally — any path change that flips `startOpen` should re-sync the collapsible.)

### Related

#635 / #647 covered the static-HTML variant of "sidebar stays collapsed" and are fixed; this is the client-side navigation case, which I don't see tracked in an open issue.

Want me to also draft the one-line PR (fork, patch, conventional-commit message) so you can open it alongside the issue?

</details>

stefanv:
A trailing-slash BASE_URL (e.g. /development/) broke
withBaseurl (double slashes in generated hrefs/favicon/stylesheet
links), and TOC sections never auto-expanded.

Normalize once in BaseUrlProvider and whenever each theme accesses the
environement variable.

Also fix NestedToc's sync effect to compute the open state from
pathname, instead of using the startOpen value.

Closes #910

### DIFF
--- a/packages/site/src/components/Navigation/TableOfContentsItems.tsx
+++ b/packages/site/src/components/Navigation/TableOfContentsItems.tsx
@@ -228,7 +228,10 @@ const NestedToc = ({ heading }: { heading: NestedHeading }) => {
           </button>
         </Collapsible.Trigger>
       </div>
-      <Collapsible.Content className="pl-3 pr-[2px] collapsible-content">
+      <Collapsible.Content
+        className="pl-3 pr-[2px] collapsible-content"
+        style={isStaticBuild ? { animation: 'none' } : undefined}
+      >
         {heading.children.map((item) => (
           <NestedToc heading={item} key={item.id} />
         ))}
--- a/packages/site/src/components/Navigation/TableOfContentsItems.tsx
+++ b/packages/site/src/components/Navigation/TableOfContentsItems.tsx
@@ -166,8 +166,12 @@ const NestedToc = ({ heading }: { heading: NestedHeading }) => {
   const nav = useNavigation();
   const [open, setOpen] = React.useState(startOpen);
   useEffect(() => {
-    if (nav.state === 'idle') setOpen(startOpen);
-  }, [nav.state]);
+    // On pathname update, derive open state
+    // We don't need heading and baseurl in dependencies, since they're fixed
+    if (nav.state !== 'idle') return;
+    setOpen(childrenOpen([heading], pathname, baseurl).includes(heading.id));
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [nav.state, pathname]);
   const exact = pathnameMatchesHeading(pathname, heading, baseurl);
   if (!heading.children || heading.children.length === 0) {
     return (
--- a/packages/site/src/components/Navigation/TableOfContentsItems.tsx
+++ b/packages/site/src/components/Navigation/TableOfContentsItems.tsx
@@ -5,6 +5,7 @@ import type { Heading } from '@myst-theme/common';
 import {
   isExternalUrl,
   useBaseurl,
+  useIsStaticBuild,
   useLinkProvider,
   useNavLinkProvider,
   useNavOpen,
@@ -51,7 +52,7 @@ function pathnameMatchesHeading(pathname: string, heading: Heading, baseurl?: st
   // not strip the basename here).
   //
   // Also strip any trailing slash, since static builds end up with one and
-  // `heading.path` is slashless.
+  // `heading.path` needs to be without.
   let normed = pathname.endsWith('/') ? pathname.slice(0, -1) : pathname;
   if (baseurl && normed.startsWith(baseurl)) normed = normed.slice(baseurl.length);
   const headingPath = heading.path;
@@ -162,6 +163,7 @@ function LinkItem({
 const NestedToc = ({ heading }: { heading: NestedHeading }) => {
   const { pathname } = useLocation();
   const baseurl = useBaseurl();
+  const isStaticBuild = useIsStaticBuild();
   const startOpen = childrenOpen([heading], pathname, baseurl).includes(heading.id);
   const nav = useNavigation();
   const [open, setOpen] = React.useState(startOpen);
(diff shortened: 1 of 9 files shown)


stefanv:
Double check this part:

```patch
-    if (nav.state === 'idle') setOpen(startOpen);
-  }, [nav.state]);
+    // On pathname update, derive open state
+    // We don't need heading and baseurl in dependencies, since they're fixed
+    if (nav.state !== 'idle') return;
+    setOpen(childrenOpen([heading], pathname, baseurl).includes(heading.id));
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [nav.state, pathname]);
```

packages/site/src/components/Navigation/TableOfContentsItems.tsx
```diff
@@ -198,7 +204,16 @@ const NestedToc = ({ heading }: { heading: NestedHeading }) => {
             'cursor-pointer': !heading.path,
           })}
           heading={heading}
-          onClick={() => setOpen(heading.path ? true : !open)}
+          onClick={() => {
+            if (heading.path) {
+              // Clicking a header with its own page navigates away.
+              // We don't want to animate before doing so, since after the reload there will be an animation anyway.
```
choldgraf, line 210: is this the main reason we need this "is static build" logic?

choldgraf (approved):
All of this looks pretty reasonable to me and the tests are nice and clear to understand and help me trust that this will do what it says it will do! I fine these URL type bugs to be really hard to test out in real life.

My one question was about the "is static build" logic and why it's necessary - but if we really do need that, then it does seem fine, just want to make sure it's not unnecessary complexity!

packages/site/src/components/Navigation/TableOfContentsItems.tsx
```diff
@@ -162,12 +163,17 @@ function LinkItem({
...
   const [open, setOpen] = React.useState(startOpen);
   useEffect(() => {
-    if (nav.state === 'idle') setOpen(startOpen);
-  }, [nav.state]);
+    // On pathname update, derive open state
+    // We don't need heading and baseurl in dependencies, since they're fixed
+    if (nav.state !== 'idle') return;
+    setOpen(childrenOpen([heading], pathname, baseurl).includes(heading.id));
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [nav.state, pathname]);
```
agoose77, line 176: Can't we just
```suggestion
    if (nav.state !== 'idle') return;
    setOpen(startOpen);
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, [nav.state, startOpen]);
```

agoose77: There's no harm in leaving this as-is, and I'd like to get the PR list down --- so I'm merging this!

PR #939, opened 2026-07-25
(6) Use Module Federation to load user renderers

agoose77:
Needs #938

This PR adds support for Module Federation to load third-party (external) plugins. For now, this hot-loads them from the site config. For static builds, we could bake the module loading in.


agoose77:
I blew the lockfile away. There was something strange happening on my machine leading to a vite error when rendering pages. It might just have been `node_modules`, so we should check again.

agoose77:
I'm closing this to make it easier to stack the other PRs.

PR #944, opened 2026-08-07
Fix top-nav dropdown children ignoring `internal_domains`

chad-earthscope:
Fix top-nav dropdown children ignoring `internal_domains`, so fully-qualified internal URLs open in the same tab instead of always opening in a new tab.

When [combining multiple repos using the BASE_URL](https://jupyterbook.org/blog/posts/2026/multi-repo) and sharing a top nav, links need to use fully specified URLs or the subsites will add BASE_URL to relative URLs.  At the same time the current code does not perform an internal-domain check for children in the top nav menu items so these always open a new tab (`target="_blank"`).  So a catch-22 when trying to use children in the shared top nav.

### DIFF
--- a/.changeset/nav-dropdown-internal-domains.md
+++ b/.changeset/nav-dropdown-internal-domains.md
@@ -1,0 +1,5 @@
+---
+"@myst-theme/site": patch
+---
+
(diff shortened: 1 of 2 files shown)


agoose77:
Perfect, thanks @chad-earthscope!

PR #950, opened 2026-09-01
fix: use site URLs in sitemap and robots

### LINKED ISSUE #669: sitemap.xml and robots.txt issue in static site
This has been raised in mystmd https://github.com/jupyter-book/mystmd/issues/1976
But I found it's an issue that also affects myst-theme:

- myst-cli does not generate static sitemap_style.xsl and sitemap.xml
- robots.txt,sitemap.xml as well as some other files use "localhost:3000" url.
- settings BASE_URL when not using a `subfolder` in the domain cause issues at the home page...

Current workaround is to use the following workflow, but I'm looking for a better solution to keep using the upstream packages,

```yaml
name: MyST GitHub Pages Deploy
on:
  push:
    branches: [main]
env:
  # `BASE_URL` determines, relative to the root of the domain, the URL that your site is served from.
  # E.g., if your site lives at `https://mydomain.org/myproject`, set `BASE_URL=/myproject`.
  # If, instead, your site lives at the root of the domain, at `https://mydomain.org`, set `BASE_URL=''`.
  # BASE_URL: https://blog.ayghri.com
  CUSTOM_DOMAIN: https://blog.ayghri.com

# Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
permissions:
  contents: read
  pages: write
  id-token: write
# Allow only one concurrent deployment, skipping runs queued between the run in-progress and latest queued.
# However, do NOT cancel in-progress runs as we want to allow these production deployments to complete.
concurrency:
  group: 'pages'
  cancel-in-progress: false
jobs:
  deploy:
    environment:
      name: github-pages
      url: ${{ steps.deployment.outputs.page_url }}
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - name: Setup Pages
        uses: actions/configure-pages@v3
      - uses: actions/setup-node@v4
        with:
          node-version: 18.x
      # - name: Install MyST
        # run: npm install -g mystmd@1.6.3
      - name: clone custom MyST
        uses: actions/checkout@v4
        with:
          repository: ayghri/mystmd
          path: .cache/mystmd-fork
      - name: install custom myst
        working-directory: .cache/mystmd-fork
        run: |
          npm install
          npm run build
          npm run link

      - name: Build HTML Assets
        run: |
          myst build --html
          sed -i -e 's|http://localhost:3000|$CUSTOM_DOMAIN|g' $(grep -RIl "http://localhost:3000" ./_build)
      - name: Upload artifact
        uses: actions/upload-pages-artifact@v3
        with:
          path: './_build/html'
      - name: Deploy to GitHub Pages
        id: deployment
        uses: actions/deploy-pages@v4
```

cnotin:
Fixes #669

Summary:
- Resolve public site URLs from SITE_URL, site.url, Read the Docs, or the request origin.
- Infer the routing and asset base path from a configured site URL, and reject conflicting BASE_URL values.
- Apply this to Book and Article sitemap, robots, and page loader data.

Testing:
- npx vitest run packages/site/src/utils.spec.ts
- npx bun run compile
- npx prettier --check (changed files)


cnotin:
This looks good 👌
image
image
image

packages/site/src/utils.spec.ts
```diff
@@ -0,0 +1,71 @@
+import { afterEach, describe, expect, it } from 'vitest';
+import type { SiteManifest } from 'myst-config';
+import { createRobotsTxt } from './seo/robots.js';
+import { createSitemap } from './seo/sitemap.js';
+import { getCanonicalSiteUrl, normalizeCanonicalUrl } from './utils.js';
+
+const request = new Request('http://localhost:3000/page');
+const config = (canonical_url?: string) =>
```
stefanv: Is this just to get type annotations working?

packages/site/src/utils.ts
```diff
@@ -1,14 +1,58 @@
...
   const domain = `${url.protocol}//${url.hostname}${url.port ? `:${url.port}` : ''}`;
   return domain;
 }
 
+type SiteManifestWithCanonicalUrl = SiteManifest & { canonical_url?: string };
+
+/**
+ * Normalize an absolute public site URL for use in generated SEO files.
+ */
+export function normalizeCanonicalUrl(value: string, source = 'canonical site URL'): string {
```
stefanv: Can probably use `normalizeBaseurl`; this fn does very little normalization (removes trailing `/`?).

packages/site/src/utils.ts
```diff
@@ -1,14 +1,58 @@
...
+  if (url.search || url.hash) {
+    throw new Error(`${source} must not include a query string or fragment: ${value}`);
+  }
+  return url.href.replace(/\/+$/, '');
+}
+
+/**
+ * Resolve the full public base URL used by sitemaps and robots.txt.
+ */
+export function getCanonicalSiteUrl(request: Request, config?: SiteManifestWithCanonicalUrl) {
```
stefanv: Should this logic rather be incorporated into how we determine BASE_URL?

packages/site/src/utils.ts
```diff
@@ -1,14 +1,58 @@
...
+  if (config?.canonical_url) {
+    return normalizeCanonicalUrl(config.canonical_url, 'site.canonical_url');
+  }
+  if (process.env.READTHEDOCS_CANONICAL_URL) {
+    return normalizeCanonicalUrl(
+      process.env.READTHEDOCS_CANONICAL_URL,
+      'READTHEDOCS_CANONICAL_URL',
+    );
+  }
+  const baseUrl = process.env.BASE_URL;
```
stefanv: Ensure that this is the only way to access BASE_URL, and check whether programmatic access isn't already possible.

netlify.toml
```diff
@@ -27,5 +27,5 @@
     # Install Python requirements and build docs
     pip install uv && \
     python -m uv pip install -r docs/requirements.txt --system && \
-    make build-docs
+    SITE_URL="$DEPLOY_PRIME_URL" make build-docs
```
stefanv, line 30: What is DEPLOY_PRIME_URL?

stefanv (commented):
This looks like it's on the right track; I suspect we can reuse some internal machinery.

netlify.toml
cnotin: https://docs.netlify.com/build/configure-builds/environment-variables/#deploy-urls-and-metadata
> DEPLOY_PRIME_URL: URL representing the primary URL for an individual deploy, or a group of them, like branch deploys and Deploy Previews; for example, https://feature-branch--petsof.netlify.app or https://deploy-preview-1--petsof.netlify.app. If you set up an [automatic deploy subdomain](https://docs.netlify.com/manage/domains/manage-domains/automatic-deploy-subdomains), this URL will update.

it's to get the random URL where it's deployed, like https://deploy-preview-950--myst-theme.netlify.app/ in this PR

packages/site/src/utils.spec.ts
cnotin: Yes. myst-theme currently depends on a released myst-config type that does not yet contain canonical_url; this intersection is a temporary compatibility bridge for the companion mystmd change https://github.com/jupyter-book/mystmd/pull/3042/changes#diff-77465c887b6c95f0d6dceb46b9f18a86789e087e6ea57269af151281a21d1c35

cnotin: Do you want a FIXME/TODO comment for later cleanup perhaps?

packages/site/src/utils.ts
cnotin: Good point. Canonical URLs still need URL parsing/protocol/query/fragment validation, but their trailing-path normalization can reuse `normalizeBaseurl`. Applying.

packages/site/src/utils.ts
cnotin: I’d keep these concepts separate: `BASE_URL` is a path-only routing and asset prefix, while the canonical URL is an absolute public origin plus optional path. I’ll centralize the path-only `BASE_URL` reading, then compose it with the request origin only as the final fallback. Extracting a `getBaseUrl()` helper

packages/site/src/utils.ts
cnotin: I found direct environment reads in Book and Article roots plus the new resolver; there is no existing server-side accessor. I’ll add one shared accessor and use it from all three locations.

cnotin:
Moved shared functions to `myst-config` package in `mystmd` so I had to import the future new package version. It's the reason why the Netlify CI fails currently until the new package version is published.

cnotin:
As per the AI policy, I've reviewed manually the code and it looks good to me. I've also steered it until it reached a state that satisfied me.

cnotin:
As discussed in https://github.com/jupyter-book/mystmd/pull/3042#issuecomment-5767304077 and above, removed `SITE_URL` and made `BASE_URL` accept both sub-folder or full URL formats

cnotin:
CI is red because it requires new code, shared in myst-common, which requires https://github.com/jupyter-book/mystmd/pull/3042 to be merged and a new version released.

PR #951, opened 2026-09-03
Render breadcrumbs in the book theme

### LINKED ISSUE #817: Add breadcrumb links to parent sections in the book theme
For multi-page projects (ie, those that would use the book theme), you often have content that's nested inside of other content. In this case, it's useful to have some context about the "parent" of a page and where it sits in the project hierarchy. This is often called **breadcrumbs** in websites.

I think we should allow book authors to include breadcrumbs in their site, it would have a structure like `🏠 > [parent page link] > current page`, maybe somewhere around here:

Image

And triggered somewhere around here:

```yaml
project:
  site:
    options:
      breadcrumbs:
        display: true # true by default
```

### Examples

Here's an example from pydata-sphinx-theme:

Image

And from docusaurus:

Image

choldgraf:
This adds parent breadcrumbs to pages in the book theme. It doesn't touch the article theme!

CleanShot 2026-09-04 at 15 41 53

### Decisions I made

- It lives just to the left of the "github for the page" buttons - that is usually whitespace and felt like the natural place to put this.
- It uses the site manifest to figure out who the parents are etc, so the URLs come from the myst build and we don't have to worry about `folders` changing behavior
- The "home" link is shown on every page other than the landing page - I figure for the landing page, you often want a different UI setup
- If the TOC has a title but no URL/file attached to it, it'll show up as an unlinked word but will still be in the breadcrumbs!
- If `subject` is also defined on a page it adds a little `|` and puts the subject to the right of the breadcrumbs, instead of clobbering it. It felt like this was a bit less-disruptive.
- That line can now grow to take up multiple lines for narrow screens or really long breadcrumb trails
- If short-title is defined, it uses that
- If it's a really long title, it'll get truncated with a `...`
- Vertically aligned the two sidebars and the meta / breadcrumbs area with light whitespace adjustments

### Concerns

- I feel like there's a lot of "UI" taking up the vertical space in general. I'd be a fan of shrinking down some of the padding so that the page title is a bit higher up, but that felt out of scope for this.

### Examples

<details>
<summary>Wide</summary>

CleanShot 2026-09-03 at 16 18 42@2x

</details>

<details>
<summary>Mobile</summary>

CleanShot 2026-09-03 at 16 14 07@2x

</details>

_Used Claude Fable to help with the implementation, along with the [`/grill-me` skill](https://www.aihero.dev/skills-grill-me) to do a bunch of design question exploration before getting to implementation, and in a few rounds of subsequent iteration on the end outcome. Then I went through and added comments and cleaned up the code here and there so it made more sense to me...I should really write this up in a blog post or something so that I don't have to write it every time 😅_

---

- closes #817

### DIFF
--- a/.changeset/book-breadcrumbs.md
+++ b/.changeset/book-breadcrumbs.md
@@ -6,3 +6,4 @@
 ---
 
 Add a breadcrumb trail above the page title in the book theme, showing the page's ancestors from the table of contents.
+The top nav now renders at the 60px height that `DEFAULT_NAV_HEIGHT` uses, so the sidebar / outline / heading scroll line up.
--- a/.changeset/book-breadcrumbs.md
+++ b/.changeset/book-breadcrumbs.md
@@ -1,0 +1,8 @@
+---
+"@myst-theme/common": minor
+"@myst-theme/frontmatter": minor
+"@myst-theme/site": minor
+"@myst-theme/book": minor
+---
+
+Add a breadcrumb trail above the page title in the book theme, showing the page's ancestors from the table of contents.
--- a/docs/reference/title-block.md
+++ b/docs/reference/title-block.md
@@ -1,0 +1,13 @@
+---
+title: Title block with a subject and a long title that needs a short title
+short_title: Title block
+subject: Reference
+subtitle: A page that exists to show off the header row above the title
+---
+
+# Title block
+
+This page sets `subject`, `subtitle`, and `short_title` in its frontmatter.
+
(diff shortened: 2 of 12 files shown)


choldgraf:
@agoose77 and @stefanv I think that you should prioritize the react upgrade so no worries if it takes a bit of time to get to this for a look-through!

choldgraf:
My GIF made me realize that the left/right sideabrs, and the middle button area, were not vertically aligned! So I just pushed a little fix to adjust the whitespace so they line up

PR #955, opened 2026-09-19
Derive compute feature flags from the project manifest

### LINKED ISSUE #400: `article-theme` cannot handle
### Describe the bug

**context**
When attempt to use `article-theme` with computational content that uses

```
project:
   jupyter:
       lite: true
```

**expectation**
We should be able to use this option in the same way as `book-theme` does.

**bug**
But instead we get a fatal error

```
Application Error
TypeError: Failed to construct 'URL': Invalid URL
    at http://localhost:3000/myst_assets_folder/_shared/chunk-6MJEHIZG.js:68048:19
    at LaunchBinder (http://localhost:3000/myst_assets_folder/_shared/chunk-6MJEHIZG.js:68083:27)
    at renderWithHooks (http://localhost:3000/myst_assets_folder/_shared/chunk-SMG2HPMX.js:14053:26)
    at updateFunctionComponent (http://localhost:3000/myst_assets_folder/_shared/chunk-SMG2HPMX.js:16459:28)
    at beginWork (http://localhost:3000/myst_assets_folder/_shared/chunk-SMG2HPMX.js:17794:22)
    at beginWork$1 (http://localhost:3000/myst_assets_folder/_shared/chunk-SMG2HPMX.js:21631:22)
    at performUnitOfWork (http://localhost:3000/myst_assets_folder/_shared/chunk-SMG2HPMX.js:21076:20)
    at workLoopSync (http://localhost:3000/myst_assets_folder/_shared/chunk-SMG2HPMX.js:21015:13)
    at renderRootSync (http://localhost:3000/myst_assets_folder/_shared/chunk-SMG2HPMX.js:20994:15)
    at recoverFromConcurrentError (http://localhost:3000/myst_assets_folder/_shared/chunk-SMG2HPMX.js:20614:28
```

### Reproduce the bug

1. set your site template to be `article-theme`
2. set the `jupyter: lite: true` option

### List your environment

_No response_

alanlujan91:
Closes #400. Also addresses the bad launch links in #636.

## Problem

On an article-theme site with `jupyter: lite: true`, no cell runs in the browser, and the control
offering to start a session replaces the page with `TypeError: Failed to construct 'URL'`. The same
project on book-theme runs its cells.

`ComputeOptionsProvider` derives the setup already, calling `useProjectManifest()` and
`thebeFrontmatterToOptions` to get `useBinder` and `useJupyterLite`. It then takes `features` as a
required prop that each theme hardcodes to its own guess at that same setup:

```
themes/article/app/routes/$.tsx   { notebookCompute: false, ..., launchBinder: true }
themes/book/app/routes/$.tsx      { notebookCompute: true,  ..., launchBinder: false }
```

The gating code behind them is identical (`Article.tsx:93`, book `ArticlePage.tsx:112`), so the
literals are the whole difference, and both of article's are wrong:

1. `notebookCompute: false` suppresses `NotebookToolbar` under any setup.
2. `launchBinder: true` under lite builds an href from `server.userServerUrl`, which thebe sets to
   `/` when there is no server, so `new URL('/')` throws during render.

Book escapes both by accident. Its `notebookCompute: true` suits either setup. Its `launchBinder`
reaches no consumer, since `themes/article/app/components/ArticlePage.tsx` is the only place that
flag is read.

## Change

`features` becomes optional and defaults from the options the provider already computed. Both themes
drop their literal.

```tsx
notebookCompute: features?.notebookCompute ?? true,
figureCompute: features?.figureCompute ?? true,
launchBinder: features?.launchBinder ?? !optionsWithOverrides?.useJupyterLite,
```

`launchBinder` keys off lite rather than binder deliberately. `useBinder` is set only for
`jupyter.binder`, so gating on it would hide the launch control under `jupyter: true` and
`jupyter.server`, which #636 reports as working today. Lite is the one setup with nowhere to launch.

`NotebookToolbar` now honours the same flag for its own launch action, which previously bypassed
`features` and opened `${server.settings.baseUrl}?token=${server.settings.token}` regardless.
In-page execution and the launch controls become independent, which is the separation proposed in
#636.

`getUserServerUrl` no longer throws on a non-absolute `userServerUrl`, which still matters for any
third-party theme overriding `launchBinder` under lite.

Each override is read with `??` rather than spread, so an explicit `undefined` cannot produce an
undefined flag, and `features` is in the memo dependencies so a changed override takes effect.

## Testing

`notebookCompute: true` and the URL guard were applied to the built bundle of a real lite site and
exercised in a browser. The toolbar appeared. `1 + 1` evaluated to `2`.

Honest caveat. This repo's build and tests have not run here, because bun is not installed on this
machine. My own reading is the only thing that has checked the rest, and CI is waiting on maintainer
approval. The binder path is untested; that `notebookCompute: true` serves it rests on book-theme
having carried that value all along.

Reproduction used:

```yaml
version: 1
project:
  title: Lite probe
  jupyter:
    lite: true
site:
  template: article-theme
```

with one page that has `kernelspec` in its frontmatter and a `{code-cell}` in its body.


alanlujan91:
Three good findings, all fixed in d803cf6. Two were real defects in my change, so thank you.

1. The `useBinder` gate (r4051823424) is the worst of them, and you are right. `thebeFrontmatterToOptions` sets `useBinder` only when `jupyter.binder` is present, so `jupyter: true` and `jupyter.server` leave it false, and my default would have hidden article-theme's launch control for configurations where it works today. #636 reports working launch buttons under `jupyter: true`. That is the case I would have broken. The flag now derives from `useJupyterLite`, the one setup with nowhere to launch:

   ```tsx
   launchBinder: features?.launchBinder ?? !optionsWithOverrides?.useJupyterLite,
   ```

2. `NotebookToolbar` bypassing the flag (r4051823407) is also right. `handleLaunch` opens `${server.settings.baseUrl}?token=${server.settings.token}` without consulting `features`, so the separation #636 asks for was half done. The toolbar now reads `useComputeOptions()` and gates its `Launch` button on the same derived flag. On scope, this was already reachable on book-theme, which has carried `notebookCompute: true` all along, so turning the toolbar on for article-theme widened an existing gap.

3. `features` missing from the memo deps (r4051823426) is fixed. It was absent before this PR as well, but the PR is what turns the override into a documented guarantee, so a stale one would have been a promise the code does not keep.

I took the fourth point from your summary too, that a partial override can produce undefined flags. Each key is now read with `??`, which handles an explicit `undefined` and keeps the context type honest.

One caveat that applies to the whole PR. I cannot run this repo's build or tests locally, so none of this is typechecked beyond my own reading, and CI here still shows `action_required` pending maintainer approval. What was verified in a browser is `notebookCompute: true` and the URL guard, applied to the built bundle of a real lite site, where the notebook toolbar appeared and `1 + 1` evaluated to `2`.

